PR #46, opened 2021-09-28
Update knative.dev dependencies to v0.26

antoineco:
Follow up to my last PR: #45

Now that we are using Knative v0.25 libs, do the next and final update to v0.26.

Notable changes:

- `controller.NewImplFull` is no longer available and was replaced with `NewContext`.
    - See https://github.com/knative/pkg/pull/2222
- `resolver.NewURIResolver` is no longer available and was replaced with `NewURIResolverFromTracker`.
    - See https://github.com/knative/pkg/pull/2228

---

`go get` update log:

```console
$ go get knative.dev/eventing@v0.26.0
go get: upgraded contrib.go.opencensus.io/exporter/prometheus v0.3.0 => v0.4.0
go get: upgraded github.com/golang/groupcache v0.0.0-20200121045136-8c9f03a8e57e => v0.0.0-20210331224755-41bb18bfe9da
go get: upgraded github.com/prometheus/common v0.26.0 => v0.30.0
go get: upgraded github.com/prometheus/statsd_exporter v0.20.0 => v0.21.0
go get: upgraded go.uber.org/zap v1.18.1 => v1.19.0
go get: upgraded golang.org/x/crypto v0.0.0-20210513164829-c07d793c2f9a => v0.0.0-20210817164053-32db794688a5
go get: upgraded golang.org/x/net v0.0.0-20210805182204-aaa1db679c0d => v0.0.0-20210825183410-e898025ed96a
go get: upgraded k8s.io/api v0.20.7 => v0.21.4
go get: upgraded k8s.io/apiextensions-apiserver v0.20.7 => v0.21.4
go get: upgraded k8s.io/apimachinery v0.20.7 => v0.21.4
go get: upgraded k8s.io/code-generator v0.20.7 => v0.21.4
go get: upgraded k8s.io/klog/v2 v2.5.0 => v2.8.0
go get: upgraded k8s.io/kube-openapi v0.0.0-20210113233702-8566a335510f => v0.0.0-20210305001622-591a79e4bda7
go get: upgraded knative.dev/eventing v0.25.1 => v0.26.0
go get: upgraded knative.dev/pkg v0.0.0-20210902173607-844a6bc45596 => v0.0.0-20210919202233-5ae482141474
go get: upgraded sigs.k8s.io/structured-merge-diff/v4 v4.0.3 => v4.1.2
```

```console
$ go get knative.dev/serving@v0.26.0
[...]

PR #48, opened 2021-09-29
Prepare update to Knative v0.26

### LINKED ISSUE #30: Remove usage of ObserveFinalizeKind
This interface was removed.
Here is some context: https://github.com/knative/pkg/issues/2021

antoineco:
The following interface has been deprecated and no longer exists in Knative 0.26:

- The `ReadOnlyFinalizer` interface is no longer available and was replaced with `OnDeletionInterface`.
    - See https://github.com/knative/pkg/pull/2248
    -  Closes #30 as a side effect.

Will be done while updating to Knative 0.26, because not yet in 0.25:

- `controller.NewImplFull` is no longer available and was replaced with `NewContext`.
    - See https://github.com/knative/pkg/pull/2222
- `resolver.NewURIResolver` is no longer available and was replaced with `NewURIResolverFromTracker`.
    - See https://github.com/knative/pkg/pull/2228

### DIFF
--- a/pkg/routing/adapter/filter/reconciler.go
+++ b/pkg/routing/adapter/filter/reconciler.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 
 	"knative.dev/pkg/controller"
 	"knative.dev/pkg/reconciler"
@@ -38,7 +39,7 @@ type Reconciler struct {
 var (
 	_ reconcilerv1alpha1.Interface         = (*Reconciler)(nil)
 	_ reconcilerv1alpha1.ReadOnlyInterface = (*Reconciler)(nil)
-	_ reconcilerv1alpha1.ReadOnlyFinalizer = (*Reconciler)(nil)
+	_ reconciler.OnDeletionInterface       = (*Reconciler)(nil)
 )
 
 // ReconcileKind implements reconcilerv1alpha1.Interface.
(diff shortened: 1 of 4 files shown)


antoineco:
@odacremolbap I decided to move the update of deprecated interfaces outside of the Kn 0.26 PR.

pkg/routing/adapter/filter/reconciler.go
```diff
@@ -67,8 +68,12 @@ func (r *Reconciler) reconcile(ctx context.Context, f *v1alpha1.Filter) error {
...
 
-// ObserveFinalizeKind implements reconcilerv1alpha1.ReadOnlyFinalizer.
-func (r *Reconciler) ObserveFinalizeKind(ctx context.Context, f *v1alpha1.Filter) reconciler.Event {
+// ObserveDeletion implements reconciler.OnDeletionInterface.
+func (r *Reconciler) ObserveDeletion(ctx context.Context, key types.NamespacedName) error {
+	f := &v1alpha1.Filter{}
+	f.SetName(key.Name)
+	f.SetNamespace(key.Namespace)
+
 	return r.finalize(ctx, f)
```
odacremolbap, line 77: @tzununbekov question not related to the validity of this PR.

Are calls from this reconciler to `RegisterHandlerFor` and `DeregisterHandlerFor` noop, and thus can be removed and there would be no need to implement `OnDeletionInterface`?

antoineco: Hm, nice catch. Not sure why we even bothered implementing a dummy interface.
https://github.com/triggermesh/triggermesh/blob/54e355dc6990364225b38f41fdf1390aa13f4423/pkg/routing/adapter/filter/adapter.go#L96-L104

odacremolbap:
I think we need some libraries updates from https://github.com/triggermesh/triggermesh/pull/46/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6 for this PR to work.

Tests are failing for me.
Either we merge both PRs or we merge them one by one, knowing that they will be in an test-error state for minutes.

@antoineco ^^

pkg/routing/adapter/filter/reconciler.go
tzununbekov: right, I think I copied these dummy methods from somewhere else where they were implemented.

antoineco:
@odacremolbap you're absolutely right, I missed that one because I was testing from `kn-0.26` branch directly.
Looks like knative/pkg removed things without actually deprecating them first, within the same release.

Fixed by moving this commit to `kn-0.26`. Tests are green now on this branch too.
Description updated.

pkg/routing/adapter/filter/reconciler.go
antoineco: See #49

PR #49, opened 2021-09-29
Remove unused Reconciler bits from Filter adapter

antoineco:
The filter adapter doesn't cache individual HTTP handlers. Instead, it queries its `Filter` lister on every HTTP request and matches on `namespace/name` directly.

Therefore, we can remove the controller logic entirely, because it was never implemented.


pkg/routing/adapter/filter/controller.go
```diff
@@ -1,52 +0,0 @@
...
-		r := &Reconciler{
-			adapter: a.(MTAdapter),
-		}
-		impl := reconcilerv1alpha1.NewImpl(ctx, r, controller.Opts(component))
-
-		informerv1alpha1.Get(ctx).Informer().AddEventHandler(pkgcontroller.HandleAll(impl.Enqueue))
-
-		return impl
-	}
-}
```
antoineco, line 52: Notice how this controller was registering an event handler that was simply doing `return nil` inside `adapter.go`.

odacremolbap (approved):

PR #50, opened 2021-09-29
Fix inconsistencies in routing API and reconciler

antoineco:
Some minor fixes in code comments and method names that I didn't know where to fit, so I'm just sending them as their own PR.

See comments for details.


pkg/apis/routing/v1alpha1/interfaces.go
```diff
@@ -42,9 +42,9 @@ type Router interface {
 	GetStatusManager() *RouterStatusManager
 	// GetEventTypes returns the event types generated by the router.
 	GetEventTypes() []string
-	// AsRouter returns a unique reference to the router suitable for
-	// use as a CloudEvent router attribute.
-	AsRouter() string
+	// AsEventSource returns a unique reference to the router suitable for
+	// use as a CloudEvent source attribute.
+	AsEventSource() string
```
antoineco, line 47: This was probably renamed as part of a mass string replacement, but the meaning of the method is really `AsEventSource`. As in "return (self) as a CloudEvent 'source' attribute".

pkg/apis/routing/v1alpha1/filter_types.go
```diff
@@ -78,7 +78,7 @@ func (f *Filter) GetStatus() *duckv1.Status {
 	return &f.Status.Status
 }
 
-// AsRouter implements Router.
-func (f *Filter) AsRouter() string {
+// AsEventSource implements Router.
+func (f *Filter) AsEventSource() string {
 	return "filter/" + f.Name
 }
```
antoineco, line 84: I didn't touch the logic here, but it's not particularly helpful because we don't know the namespace of the object.
Mind if I add it here too?

pkg/routing/reconciler/common/labels.go
```diff
@@ -33,7 +33,7 @@ const (
 
 // Common label values
 const (
-	partOf           = "knative-sources"
-	managedBy        = "knative-sources-controller"
+	partOf           = "triggermesh"
+	managedBy        = "triggermesh-controller"
 	componentAdapter = "adapter"
```
antoineco, line 38: Now I know where this code was copied from 😉

pkg/routing/reconciler/common/reconcile.go
```diff
@@ -100,16 +100,16 @@ func (r *GenericDeploymentReconciler) ReconcileAdapter(ctx context.Context, ab A
...
+	sink := router.GetSink()
 
-	if sinkRef := &router.GetSink().Ref; *sinkRef != nil && (*sinkRef).Namespace == "" {
-		(*sinkRef).Namespace = router.GetNamespace()
+	if sinkRef := sink.Ref; sinkRef != nil && sinkRef.Namespace == "" {
+		sinkRef.Namespace = router.GetNamespace()
 	}
 
-	return r.SinkResolver.URIFromDestinationV1(ctx, sink, router)
+	return r.SinkResolver.URIFromDestinationV1(ctx, *sink, router)
```
antoineco, line 109: We store the result of `GetSink()` into a variable which we don't use in the conditional below.
With this change, we do.

Also there was a lot of unnecessary pointer magic here (I'm the original perpetrator of this crime).

pkg/sources/reconciler/common/adapter.go
```diff
@@ -113,7 +113,6 @@ func commonAdapterDeploymentOptions(src v1alpha1.EventSource) []resource.ObjectO
 // NewAdapterKnService is a wrapper around resource.NewKnService which
 // pre-populates attributes common to all adapters backed by a Knative Service.
 func NewAdapterKnService(src v1alpha1.EventSource, sinkURI *apis.URL, opts ...resource.ObjectOption) *servingv1.Service {
-	app := ComponentName(src)
```
antoineco, line 116: This variable is very unnecessary.

odacremolbap (approved):

PR #64, opened 2021-09-30
Read adapter's Image from implicit env var name

antoineco:
The kind of the Source reconciled by the controller is passed to `envconfig.MustProcess()`, which automatically resolves Image variables to `<KIND>_IMAGE`. Overriding this implicit name by hand in every reconciler is error prone and should be avoided.

This PR removes the only three occurrences where the implicit variable name was being overridden.


sebgoa:
this was closed without merging ?

antoineco:
Yes but this was done by GitHub, and not by Jeff (although GitHub reports it was). That's because the target branch was removed as the result of merging another PR.

It was reopened and merged in #65.

PR #66, opened 2021-10-01
Add GitHub Actions workflow for checking OSS licenses

antoineco:
Integrates with GitHub Actions to provide immediate feedback on PRs when OSS licenses aren't applied properly:
- Apache license headers are missing / non-compliant
- The LICENSES directory is outdated

I deliberately pushed a broken commit at https://github.com/antoineco/triggermesh/pull/1 to demonstrate what is reported when we don't comply.


cmd/triggermesh-controller/main.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 TriggerMesh Inc.
+Copyright 2020 TigerMesh Inc.
```
odacremolbap, line 2: mmm ...

odacremolbap: aah, the bogus commit ...

odacremolbap (approved):
nice!!!

+1 if last commit reverted

PR #76, opened 2021-10-01
Validate license headers in YAML files

antoineco:
On top of validating Go files, validate our YAML manifests.


.github/workflows/config/reviewdog.yaml
```diff
@@ -12,3 +12,12 @@ runner:
     errorformat:
       - '%A%f:%l: %m'
       - '%C%.%#'
+  license-yaml:
+    cmd: >-
+      boilerplate-check check
+      --boilerplate=hack/boilerplate/boilerplate.yaml.txt
+      --file-extension=yaml
+      --exclude='(^(\.|config/samples/))'
```
antoineco, line 20: I wasn't sure whether we wanted to enforce headers inside samples or not.
Thoughts?

JeffNeff: +1 they should be there

odacremolbap: No strong opinion but mostly concur with Jeff.

config/300-filter-crd.yaml
```diff
@@ -4,7 +4,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#     https://www.apache.org/licenses/LICENSE-2.0
+#     http://www.apache.org/licenses/LICENSE-2.0
```
cab105, line 7: Does it matter if using http vs. https, or did the linter complain?

antoineco: It complained yes. The only thing it ignores is the year.

I don't think it matters personally, but I won't why these 4 specific files had HTTPS.

cab105: Guess it depends on what was used as the basis for the copy/pasta boilerplate.

.github/workflows/config/reviewdog.yaml
antoineco: Thanks for the comments!
I added missing headers to all samples reported by the linter.

odacremolbap (approved):

PR #80, opened 2021-10-01
Updates AWS Targets CRD to Reflect Accepted/Response Event Types

JeffNeff:


cab105 (approved):
Good catch!

config/300-aws-crd.yaml
```diff
@@ -144,6 +147,15 @@ metadata:
     knative.dev/crd-install: 'true'
     triggermesh.io/crd-install: 'true'
     duck.knative.dev/addressable: 'true'
+  annotations:
+    registry.triggermesh.io/acceptedEventTypes: |
+      [
+        { "type": "*" }
+      ]
```
antoineco, line 154: In the future I think we can omit this entirely and just assume that a target accepts any type unless otherwise specified. At least I think that could be a safe assumption to make.

JeffNeff: I really like that its specified.. but if you have a strong opinion against this we an remove it.

JeffNeff: its lets, at least me, know that its been looked at and its supposed to accept arb. events.

antoineco: I don't have a strong opinion against no, not at all.

PR #83, opened 2021-10-01
Implements Package Replier in Splunk

JeffNeff:


pkg/targets/adapter/splunktarget/adapter_test.go
```diff
@@ -18,14 +18,15 @@ package splunktarget
...
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/assert"
-
 	cloudevents "github.com/cloudevents/sdk-go/v2"
+	"github.com/stretchr/testify/assert"
 
+	"github.com/triggermesh/triggermesh/pkg/apis/targets/v1alpha1"
+	targetce "github.com/triggermesh/triggermesh/pkg/targets/adapter/cloudevents"
```
antoineco, line 29: Care to separate these two from knative.dev imports?

pkg/targets/adapter/splunktarget/adapter_test.go
```diff
@@ -58,36 +59,43 @@ func (c *mockedSplunkClient) LogEvent(in *splunk.Event) error {
 func TestReceive(t *testing.T) {
 	testCases := map[string]struct {
 		client       *mockedSplunkClient
-		expectResult cloudevents.Result
+		expectResult *errorEvent
```
antoineco, line 62: `expectEvent` maybe?
Also shouldn't we still check the value of `cloudevents.Result`?

pkg/targets/adapter/splunktarget/adapter.go
```diff
@@ -82,13 +90,21 @@ func NewTarget(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClien
 		logger.Panicw("Invalid HEC endpoint URL "+env.HECEndpoint, zap.Error(err))
 	}
 
+	replier, err := targetce.New(env.Component, logger.Named("replier"),
+		targetce.ReplierWithStatefulHeaders(env.BridgeIdentifier),
+		targetce.ReplierWithStaticResponseType(v1alpha1.EventTypeSplunkResponse),
+		targetce.ReplierWithPayloadPolicy(targetce.PayloadPolicy(env.CloudEventPayloadPolicy)))
+	if err != nil {
+		logger.Panicf("Error creating CloudEvents replier: %v", err)
+	}
```
antoineco, line 99: Will this panic if `EVENTS_PAYLOAD_POLICY` isn't set?
I see the controller only sets it conditionally.

pkg/targets/adapter/splunktarget/adapter_test.go
```diff
@@ -58,36 +59,43 @@ func (c *mockedSplunkClient) LogEvent(in *splunk.Event) error {
...
 				logger:       logtesting.TestLogger(t),
 				ceClient:     adaptertest.NewTestClient(),
 				spClient:     tc.client,
 				defaultIndex: tDefaultIndex,
+				replier:      replier,
 			}
 
 			// invoke event callback
-			res := a.receive(context.Background(), newEvent(t))
+			fmt.Println(a)
```
antoineco, line 91: If you want to log things in tests, prefer `t.Log` or `t.Logf`.

pkg/targets/adapter/splunktarget/adapter_test.go
```diff
@@ -58,36 +59,43 @@ func (c *mockedSplunkClient) LogEvent(in *splunk.Event) error {
...
+			e, _ := a.receive(context.Background(), newEvent(t))
 
-			assert.Lenf(t, tc.client.inputRecorder, 1, "Client records a single request")
-			assert.EqualError(t, res, tc.expectResult.Error())
```
antoineco, line 90: We should keep testing that. The fact that we introduce replies doesn't mean all other tests go away.

antoineco (changes requested):
Requesting changes because we shouldn't delete existing tests.

pkg/targets/adapter/splunktarget/adapter_test.go
JeffNeff: got it!

pkg/targets/adapter/splunktarget/adapter_test.go
JeffNeff: got it!

pkg/targets/adapter/splunktarget/adapter.go
JeffNeff: No. It does not. same for `env.CloudEventPayloadPolicy`

pkg/targets/adapter/splunktarget/adapter_test.go
JeffNeff: got it!

pkg/targets/adapter/splunktarget/adapter_test.go
JeffNeff: I put `assert.Lenf(t, tc.client.inputRecorder, 1, "Client records a single request")` back but im having troubles getting `assert.EqualError(t, res, tc.expectResult.Error())` to work.. but isnt that replaced with the logic in `assert.Equal(t, eE, tc.expectEvent)` ?

PR #85, opened 2021-10-02
Updates Replier's Event Type Mapping Failure Logic

JeffNeff:
Previously we would fail when mapping event types. I think we should instead append the incoming event type 
 with ".response" and use this. So for example, if the incoming event type is `com.example.myevent` the response event type will be `com.example.myevent.response`

### DIFF
--- a/pkg/targets/adapter/cloudevents/doc.go
+++ b/pkg/targets/adapter/cloudevents/doc.go
@@ -98,6 +98,10 @@ Usage:
 
 	// Create replier that will use a different static type response
 	// depending on the incoming type.
+	// If the incoming event type is not mapped the response will be
+	// the same as the incoming event type with `.response` appended.
+	// ie. if the incoming event type is `com.example.myevent` the
+	// response will be `com.example.myevent.response`
 	mt = map[string]string{
 		"search.snowball.io": "list.snowball.io",
 		"update.snowball.io": "product.snowball.io",
--- a/pkg/targets/adapter/cloudevents/response_header.go
+++ b/pkg/targets/adapter/cloudevents/response_header.go
@@ -17,8 +17,6 @@ limitations under the License.
 package cloudevents
 
 import (
-	"fmt"
-
 	cloudevents "github.com/cloudevents/sdk-go/v2"
 )
 
@@ -44,7 +42,7 @@ func MappedResponseType(eventTypes map[string]string) ResponseHeaderValue {
 		if ok {
 			return v, nil
 		}
-		return "", fmt.Errorf("incoming type %q cannot be mapped to an outgoing event type", event.Type())
+		return event.Type() + ".response", nil
 	}
 }
 


odacremolbap:
Can you clarify the case you are covering here?

I guess that is a Target that accepts either 
- any event type, and you want it to reply with `<incoming-event> + .response`
- specific event types, and you want it to reply with `my.response.type`

This is what I don't like from this PR:
- If a developer made a mistake configuring the mapped responses, it will go unnoticed. I rather error than silently move forward.
- We have a response suffix option where you choose the suffix, but this one hardcodes to `.reponse`. While `.response` is the default suffix added if no option is selected, we need to let people choose, and in the case of the mapped response we are not doing that.

What I would do here is create at `response_header.go`, create a new option that can process an array of options until one of them returns a type:

```go
// ResponseFallbackOptionsType uses an array of ResponseHeaderValue options to
// iterate on until one of them returns a type.
func ResponseFallbackOptionsType(rs []ResponseHeaderValue) ResponseHeaderValue {
	return func(event *cloudevents.Event) (t string, err error) {
		for _, r := range rs {
			t, err = r(event)
			if err != nil {
				return t, nil
			}
		}
		return "", err
	}
}
```

then at `replier_options.go` create the option you are looking for:
```go
// ReplierWithMappedErrorAndSuffixDefaultResponseType uses a map string to look up response type, if
// no match is found it fallsback to a default Suffix being added to the incoming type.
func ReplierWithMappedErrorAndSuffixDefaultResponseType(resTypes map[string]string, defaultSuffix string) ReplierOption {
	return func(c *Replier) error {
		c.responseErrorType = ResponseFallbackOptionsType(
			[]ResponseHeaderValue{
				MappedResponseType(resTypes),
				SuffixResponseType(defaultSuffix),
			},
		)
		return nil
	}
}
```

Disclaimer: I haven't tested that code above, if you find a simpler way of doing this but that let people choose the suffix, do it your own way.

odacremolbap:
@JeffNeff 
are you solving this somewhere else?
or is the scenario for needing this mapping logic not yet here?

I was thinking on re-evaluating after the experience gained this last year, and already created [a ticket in that direction](https://github.com/triggermesh/triggermesh/issues/91).

JeffNeff:
I was aiming to include the Replier package into the AWS Targets with this. But I think maybe a new function `ReplierWithAribitraryEventTypes()` that would allow the dev to accept incoming events and return them with appended data to the type (".response" or ".error") might be a better approach.

JeffNeff:
What you posted does help me with some of the targets tho.

JeffNeff:
just fwiw: I still dont think this function should fail as it does.

JeffNeff:
I was not able to get this working. I was able to get it working like this tho -> https://github.com/triggermesh/triggermesh/pull/101 can you please review.

PR #89, opened 2021-10-04
Adds circle ci configuration

### LINKED ISSUE #88: Add CI/CD integration
Build and publish releases using CircleCI infra

sameersbn:
Depends #86
Closes #88

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,130 @@
+version: 2.1
+
+orbs:
+  aws-eks: circleci/aws-eks@1
+  gcp-cli: circleci/gcp-cli@1
+  go: circleci/go@1
+
+jobs:
+  checkout:
+    executor:
+      name: go/default
+      tag: '1.17'
+    steps:
+      - checkout
+      - go/mod-download-cached
+      - persist_to_workspace:
+          root: ~/
+          paths:
+            - go
+            - project
+
+  build:
+    executor:
+      name: go/default
+      tag: '1.17'
+    steps:
+      - attach_workspace:
+          at: ~/
+      - run:
+          name: Building package
+          command: make build
+
+  test:
+    executor:
+      name: go/default
+      tag: '1.17'
+    steps:
+      - attach_workspace:
+          at: ~/
+      - run:
+          name: Run fmt-test
+          command: make fmt-test
+      - run:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,136 @@
...
+          at: ~/
+      - run:
+          name: Run fmt-test
+          command: make fmt-test
+      - run:
+          name: Installing golangci-lint
+          command: make install-golangci-lint
+      - run:
+          name: Run lint
+          command: make lint
```
sameersbn, line 48: Looking at the github actions, i guess `lint` can be removed

antoineco: I think so too. It would be redundant, and static analysis is a really good fit for GitHub Actions since you get the result directly in the code diff.

I'd remove that part. (Unless some CircleCI integration can provide the same output)

.circleci/config.yml
```diff
@@ -0,0 +1,136 @@
...
+  build:
+    executor:
+      name: go/default
+      tag: '1.17'
+    steps:
+      - attach_workspace:
+          at: ~/
+      - run:
+          name: Building package
+          command: make build
```
cab105, line 31: Would we want to run codegen and build? If so, we can get away with just `make`

cab105 (approved):
Left a question if we want to address it. Otherwise PR is good for me.

.circleci/config.yml
antoineco: Running codegen would be useful, I agree, but only if we act on the output. E.g. fail the build if this step generates a diff, meaning devs need to commit the output of codegen.

cab105: This would be doable. `_output` is in the `.gitignore` so it shouldn't create any false positives.

.circleci/config.yml
sameersbn: removed!

.circleci/config.yml
sameersbn: IMO `codegen` should not be run in the CI, especially at the `build` stage. Maybe it could be run in the `test` stage where the build fails if a diff i generated as @antoineco  mentioned.

sameersbn: I am merging the PR because of the tight schedule.

PR #97, opened 2021-10-05
cloudbuild: set timeout to 1800s

sameersbn:
Last two CI jobs failed as a result of build timeouts

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,7 +60,7 @@ jobs:
       - gcp-cli/initialize
       - run:
           name: Publishing docker image
-          command: IMAGE_SHA=${CIRCLE_SHA1} IMAGE_TAG=${CIRCLE_TAG:-latest} make -j6 cloudbuild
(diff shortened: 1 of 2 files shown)


antoineco (commented):
That seems short to me considering how inefficient the build system is, but let's see.

antoineco (approved):

sameersbn:
the timeout is for each image build. with the previous 20m timeout there were two images that were taking longer to build. just trying with 30m to see if it helps

antoineco:
Ah right, it's per image, not the whole run. Sweet! 👍

sameersbn:
generally each image build does not take 20m, maybe 12m on average. but there are rare instances where the build kinda is stuck for more than 20m. I think it's more of a gcp or kaniko issue.

We can optimize the build after the release. We could just build the images in circleci itself and push. mostly likely it'll be faster.

antoineco:
Yep, totally makes sense.

PR #98, opened 2021-10-05
Remove 'build' step from CircleCI flow

antoineco:
@sameersbn this might be controversial so feel free to reject this PR and I won't insist 😛 

I'm proposing that we remove the `build` step (job) from the CircleCI flow, because
- the code is compiled in its entirety during static checks already (by `golangci-lint`)
- the code is also compiled during tests (the code that's being tested, at least)

That's at least 2 places where compilation errors would be caught.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,17 +19,6 @@ jobs:
             - go
             - project
 
-  build:
-    executor:
-      name: go/default
-      tag: '1.17'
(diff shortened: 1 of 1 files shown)


sameersbn:
👍 yes, it makes sense to remove the `build` step from the CI workflow.

sameersbn (approved):

PR #105, opened 2021-10-06
Remove direct dgrijalva/jwt-go dependency

### LINKED ISSUE #104: Replace insecure library: CVE-2020-26160
We have a direct dependency and many indirect ones to https://github.com/dgrijalva/jwt-go
All of them should be replaced with non vulnerable libraries.

odacremolbap:
- Remove direct `dgrijalva/jwt-go` depencendy.
- Upgrade some but not all the libraries that use it as an indirect dependency.

Closes #104 

There are still indirect dependencies that might need some more effort to remove.
https://github.com/knative/pkg/blob/a00ba487121ef45e7f14f1d59187ea6bb498bb4e/go.sum#L136


tzununbekov (approved):

go.mod
```diff
@@ -34,19 +34,19 @@ require (
...
-	github.com/Azure/go-autorest/autorest/adal v0.9.13
+	github.com/Azure/go-autorest/autorest v0.11.21
+	github.com/Azure/go-autorest/autorest/adal v0.9.16
 	github.com/Azure/go-autorest/autorest/azure/auth v0.5.8
 	github.com/Azure/go-autorest/autorest/to v0.4.0
 	github.com/ZachtimusPrime/Go-Splunk-HTTP/splunk/v2 v2.0.1
 	github.com/aliyun/aliyun-oss-go-sdk v2.1.9+incompatible
 	github.com/amenzhinsky/iothub v0.7.0
-	github.com/andygrunwald/go-jira v1.13.0
+	github.com/andygrunwald/go-jira v1.14.0
```
antoineco, line 44: It's still there, is that expected?

go.mod
```diff
@@ -34,19 +34,19 @@ require (
 	github.com/Azure/azure-sdk-for-go v57.1.0+incompatible
 	github.com/Azure/azure-service-bus-go v0.11.1
 	github.com/Azure/azure-storage-queue-go v0.0.0-20191125232315-636801874cdd
-	github.com/Azure/go-autorest/autorest v0.11.20
-	github.com/Azure/go-autorest/autorest/adal v0.9.13
+	github.com/Azure/go-autorest/autorest v0.11.21
+	github.com/Azure/go-autorest/autorest/adal v0.9.16
```
antoineco, line 38: Are those unrelated version bumps needed?
Can we push them separately?

odacremolbap: no, they are not needed.

I wanted to also remove transitive dependencies to clean `go.sum`, but it is tougher than expected.
I'll remove them

odacremolbap: the rabbit hole is here at the moment

https://github.com/etcd-io/etcd/issues/13390

PR #106, opened 2021-10-06
Replace GCloudPubSubResourceName with GCloudResourceName

FranBarrera:


antoineco (approved):
Thanks for the clean up!

PR #108, opened 2021-10-06
release: attach yaml manifest to releases

sameersbn:


Makefile
```diff
@@ -96,6 +99,9 @@ release: ## Build release binaries
 		done ; \
 	done
 
+	$(KUBECTL) create -f config --dry-run=client -o yaml | \
+	  $(SED) 's|ko://github.com/triggermesh/triggermesh/cmd/\(.*\)|$(IMAGE_REPO)/\1:${IMAGE_TAG}|' > $(DIST_DIR)/triggermesh-core.yaml
```
sameersbn, line 103: Since the YAML manifests are generated from the `config/` directory, the generated manifest won't include the `triggermesh` namespace creation. The namespace was removed from the configs to safeguard us from unintentionally deleting the `triggermesh` namespace.

Without the namespace in the config, the install guide would require that the user first create the `triggermesh` namespace.

odacremolbap: we could have the namespace creation somewhere in config and have a kustomize file, then issue a
```
$(KUBECTL)  create -k config/release --dry-run=client ...
```
or maybe remove the samples from config, add a namespace subfolder and issue a
```
$(KUBECTL) create -Rf config --dry-run=client ...
```

antoineco, line 103: The name "core" was banned from our vocabulary for marketing reasons. We should avoid using it anywhere public.

sameersbn: so just `triggermesh.yaml` then.

antoineco (approved):
👍 

Talking about releases, I've also been wondering whether generating all those binaries really made sense for software that is only supposed to run in containers. But that's the subject of another discussion.

sameersbn:
> 👍
> 
> Talking about releases, I've also been wondering whether generating all those binaries really made sense for software that is only supposed to run in containers. But that's the subject of another discussion.

We could drop the binaries from the release. Does not really make sense to have them attached to the release. Not only they are unnecessary, but they would bury the `triggermesh.yaml` file.

sameersbn:
@antoineco removed the binary artifacts from releases. PTAL

Makefile
```diff
@@ -81,20 +84,9 @@ deploy: ## Deploy TriggerMesh stack to default Kubernetes cluster using ko
 undeploy: ## Remove TriggerMesh stack from default Kubernetes cluster using ko
 	$(KO) delete -f $(BASE_DIR)/config
 
-release: ## Build release binaries
-	@set -e ; \
-	for bin in $(COMMANDS) ; do \
-		for platform in $(TARGETS); do \
```
antoineco, line 87: `$(TARGETS)` is possibly unused as the result of this change?

antoineco (approved):
@sameersbn thanks a lot! I think it makes more sense than publishing raw binaries.

Makefile
sameersbn: fixed

Makefile
sameersbn: @odacremolbap does this make sense? https://github.com/triggermesh/triggermesh/pull/108/commits/01350dc44d38806decbceba558e0c661b041dd37

sameersbn: i can confirm that `ko` does not bother about the contents of `config/namespace`

antoineco: I liked the idea of making the namespace part of a kustomization, but I guess the subdirectory works too. Neither `ko` nor `kubectl` recurse by default.

PR #110, opened 2021-10-06
Filter: Additional expression check, Tests

tzununbekov:
During manual tests, I stumbled upon a malformed filter expression that was not handled properly and caused webhook panic. This PR adds missing validation and tests for various expression cases.

### DIFF
--- a/pkg/routing/eventfilter/cel/cel.go
+++ b/pkg/routing/eventfilter/cel/cel.go
@@ -27,7 +27,7 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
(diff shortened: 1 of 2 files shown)


antoineco (approved):
Approving with a suggestion regarding test reports.

tzununbekov:
@antoineco could you elaborate on "suggestion regarding test reports"?

antoineco:
@tzununbekov so sorry! My comment on the code wasn't saved when I clicked "approve" and I don't know why 🤔 

I was simply suggesting we use `t.Run(name, ....)` inside the test loop otherwise the test report doesn't what test case is falling.

PR #111, opened 2021-10-06
ci: add `no_output_timeout: 30m` to publish-image step

sameersbn:


odacremolbap (approved):
so this is to avoid truncating logs, right?

cab105 (approved):

PR #112, opened 2021-10-06
Use standard naming scheme across all cmds

### LINKED ISSUE #12: Standardize naming scheme of programs cmd and config manifests
The naming convention of their prior homes were ported over to initially get the code in place first. This task ensures that it is addressed.

antoineco:
Closes #12

Updated:
- directory names under `cmd/`
- `ko` targets inside `config/500-controller.yaml`
- Make goals inside Dockerfiles
- references to default images in reconcilers implementations


odacremolbap:
> LGTM. But the overlays/staging/triggermesh-core/deployments.yaml patch should be updated so that the ci will properly patch the manifests

I don't follow, env vars are the same, it is just the image that changes, isn't it?

sameersbn:
> > LGTM. But the overlays/staging/triggermesh-core/deployments.yaml patch should be updated so that the ci will properly patch the manifests
> 
> I don't follow, env vars are the same, it is just the image that changes, isn't it?

The sed subst in the CI looks for a pattern that matches the cmd (https://github.com/triggermesh/triggermesh/blob/main/.circleci/config.yml#L89). Since some of the commands have been renamed, there won't be a match. So they need to be in sync.

antoineco:
@sameersbn on it.

PR #113, opened 2021-10-06
ci: fix overlays patch update

sameersbn:


.circleci/config.yml
```diff
@@ -84,7 +84,7 @@ jobs:
           working_directory: tmconfig/
           command: |
             IMAGE_REPO=$(sed -n -e 's/^IMAGE_REPO[[:space:]].*=[[:space:]]*\(.*\)$/\1/p' ~/project/Makefile)
-            for cmd in $(sed -n -e 's/^COMMANDS[[:space:]]*=[[:space:]]*\(.*\)$/\1/p' ~/project/Makefile); do
+            for cmd in $(ls ~/project/cmd/); do
```
antoineco, line 87: That pattern is usually considered [fragile](https://github.com/koalaman/shellcheck/wiki/SC2045).

Here is a suggestion:

```sh
for cmd in cmd/*; do
    echo ${cmd##*/}       # option 1
    echo $(basename $cmd) # option 2
done
```

.circleci/config.yml
```diff
@@ -84,7 +84,7 @@ jobs:
...
               sed -i overlays/<< parameters.cluster >>/triggermesh-core/deployments.yaml \
                 -e "s|gcr.io/triggermesh\(-private\)\?/${cmd}:.*|${IMAGE_REPO}/${cmd}:${CIRCLE_TAG:-${CIRCLE_SHA1}}|g"
             done
```
antoineco, line 90: Sorry can't comment on it, but there is also a "aws" at line 101 that needs to be removed.

sameersbn: The issue with that syntax is it will include `~/project/cmd/` in the listing. If I have to adopt this syntax then the block will have to include substitution to trip that part

.circleci/config.yml
sameersbn: updated! and applied the same to hack/release-notes.sh

antoineco (approved):

PR #116, opened 2021-10-06
Split out AWS Target Components

### LINKED ISSUE #102: Split out AWS Target Components
Various AWS Targets were originally created using a single target adapter to leverage common code specific to AWS. This should be split up where there is a 1:1 mapping between adapter and their CRD.

JeffNeff:
Closes #102


JeffNeff:
@cab105 https://github.com/triggermesh/triggermesh/issues/125

PR #118, opened 2021-10-07
Go linter warnings fixed and code quality badge added

tzununbekov:
- [Go Report Card](https://goreportcard.com/) badge added 
- Multiple `golint` warnings fixed. Feel free to update new code comments if they don't make much sense.

Only two warnings left. I'm not sure what would be the right name for common labels that will avoid `common` in its name:
```
pkg/routing/reconciler/common/adapter.go:154:6: func name will be used as common.CommonObjectLabels by other packages, and that stutters; consider calling this ObjectLabels
pkg/sources/reconciler/common/adapter.go:237:6: func name will be used as common.CommonObjectLabels by other packages, and that stutters; consider calling this ObjectLabels
```

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -16,13 +16,10 @@ linters-settings:
   stylecheck:
     checks:
     - all
+    - -ST1000
 
 issues:
   include:
     # don't exclude stylecheck issues about missing or malformed comments on package|func|type|var|const
     # checks: ST1000,ST1020,ST1021,ST1022 (https://staticcheck.io/docs/checks)
   - EXC0011
-  exclude-rules:
-    - linters:
-       - stylecheck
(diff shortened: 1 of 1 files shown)


tzununbekov:
If we could make `stylecheck` require comments on exported functions/types perhaps that would be enough. But in general, this time I run aggressive `golint` and all these recommendations were not that pointless. Maybe if we enable it in CI and after this PR there will be 0 warnings, it would be easier for us to stick to this practice?

antoineco:
Yes definitely. I was not saying this is a bad idea, just that stylecheck and golint have the same purpose.

We could select what checks we want to enable/disable like this:
https://golangci-lint.run/usage/linters/#stylecheck

PR #137, opened 2021-10-10
Optimize Go build cache in CI

antoineco:
I performed the following changes to the CircleCI config:
- The code (including test code) is now built once immediately after checkout to populate the Go build cache.
- This build cache is saved, and restored in subsequent jobs (test, publish).
- We use a 2-key caching strategy explained in a comment inside the CircleCI config.
- The source code is checked out again in every job instead of saving/restoring the workspace, which turns out to be faster.

This optimization of the build cache should speed up the current `test` job and the future `publish` job proposed in #132.

---

Proof that `checkout` is MUCH faster than `attach_workspace`:

image

image

---

Proof that save/restore of the Go mod and build caches work as expected:

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,35 +6,61 @@ orbs:
   go: circleci/go@1
 
 jobs:
-  checkout:
+  checkout-build:
     executor:
       name: go/default
       tag: '1.17'
     steps:
       - checkout
       - go/mod-download-cached
-      - persist_to_workspace:
-          root: ~/
+      - restore_cache:
+          name: Restoring Go build cache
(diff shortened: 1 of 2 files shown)


antoineco:
I need to dig whether we could improve the resource utilization (e.g. compile tests before running them), because linker processes are getting killed:

```
/usr/local/go/pkg/tool/linux_amd64/link: signal: killed
```

The current runner has 4 GB of memory, which should be enough for unit tests.

antoineco:
I'm closing this because it doesn't yield the result I expected (faster tests execution).

sameersbn:
oh ignore my question, I see that you were trying to speed that up itself.

PR #165, opened 2021-10-14
Increase Code Linting timeout to 3m

### LINKED ISSUE #163: Code Linting - golangci-lint: timeout exceeded
level=error msg="Timeout exceeded: try increasing it by passing --timeout option"

FranBarrera:
Closes #163

Default timeout is 1m.

### DIFF
--- a/.github/workflows/static.yaml
+++ b/.github/workflows/static.yaml
@@ -26,3 +26,4 @@ jobs:
       uses: golangci/golangci-lint-action@v2
       with:
         skip-go-installation: true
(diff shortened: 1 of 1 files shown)


antoineco:
The command looks good to me, you won't see command lines populated with the arguments from the config, it is just read directly. By setting this command line flag, we override our own global default and actually **shorten** it from 5m to 3m.

As you can see in the run below, the value of 5m was properly applied before this change:

image

https://github.com/triggermesh/triggermesh/actions/runs/1337126562

PR #179, opened 2021-10-18
Function reconciliation fix

tzununbekov:
In this PR we're replacing `KnSvcEnvFromMap` function that sets env variables from maps "as is" with `KnSvcEnvVars` that accepts an ordered list of items. Having env variables list randomly ordered caused `semantic.DeepEqual` to falsely report the changes in the reconciliation loop.

### DIFF
--- a/pkg/function/function.go
+++ b/pkg/function/function.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"sort"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


antoineco:
Nice catch!

And yes, the order matters because Kubernetes allows you to reference an environment variable from another environment variable, as long it is defined earlier in the `env:` array: https://kubernetes.io/docs/tasks/inject-data-application/define-interdependent-environment-variables/

PR #191, opened 2021-10-20
Reorganize modules inside `go.mod`

antoineco:
I figured we ended up with 3 different `require` blocks inside `go.mod` after merging the old TriggerMesh repos and updating to Go 1.17, but I can't really make sense of the way things are grouped. I think it was just some legacy from `triggermesh/aws-event-sources`, which was the first repo being merged here.

In this PR, I grouped selected modules as follows:
1. Knative + CloudEvents (direct dependencies)
1. ~Kubernetes (direct dependencies)~
1. All other direct dependencies
1. All indirect dependencies (Go 1.17 does that automatically!)

Using `go get` to update a dependency does _not_ affect the way modules are grouped (e.g. `go get knative.dev/eventing@v0.26.1`).

I am willing to accept any suggestion to use a different grouping (or no grouping at all!).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,26 +2,17 @@ module github.com/triggermesh/triggermesh
 
 go 1.17
 
-// Top-level module control over the exact version used for important direct dependencies.
-// https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive
-replace k8s.io/client-go => k8s.io/client-go v0.21.4
-
+// Knative and CloudEvents are the common denominator to all TriggerMesh components.
 require (
 	github.com/cloudevents/sdk-go/v2 v2.4.1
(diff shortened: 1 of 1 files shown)


antoineco:
@cab105 we definitely could.

The only reason why I kept Knative + CloudEvents separate is because they are the common denominator to all our components, and the only ones (almost) I ever update via `go get` on major releases.

I don't have a good reason for keeping `k8s.io` separate, these could as well be grouped with everything else, especially since their versions is dictated by Knative.

The fmt linter would frown upon whitespaces though, and anything that runs [`go mod edit -fmt`](https://golang.org/ref/mod#go-mod-edit) (including `go mod tidy`)  would overwrite those newlines.

PR #203, opened 2021-10-21
Escape user input in HTTP responses to avoid reflected XSS

antoineco:
Addresses a potential vulnerability found by CodeQL.

image

> Directly writing user input (for example, an HTTP request parameter) to an HTTP response without properly sanitizing the input first, allows for a cross-site scripting vulnerability.
>
> This kind of vulnerability is also called _reflected_ cross-site scripting, to distinguish it from other types of cross-site scripting.
>
> **Recommendation**
>
> To guard against cross-site scripting, consider using contextual output encoding/escaping before writing user input to the response, or one of the other solutions that are mentioned in the references.
>
> **References**
>
> -  OWASP: [XSS (Cross Site Scripting) Prevention Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html).
> -  OWASP: [Types of Cross-Site Scripting](https://www.owasp.org/index.php/Types_of_Cross-Site_Scripting).
> -  Wikipedia: [Cross-site scripting](http://en.wikipedia.org/wiki/Cross-site_scripting).
> -  Common Weakness Enumeration: [CWE-79](https://cwe.mitre.org/data/definitions/79.html).
> -  Common Weakness Enumeration: [CWE-116](https://cwe.mitre.org/data/definitions/116.html).

### DIFF
--- a/pkg/sources/adapter/awssnssource/handler/handler.go
+++ b/pkg/sources/adapter/awssnssource/handler/handler.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"html"
 	"io/ioutil"
 	"net/http"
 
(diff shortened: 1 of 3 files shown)


antoineco:
To be clear, this is not as bad as the report makes it look. Both affected services are **webhooks** (respectfully for SNS and Zendesk), meaning none of them is ever accessed through a web browser.

This commit merely implements what's commonly accepted as a best practice in order to make security scanners happy.

PR #243, opened 2021-11-04
Add possibility to configure AWS endpoint in SQS source

### LINKED ISSUE #235: Allow setting the AWS endpoint url for the AWSSQSSource
I'm trying to test out the AWSSQSSource with [localstack](https://github.com/localstack/localstack). For things to work, I think I need to be able to set the AWS endpoint url to something like `http://localhost:4566`. Overriding the endpoint is possible with, for example, the AWS CLI:

```
aws --endpoint-url=http://localhost:4566
```
It seems like it's also possible with the Go AWS SDK, using an endpoint resolver function: https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/endpoints/

However, it seems that overriding the AWS endpoint is not possible with a `AWSSQSSource`. The only related exposed configuration is the `arn`, which is used to configure the region on the client: https://github.com/triggermesh/triggermesh/blob/dcce7c27b6901bcfb8a3a34720605deec51a570d/pkg/sources/adapter/awssqssource/adapter.go#L133 . Then I think the default endpoint is used based on the service (SQS) and the region.

antoineco:
Closes #235 

Allows overriding the default AWS API endpoint (`http://...amazonaws.com`) with a custom endpoint URL, which allows users of this event source to target alternative API-compatible implementations such as ElasticMQ or Localstack.

---

### Demo

**Prerequisite:** Localstack running in the same Kubernetes cluster as the source. (I used their Helm chart)

Create a queue:

```console
$ aws sqs create-queue --queue-name foo --endpoint=http://host.docker.internal:31566
{
    "QueueUrl": "http://localhost:31566/000000000000/foo"
}
```

Create a source instance for this queue:

```yaml
kind: AWSSQSSource
metadata:
  name: sample
spec:
  arn: arn:aws:sqs:eu-central-1:000000000000:foo

  credentials:
    accessKeyID:
      value: admin
    secretAccessKey:
      value: hunter2

  # This was added in this PR
  endpoint:
    url: http://localstack.localstack:4566

  sink:
    ref:
      apiVersion: v1
      kind: Service
      name: event-display
```

Ensure it becomes Ready:

```console
$ kubectl get awssqssource/sample
NAME     READY   REASON   SINK                                             AGE
sample   True             http://event-display.default.svc.cluster.local   19m
```

Send a message:

```console
$ aws sqs send-message --queue-url http://localhost:31566/000000000000/foo --endpoint http://localhost:31566 --message-body test
{
    "MD5OfMessageBody": "098f6bcd4621d373cade4e832627b4f6",
    "MessageId": "3b0db203-468f-f081-83a6-428523f4a028"
}
```

WIN:

```jsonc
☁  cloudevents.Event
Context Attributes,
  specversion: 1.0
  type: com.amazon.sqs.message
  source: arn:aws:sqs:eu-central-1:000000000000:foo
  id: 3b0db203-468f-f081-83a6-428523f4a028
  time: 2021-11-04T15:56:39.3001906Z
  datacontenttype: application/json
Data,
  {
    "Attributes": {
      "ApproximateFirstReceiveTimestamp": "1636041399269",
      "ApproximateReceiveCount": "1",
      "SenderId": "AIDAIT2UOQQY3AUEKVGXU",
      "SentTimestamp": "1636041399268"
    },
    "Body": "test",
    "MD5OfBody": "098f6bcd4621d373cade4e832627b4f6",
    "MD5OfMessageAttributes": null,
    "MessageAttributes": null,
    "MessageId": "3b0db203-468f-f081-83a6-428523f4a028",
    "ReceiptHandle": "tliyexekijnvvxscuaoefvbsmgzydwgskceytntklpgaajkebzemdauxchkoufmwnvfbqbiahpgvrmjqgrxyqjyzsxxalrncsyppejazmtpjikarwysuwrodvgngblagxsagzaqertdpfzfkfozjssnrgdzecuxtvsktjyiabdeibbwxqairvnqtb"
  }
```


pkg/apis/sources/v1alpha1/aws_common_types.go
```diff
@@ -0,0 +1,35 @@
...
+package v1alpha1
+
+import "knative.dev/pkg/apis"
+
+// AWSSecurityCredentials represents a set of AWS security credentials.
+// See https://docs.aws.amazon.com/general/latest/gr/aws-security-credentials.html
+type AWSSecurityCredentials struct {
+	AccessKeyID     ValueFromField `json:"accessKeyID"`
+	SecretAccessKey ValueFromField `json:"secretAccessKey"`
+}
```
antoineco, line 26: I relocated these from `common_types.go` and created a new file `aws_common_types.go`, since we now have several AWS-specific nested types.

pkg/sources/adapter/common/aws_endpoint.go
```diff
@@ -0,0 +1,49 @@
...
+			}
+
+			return endpoints.ResolvedEndpoint{
+				URL:         endpointURL.String(),
+				PartitionID: partition,
+			}, nil
+		}
+
+		return endpoints.DefaultResolver().EndpointFor(service, region, opts...)
+	}
```
antoineco, line 46: If `AWS_ENDPOINT_URL` is set and is a valid URL, we return an endpoint matching that exact URL.

If not, fall back to the default resolver (public AWS endpoint).

antoineco (commented):
Notes to reviewers.

JeffNeff:
Very cool! Nice addition!

JeffNeff (approved):

config/300-awssqssource.yaml
```diff
@@ -118,6 +118,14 @@ spec:
                     oneOf:
                     - required: [value]
                     - required: [valueFromSecret]
+              endpoint:
```
cab105, line 121: With this as an alternate way of specifying the SQS Queue, then should remove the `arn` requirement and add a `oneOf` to encompass arn or endpoint.

cab105 (commented):
Nice. Left one minor nit in the source yaml to be updated.

config/300-awssqssource.yaml
cab105: Are they not the same thing? My understanding at least for the ARN with SQS is that it converts the ARN into the public URL endpoint for handling the API requests and streaming events

Borrowing from the SQS Console:
image

From their docs:
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-making-api-requests.html

In a sense, I can see setting endpoint, and then not really caring about the values for the ARN save for the queue name.

antoineco: You're right, in the public AWS it does, but this implementation is specific to the public AWS cloud.

The URL of a queue needs to be retrieved with a call to `GetQueueUrl` if all you have is its ARN.

In Localstack, the URL of a queue is `localhost:port/accountId/queueName`, and I would assume other solutions have their own URL scheme:

```
$ aws sqs get-queue-url --queue-name foo --endpoint=http://host.docker.internal:31566
{
    "QueueUrl": "http://localhost:31566/000000000000/foo"
}
```

Issue #244, opened 2021-11-04
Support alternative auth methods such as Kube2IAM and IAM Roles for Service Accounts


AceHack:
I would like to use the credentials built into the POD via Kube2IAM instead of cred file

antoineco:
I'm taking a stab at this.

I will only implement the support for IAM roles in the SQS source for the time being. If it's working as expected, we can enable the feature in other AWS sources as well.

The S3 and SNS sources are considered special and won't support this feature, because the controller currently requires static credentials provided by these sources in order to be able to reconcile external resources. We can revisit in the future and think about ways to tackle this specific problem

sebgoa:
@AceHack @bricelalu there is a sample manifest in https://github.com/triggermesh/triggermesh/pull/248 which shows you a solution for using AWS IAM roles.

isaac88:
@antoineco  It'll be really useful if it's supported on other AWS event source like for example AWSS3Source.

FYI it's really works great with the AWSSQSSource 👍  Thanks for that feature.

antoineco:
@isaac88 it's in progress! You'll be happy to see this:
- #783

vashibhavin:
This is the recommended way of authenticating with AWS via Kubernetes. This really should be standard for all AWS Event sources.

sebgoa:
@vashibhavin are you saying that there is an issue with this feature ? This was merged and is being used.

vashibhavin:
Sorry for the confusion. I was testing out the Kinesis source and realized that it wasn't available. GH search for that brought me to this issue. So just wanted to mention that without support for IRSA, this becomes pretty hard to adopt for EKS.

sebgoa:
So let's open another issue to get IRSA support all the AWS sources, especially the Kinesis one. Is Kinesis source the one that you want to use ?

vashibhavin:
Based on https://github.com/triggermesh/triggermesh/pull/798 it looks like Kinesis event source may already support IRSA.

I am just confused between https://github.com/triggermesh/aws-event-sources/blob/main/config/300-awskinesis.yaml and https://github.com/triggermesh/triggermesh/blob/main/config/300-awskinesissource.yaml .

Do you know what the difference between the 2 repositories is? One of them seems to support IRSA and another doesn't.

jmcx:
[...]

PR #245, opened 2021-11-04
Move `credentials` under `auth` in SQS source spec

antoineco:
This is a **breaking change** which prepares the field for #244.

Changes the spec from

```yaml
kind: AWSSQSSource
metadata:
  name: sample
spec:
  arn: arn:aws:sqs:eu-central-1:000000000000:foo

  credentials:
    accessKeyID:
      value: foo
    secretAccessKey:
      value: foo
```

to

```yaml
kind: AWSSQSSource
metadata:
  name: sample
spec:
  arn: arn:aws:sqs:eu-central-1:000000000000:foo

  # NEW - auth modes are nested under 'auth'
  auth:
    credentials:
      accessKeyID:
        value: foo
      secretAccessKey:
        value: foo
```

### Rationale

We want to support multiple modes of authentication in AWS sources moving forward - like we already do in Azure sources, for example - and not only `credentials` (for "AWS security credentials"). Our first objective is to support "IAM Role" authentication.

To enable this, I think it would be clearer if we could expose those different methods under a top-level `auth` object in the spec. Again, this is exactly how it's currently done in Azure sources, so I'm simply trying to remain consistent.

Example:

```yaml
kind: AWSSQSSource
metadata:
  name: sample
spec:
  arn: arn:aws:sqs:eu-central-1:000000000000:foo

  auth:
    iamRole: arn:aws:iam::role-sqs-A
```

### Why not shipping everything in a single PR?

I want to make this update as smooth as possible. By merging the spec change early, I would have time to adjust the frontend, etc. in parallel, without having to also worry about other (maybe heavier) changes to support the new auth mode.

If anyone would prefer me to ship both the API change and the feature in a single PR, I can also do that 👍


config/300-awssqssource.yaml
```diff
@@ -69,55 +69,59 @@ spec:
...
+              auth:
+                description: Authentication method to interact with the Amazon SQS API.
                 type: object
                 properties:
-                  accessKeyID:
-                    description: Access key ID.
+                  credentials:
+                    description: Security credentials authentication. For more information about AWS security
+                      credentials, please refer to the AWS General Reference at
+                      https://docs.aws.amazon.com/general/latest/gr/aws-security-credentials.html.
```
antoineco, line 79: GitHub shows a bunch of changes in this file, but all I did (apart from adding `auth` and a description) was indent everything 2 spaces to the right 🤡

odacremolbap (approved):

PR #246, opened 2021-11-05
Function runtimes update

### LINKED ISSUE #238: Function should not send a response when returned Null
Function should return no event/no payload when the Null value for the language is returned.

Right now this code is producing a new event as a reply:
```yaml
apiVersion: extensions.triggermesh.io/v1alpha1
kind: Function
metadata:
  name: ocimetrics-to-logz
spec:
  runtime: python
  responseIsEvent: true
  public: false
  ceOverrides:
    extensions:
      source: ocimetrics-to-logz
  entrypoint: handleOCImetrics
  code: |
    def handleOCImetrics(event, context):
      return None
```

Function should be able to
- If `Null` is returned ACK should be sent back with default attributes.
- If an event is returned but the `data` element contains the `Null` value, and ACK should be sent back with user attributes.
- If an event is returned with `data` in it (not `Null` but some content or the 0 value), return ACK plus the payload, just as it does now.

tzununbekov:
KLR images update. Function CRD supports empty responses.
Closes #238

### DIFF
--- a/config/500-controller.yaml
+++ b/config/500-controller.yaml
@@ -174,11 +174,11 @@ spec:
           value: ko://github.com/triggermesh/triggermesh/cmd/splitter-adapter
         # Function Runtimes
         - name: RUNTIME_KLR_PYTHON
-          value: gcr.io/triggermesh/knative-lambda-python37:v1.9.1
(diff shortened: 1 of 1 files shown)


odacremolbap (approved):
:balloon: :cake: :pizza:

PR #247, opened 2021-11-05
Add serviceAccountProvider interface to source kinds

antoineco:
Another incremental step towards #244.

Adds a `serviceAccountProvider` micro-interface which allows an instance of a Source kind to signal that it requires its **own** ServiceAccount instead of the default shared ServiceAccount.
E.g., instead of using the shared `foosource-adapter` ServiceAccount, a source instance can implement `WantsOwnServiceAccount()` and return `true`, and the adapter reconciler will sync a dedicated ServiceAccount for that particular object instead.

```console
$ kubectl get serviceaccount
NAME                      SECRETS   AGE
foosource-adapter         1         1m    <- shared
foosource-i-mysource      1         1m    <- object-specific
foosource-i-othersource   1         1m    <- object-specific
```

The immediate goal is to allow AWS sources to support "IAM role" authentication, which is a credential-less authentication mode that is enabled by annotating a ServiceAccount.
The actual implementation of the auth mechanism will follow in another PR, but I already added a dummy `nil` implementation of the required method to `awssqs_lifecycle.go`.

### DIFF
--- a/pkg/apis/sources/v1alpha1/awssqs_lifecycle.go
+++ b/pkg/apis/sources/v1alpha1/awssqs_lifecycle.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v1alpha1
 
 import (
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	"knative.dev/pkg/apis"
@@ -66,4 +67,14 @@ func (s *AWSSQSSource) GetEventTypes() []string {
 // AsEventSource implements EventSource.
 func (s *AWSSQSSource) AsEventSource() string {
 	return s.Spec.ARN.String()
+}
+
+// WantsOwnServiceAccount implements serviceAccountProvider.
+func (*AWSSQSSource) WantsOwnServiceAccount() bool {
+	return false
+}
+
+// ServiceAccountOptions implements serviceAccountProvider.
+func (*AWSSQSSource) ServiceAccountOptions() []func(*corev1.ServiceAccount) {
+	return nil
 }
--- a/pkg/apis/sources/v1alpha1/interfaces.go
+++ b/pkg/apis/sources/v1alpha1/interfaces.go
@@ -19,6 +19,7 @@ package v1alpha1
 import (
 	"context"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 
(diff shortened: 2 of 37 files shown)


pkg/apis/sources/v1alpha1/interfaces.go
```diff
@@ -58,6 +59,31 @@ func IsMultiTenant(src EventSource) bool {
 	return ok && mt.IsMultiTenant()
 }
 
+// serviceAccountProvider is implemented by source types which are able to
+// influence the shape of the ServiceAccount used by their own receive adapter.
+type serviceAccountProvider interface {
+	WantsOwnServiceAccount() bool
+	ServiceAccountOptions() []func(*corev1.ServiceAccount)
+}
```
antoineco, line 67: This is the interface.

It has two methods:
- `WantsOwnServiceAccount()` controls whether or not the object needs its own ServiceAccount or not.
- `ServiceAccountOptions()` can be used to mutate that ServiceAccount if needed (e.g. inject a label or annotation)

pkg/apis/sources/v1alpha1/awssqs_lifecycle.go
```diff
@@ -67,3 +68,13 @@ func (s *AWSSQSSource) GetEventTypes() []string {
...
+
+// WantsOwnServiceAccount implements serviceAccountProvider.
+func (*AWSSQSSource) WantsOwnServiceAccount() bool {
+	return false
+}
+
+// ServiceAccountOptions implements serviceAccountProvider.
+func (*AWSSQSSource) ServiceAccountOptions() []func(*corev1.ServiceAccount) {
+	return nil
+}
```
antoineco, line 80: Here is a dummy implementation of the interface which, for now, doesn't do anything.

In a subsequent PR, that source will
- return `WantsOwnServiceAccount -> true` when the user selects the "IAM role" authentication.
- return a ServiceAccountOptions which sets the `eks.amazonaws.com/role-arn` to the arn of the selected IAM role.

pkg/sources/reconciler/common/reconcile.go
```diff
@@ -53,7 +53,7 @@ var knativeServingAnnotations = []string{
 // RBACOwnersLister returns a list of OwnerRefable to be set as a the
 // OwnerReferences metadata attribute of a ServiceAccount.
 type RBACOwnersLister interface {
-	RBACOwners(namespace string) ([]kmeta.OwnerRefable, error)
+	RBACOwners(src v1alpha1.EventSource) ([]kmeta.OwnerRefable, error)
 }
```
antoineco, line 57: I had to change that interface's signature and propagate the change to _all_ sources, because in case the source sets `WantsOwnServiceAccount -> true`, we want the list of RBAC owners to contain _just_ the current object.

That's why you'll see a lot of tiny changes under `pkg/sources/reconciler`.

antoineco (commented):
Quick tour of the changes.

pkg/apis/sources/v1alpha1/interfaces.go
```diff
@@ -58,6 +59,31 @@ func IsMultiTenant(src EventSource) bool {
...
+	saProvider, ok := src.(serviceAccountProvider)
+	return ok && saProvider.WantsOwnServiceAccount()
+}
+
+// ServiceAccountOptions returns functional options for mutating the
+// ServiceAccount associated with a given source instance.
+func ServiceAccountOptions(src EventSource) []func(*corev1.ServiceAccount) {
+	saProvider, ok := src.(serviceAccountProvider)
+	if !ok {
+		return nil
```
cab105, line 81: Since the return type is bool, shouldn't this be `false`?

cab105 (approved):
Looks nice and clean. Left a very minor nit that isn't a showstopper.

pkg/apis/sources/v1alpha1/interfaces.go
antoineco: On this method the return type is `[]func(*corev1.ServiceAccount)`.
The other method `WantsOwnServiceAccount()` is the one which returns a boolean.

There are two methods because a source may want its own ServiceAccount but not necessarily need to mutate the shape of that ServiceAccount (just to be able to bind it to a custom RBAC role, for example).

cab105: Derp, okay I'll make an appointment to see the optometrist again as my eyes wen towards line 71 instead of 78 for that.

PR #253, opened 2021-11-09
Creates A New Target Scaffolding/Boilerplate Generator  

JeffNeff:
This PR Intends to create a new Target scaffolding/boilerplate generator by means of Go templates.


hack/newTargetGen/main.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+}
+
+func main() {
+	temp := &component{}
+	// var capsName string
+
+	fmt.Print("Enter the target name: ")
+	fmt.Scanf("%s", &temp.UppercaseName)
```
antoineco, line 59: I don't particularly like the fact that it's interactive, because it's hard to test and an interactive shell is required in order to use the tool.

Couldn't we simply require a `--kind` flag and run the generator without requiring the caller to go through another prompt?

hack/newTargetGen/main.go
```diff
@@ -0,0 +1,145 @@
...
+
+	temp.Name = strings.ToLower(temp.UppercaseName)
+	temp.FullCaps = strings.ToUpper(temp.UppercaseName)
+	// fmt.Print("Enter the ALL CAPS VERISON of the target name: ")
+	// fmt.Scanf("%s", &capsName)
+
+	// TODO add naming validation here
```
antoineco, line 66: leftovers

hack/newTargetGen/main.go
```diff
@@ -0,0 +1,145 @@
...
+
+	// make cmd folder
+	path := fmt.Sprintf("cmd/%s", temp.Name+"-adapter")
+	newpath := filepath.Join("../../", path)
+	err := os.MkdirAll(newpath, os.ModePerm)
+	if err != nil {
+		fmt.Println(err)
+	}
+
+	// // make adapter folder
```
antoineco, line 76: There are several `// //` comments in this file. This needs to be cleaned up a little.

hack/newTargetGen/main.go
```diff
@@ -0,0 +1,145 @@
...
+	temp.Name = strings.ToLower(temp.UppercaseName)
+	temp.FullCaps = strings.ToUpper(temp.UppercaseName)
+	// fmt.Print("Enter the ALL CAPS VERISON of the target name: ")
+	// fmt.Scanf("%s", &capsName)
+
+	// TODO add naming validation here
+
+	// make cmd folder
+	path := fmt.Sprintf("cmd/%s", temp.Name+"-adapter")
+	newpath := filepath.Join("../../", path)
```
antoineco, line 70: This is very fragile and assumes the caller is always inside `hack/newTargetGen/` when running the generator.
I think it would be saner to assume the _current directory_ is the base directory, and let the user select an alternative path using a flag.

hack/newTargetGen/main.go
```diff
@@ -0,0 +1,145 @@
...
+
+	temp.Name = strings.ToLower(temp.UppercaseName)
+	temp.FullCaps = strings.ToUpper(temp.UppercaseName)
+	// fmt.Print("Enter the ALL CAPS VERISON of the target name: ")
+	// fmt.Scanf("%s", &capsName)
+
+	// TODO add naming validation here
+
+	// make cmd folder
+	path := fmt.Sprintf("cmd/%s", temp.Name+"-adapter")
```
antoineco, line 69: `Sprintf` feels a bit like an overkill for a string concatenation.
The same result can be achieved without any function call, and passed directory to your path manipulation below: `"cmd/" + temp.Name+"-adapter"`.

hack/newTargetGen/scaffolding/cmd/newtarget-adapter/main.go
```diff
@@ -0,0 +1,26 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"github.com/triggermesh/triggermesh/pkg/targets/adapter/{{.Name}}"
+	pkgadapter "knative.dev/eventing/pkg/adapter/v2"
```
antoineco, line 21: nit: This alias could be removed. There is no clash with another `adapter` import, so you can simply use it as `adapter` (without any aliasing, "/v2" is ignored by Go due to module versioning conventions).

hack/newTargetGen/scaffolding/config/301-newtarget.yaml
```diff
@@ -0,0 +1,123 @@
...
+  versions:
+  - name: v1alpha1
+    served: true
+    storage: true
+    subresources:
+      status: {}
+    schema:
+      openAPIV3Schema:
+        type: object
+        description: TriggerMesh event target for Alibaba Object Storage Service.
```
antoineco, line 52: Nope

I guess you could set this to "TriggerMesh event target for {{.Name}}." which good enough to bootstrap the CRD.

hack/newTargetGen/scaffolding/config/301-newtarget.yaml
```diff
@@ -0,0 +1,123 @@
...
+        properties:
+          spec:
+            description: Desired state of event target.
```
antoineco, line 55: "the" event target.

hack/newTargetGen/scaffolding/config/301-newtarget.yaml
```diff
@@ -0,0 +1,123 @@
...
+            type: object
+            properties:
+
+          status:
+            type: object
+            description: Reported status of the event target.
+            properties:
+              acceptedEventTypes:
+                description: Array of cloud events this target will accept for processing.
```
antoineco, line 64: "CloudEvents"

hack/newTargetGen/main.go
```diff
@@ -0,0 +1,145 @@
...
+	"fmt"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+	"text/template"
+)
+
+type component struct {
+	Name          string
```
antoineco, line 29: This TitleCase "name" is called a _Kind_ in the Kubernetes jargon.

hack/newTargetGen/scaffolding/pkg/apis/targets/v1alpha1/newtarget_types.go
```diff
@@ -0,0 +1,73 @@
...
+// Check the interfaces {{.UppercaseName}} should be implementing.
+var (
+	_ runtime.Object            = (*{{.UppercaseName}})(nil)
+	_ kmeta.OwnerRefable        = (*{{.UppercaseName}})(nil)
+	_ targets.IntegrationTarget = (*{{.UppercaseName}})(nil)
+	_ targets.EventSource       = (*{{.UppercaseName}})(nil)
+	_ duckv1.KRShaped           = (*{{.UppercaseName}})(nil)
+)
+
+// {{.UppercaseName}}Spec holds the desired state of the {{.UppercaseName}}.
```
antoineco, line 50: "of the event target." would be more than enough and less error prone.

hack/newTargetGen/scaffolding/pkg/apis/targets/v1alpha1/newtarget_types.go
```diff
@@ -0,0 +1,73 @@
...
+type {{.UppercaseName}}Spec struct {
+
+	// EventOptions for targets
+	EventOptions *EventOptions `json:"eventOptions,omitempty"`
+}
+
+// {{.UppercaseName}}Status communicates the observed state of the {{.UppercaseName}} (from the controller).
```
antoineco, line 57: "of the event target." would be more than enough and less error prone.

hack/newTargetGen/scaffolding/pkg/apis/targets/v1alpha1/newtarget_types.go
```diff
@@ -0,0 +1,73 @@
...
+	duckv1.Status        `json:",inline"`
+	duckv1.AddressStatus `json:",inline"`
+
+	// Accepted/emitted CloudEvent attributes
+	CloudEventStatus `json:",inline"`
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// {{.UppercaseName}}List is a list of {{.UppercaseName}} resources
```
antoineco, line 68: "is a list of event target instances." would be more than enough and less error prone.

hack/newTargetGen/scaffolding/pkg/targets/reconciler/newtarget/adapter.go
```diff
@@ -0,0 +1,83 @@
...
+	envEventsPayloadPolicy = "EVENTS_PAYLOAD_POLICY"
+)
+
+// adapterConfig contains properties used to configure the target's adapter.
+// Public fields are automatically populated by envconfig.
+type adapterConfig struct {
+	// Configuration accessor for logging/metrics/tracing
+	obsConfig source.ConfigAccessor
+	// Container image
+	Image string `envconfig:"{{.FullCaps}}_ADAPTER_IMAGE" default:"gcr.io/triggermesh-private/{{.Name}}-adapter"`
```
antoineco, line 43: We don't use "-private" anymore.

hack/newTargetGen/main.go
```diff
@@ -0,0 +1,145 @@
...
+	UppercaseName string
```
antoineco, line 30: I think you meant TitleCase, not UPPERCASE.
And it's called a _Kind_ in the Kubernetes jargon.

antoineco (commented):
A bit rough still, but the switch to templates is a great improvement IMO! 👏

hack/newTargetGen/main.go
JeffNeff: could i have a little help on this one? Im not exactly positive how to do this :/

hack/newTargetGen/main.go
```diff
@@ -0,0 +1,145 @@
...
+	// fmt.Scanf("%s", &capsName)
+
+	// TODO add naming validation here
+
+	// make cmd folder
+	path := fmt.Sprintf("cmd/%s", temp.Name+"-adapter")
+	newpath := filepath.Join("../../", path)
+	err := os.MkdirAll(newpath, os.ModePerm)
+	if err != nil {
+		fmt.Println(err)
```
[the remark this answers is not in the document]
JeffNeff: The error handling throughout is horrendous, I agree! Ill revisit this.

JeffNeff (commented):
I anciently entered a review mode i dont know how to get out of lol.

hack/newTargetGen/main.go
antoineco: It's easier than your imagine :)

I'm sharing some snippets below, but in a nutshell you simply need to use `./` instead of `../../` (and ideally allow users to override that value if they run from a directory which is not the root of the repo).

1. Set a flag which default value is `./` (always represents the current working directory)

 
   https://github.com/triggermesh/triggermesh/blob/a490818b065d34b7528584936eb20f25b9ad0738/hack/rbac-check/main.go#L192-L193

1. Sanitize the user input (just a good practice)

 
   https://github.com/triggermesh/triggermesh/blob/a490818b065d34b7528584936eb20f25b9ad0738/hack/rbac-check/main.go#L199

1. Compose all the paths as a join of this base path and the target directory (like you already did)

    https://github.com/triggermesh/triggermesh/blob/a490818b065d34b7528584936eb20f25b9ad0738/hack/rbac-check/main.go#L96

hack/targetgen/main.go
```diff
@@ -0,0 +1,183 @@
...
+		return err
+	}
+	defer file.Close()
+
+	err = tmp.Execute(file, a)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
antoineco, line 182: Since this is the last function call, and it returns what you need, you can return the result directly:

```suggestion
	return tmp.Execute(file, a)
```

hack/targetgen/scaffolding/config/301-newtarget.yaml
```diff
@@ -0,0 +1,122 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: {{.Kind}}s.targets.triggermesh.io
```
antoineco, line 18: I didn't realize before reading this but `Kind` and `TitleCase` seem to be used wrongly across the code.
Ref. to my previous comment: https://github.com/triggermesh/triggermesh/pull/253#discussion_r745583399

Kind is `MyCustomKind` as per Kubernetes definition (check the `kind:` field in any YAML).

We should use proper meaning for those things.
`Kind` is be what you currently call `TitleCase` (and what the user passes using a flag).

hack/targetgen/scaffolding/pkg/apis/targets/v1alpha1/newtarget_lifecycle.go
```diff
@@ -0,0 +1,113 @@
...
+// GetEventTypes implements EventSource.
+func (*{{.TitleCase}}) GetEventTypes() []string {
+	return []string{
+		EventType{{.TitleCase}}GenericResponse,
+	}
+}
+
+// AsEventSource implements targets.EventSource.
+func (s *{{.TitleCase}}) AsEventSource() string {
+	return "https://" + "SOMETHINGUSEFULE"
```
antoineco, line 48: "USEFULE"?

antoineco (approved):

PR #256, opened 2021-11-10
Use dedicated SDK for Azure Service Bus

### LINKED ISSUE #249: Switch to Azure SDKs for Event Hubs and Service Bus in source reconcilers
Reconcilers for Azure event sources (control plane) currently use the [Azure SDK for Go](https://github.com/Azure/azure-sdk-for-go), whereas the adapters (data plane) use the dedicated [Event Hubs client](https://github.com/Azure/azure-event-hubs-go) and [Service Bus client](https://github.com/Azure/azure-service-bus-go).

Related packages: [pkg/sources/client/azure](https://github.com/triggermesh/triggermesh/tree/c13be0932d4608455edf31cfb6da07ce96ba492f/pkg/sources/client/azure)

There is no true justification for using the Azure SDK in those cases, because the dedicated clients are able to deal with **both the management and data APIs**.
Not only this leads to some inconsistent usage of different clients across the management and control plane, but the Azure SDK supports **only [AAD](https://docs.microsoft.com/en-us/azure/active-directory/fundamentals/active-directory-whatis) authentication** (e.g. using a Service Principal), whereas the dedicated clients support **both AAD and [SAS](https://docs.microsoft.com/en-us/azure/storage/common/storage-sas-overview) authentication** (e.g. in a Connection String).

By switching to the dedicated clients (wherever possible), we could more easily offer support for SAS authentication, as an alternative to AAD.

antoineco:
Closes #249

The main goal is to support multiple authentication modes: **Azure Active Directory** (Service Principal) _and_ **SAS Token** (e.g. connection string).

(The old autorest-based client supports only the former).

I will update the specs of our two Service Bus sources (Queue and Topic) in a follow-up PR.

---

There are still a few TO-DOs to a ddressbefore I can mark this as ready:
- [ ] Update error handling, because the errors returned by this SDK are different
- [ ] Update tests


pkg/sources/client/azure/servicebus/client.go
```diff
@@ -0,0 +1,145 @@
...
+			//   api-version=1.0  }} 00000000-0000-0000-0000-000000000000 https://servicebus.azure.net/ true
+			//   300000000000} 0xc001f59380 0xc0029ac8a0 <nil> [] 0}"
+			//
+			// This error contains unique elements such as pointer addresses, and therefore needs to be
+			// sanitized to avoid causing an infinite loop of reconciliations when written to an API
+			// object's status.
+			if strings.Contains(err.Error(), "failed to refersh token") { // (!) the typo is not a mistake
+				err = errors.New("failed to refresh Service Principal token. The provided secret is " +
+					"either invalid or expired")
+			}
```
antoineco, line 104: This whole error handling based on string parsing is disgusting and won't be maitainable over time.
We need a better way to figure what actually happened when an error occurs, and this SDK isn't mature enough to provide us that answer.

pkg/sources/reconciler/azureservicebustopicsource/subscription.go
```diff
@@ -280,23 +195,16 @@ func recursErrMsg(errMsg string, err error) string {
...
-	if dErr := (autorest.DetailedError{}); errors.As(err, &dErr) {
-		if code, ok := dErr.StatusCode.(int); ok {
-			return code == http.StatusUnauthorized || code == http.StatusForbidden
-		}
-	}
-
-	return false
+	// FIXME(antoineco): the SDK returns unstructured errors
+	// request failed: 401 SubCode=40100: Unauthorized : Unauthorized access for 'DeleteSubscription' operation on endpoint 'sb://triggermesh-dev.servicebus.windows.net/triggermesh-dev/subscriptions/io.triggermesh.azureservicebussources-4197625275?api-version=2017-04'. Tracking Id: bbbccdfa-9277-407f-891a-18a59fa82c7c_G20
+	return strings.Contains(err.Error(), "Unauthorized access")
```
antoineco, line 207: Another occurrence terrible error handling based on string parsing.
This time for "Unauthorized" errors, which our logic expects to be handled properly.

pkg/sources/reconciler/azureservicebustopicsource/subscription.go
```diff
@@ -280,23 +195,16 @@ func recursErrMsg(errMsg string, err error) string {
 // isNotFound returns whether the given error indicates that some Azure
 // resource was not found.
 func isNotFound(err error) bool {
-	if dErr := (autorest.DetailedError{}); errors.As(err, &dErr) {
-		return dErr.StatusCode == http.StatusNotFound
-	}
-	return false
+	return errors.Is(err, servicebussdk.ErrNotFound{})
```
antoineco, line 198: NotFound error are also a pain. This code doesn't work because I should have used `As()` instead of `Is()`, but my real concern here is that even upstream is [hacking its way around those errors](https://github.com/Azure/azure-service-bus-go/blob/d0f6cc33729ca31fb8c75523f9bef00adb175e0a/subscription_manager.go#L340-L344).

antoineco (commented):
Despite all the work I put into this branch, I am going to close this pull request for the reasons mentioned below.

The SDK for Service Bus is simply too immature for us to consider. Migrating to it would be a giant step back for us in terms of error handling, and we could only use it in **one** source anyway (Service Bus Topics), because management APIs don't support SAS authentication in general.

PR #257, opened 2021-11-11
Creates an Event Hubs Target

JeffNeff:
I wanted to try and keep the Target as similar to the Source as possible. To do this, I was faced to make some decisions on moving or sharing some reconciliation logic. I _think_ the proper way would be to create a `common` folder for both the Targets and Sources reconcilers.. but I did not want to make this decision ( and that's probably outside the scope of this PR.) So I ended up referencing some resources within Sources/reconciler/common. 

I DID, however, add `pkg/targets/reconciler/common/adapter.go` So that I could have `MaybeAppendValueFromEnvVar` and also updated the `pkg/targets/reconciler/resources/kservice.go` to include `EnvVar`.

 I made these updates to keep the reconcilers looking as close to each other as possible but also because I think that these methods are pretty and I want us to be able to use them for building Targets moving forward.
 
If we are not happy with these changes, I'm glad to revert these changes and re-write the reconciler the way we normally write the Target's


config/samples/targets/azureeventhubstarget/azureeventhubstarget.yaml
```diff
@@ -0,0 +1,42 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Sample AzureEventHubSource object.
+#
+# For a list and description of all available attributes, execute the following command against a cluster where this
+# Custom Resource Definition has been registered:
+#
+#   kubectl explain azureeventhubtargets.targets.triggermesh.io
```
antoineco, line 20:
```suggestion
#   kubectl explain azureeventhubstargets.targets.triggermesh.io
```

go.mod
```diff
@@ -21,6 +21,7 @@ require (
 	cloud.google.com/go/pubsub v1.17.1
 	cloud.google.com/go/storage v1.18.2
 	cloud.google.com/go/workflows v1.0.0
+	github.com/Azure/azure-event-hubs-go v1.3.1
 	github.com/Azure/azure-event-hubs-go/v3 v3.3.16
```
antoineco, line 25: You must have an import somewhere that is missing `/v3/`, that's why it pulled this old version.

pkg/targets/adapter/azureeventhubstarget/adapter.go
```diff
@@ -0,0 +1,92 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package azureeventhubstarget
+
+import (
+	"context"
+
+	eventhub "github.com/Azure/azure-event-hubs-go"
```
antoineco, line 22: There it is.

```suggestion
	eventhub "github.com/Azure/azure-event-hubs-go/v3"
```

antoineco (commented):
Just a comment on that import. Let me know when it can be reviewed.

JeffNeff:
Some of the failing RBAC tests i understand, and will address. But some of them... I do not. The first one for example 
```
[triggermesh-controller] in node "resources" at line 65: (-:expect, +:got)
  (
  	"""
  	... // 13 identical lines
  	- azureeventgridsources
  	- azureeventhubsources
+ 	- azureeventhubstargets
  	- azureiothubsources
  	- azurequeuestoragesources
  	... // 16 identical lines
  	"""
  )
  ```
  
it wants me to add in this target with the sources?

**Comment by antoineco** [comment]
@JeffNeff is your CRD file starting with 300 or 301?
You created the naming scheme :)

**Comment by JeffNeff** [comment]
yea.. copy pasta got me hard there. good catch.

File: pkg/apis/targets/v1alpha1/azure_common_types.go [hunk 4]
```diff
@@ -0,0 +1,407 @@
+/*
```
antoineco, line 1: It would be nice to also copy the tests in `azure_common_types_test.go` (since this code was copied from sources), because it's quite essential to ensure the deserialization from string (in Kubernetes) to structs (in this Go code) is working as intended.
By having tests, we can more confidently change this code in the future.

pkg/targets/adapter/azureeventhubstarget/config.go
```diff
@@ -0,0 +1,42 @@
...
+	BridgeIdentifier string `envconfig:"EVENTS_BRIDGE_IDENTIFIER"`
+
+	// CloudEvents responses parametrization
+	CloudEventPayloadPolicy string `envconfig:"EVENTS_PAYLOAD_POLICY" default:"error"`
+
+	// DiscardCEContext chooses to keep or discard the incoming cloudevent context
+	DiscardCEContext bool `envconfig:"DISCARD_CE_CONTEXT"`
+
+	// AzureEventHub connection parameters
+	ConnectionString string `envconfig:"EVENTHUB_CONNECTION_STRING"`
```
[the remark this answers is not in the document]
JeffNeff: Sorry this is leftover from my inital revision of the Target. Before I introduced the `Azure Auth` thingy.

JeffNeff:
Thank you for your review @antoineco !!!

antoineco (approved):

PR #259, opened 2021-11-12
Several improvements to Azure packages for sources

antoineco:
Port the few bits which I found interesting from #256, without the migration to the new SDK (which is not mature enough for us to use, see PR).

- Move `sources/azure` to `sources/auth/azure`
- Move Permanent error behaviour to `sources/auth` package
- Rename `azure.Authorizer()` to `azure.NewAADAuthorizer()`
- Remove Subscription comparison from Service Bus Topics source

### DIFF
--- a/pkg/apis/sources/v1alpha1/azureservicebustopic_lifecycle.go
+++ b/pkg/apis/sources/v1alpha1/azureservicebustopic_lifecycle.go
@@ -72,24 +72,24 @@ func (*AzureServiceBusTopicSource) GetEventTypes() []string {
 
 // Status conditions
 const (
-	// AzureServiceBusTopicSubscribed has status True when the source has subscribed to a topic.
-	AzureServiceBusTopicSubscribed apis.ConditionType = "Subscribed"
+	// AzureServiceBusTopicConditionSubscribed has status True when the source has subscribed to a topic.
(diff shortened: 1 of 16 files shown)


odacremolbap (approved):
nice.

In the future we could evolve to create a global Credentials error (not only for Azure). When that happens and other devs implement their own errors for the interface I will probably add some concerns on the `Fatal` meaning to make it more `credentials are not valid, retrying over and over could potentially lead to a leaking resource without human intervention`. I don't have a better alternative for that name right now.

Pls merge, we can talk about that name if we extract the creds error to the non Azure world.

:shipit:

antoineco:
I agree. As these sources evolve, I'm seeing more and more utilities that can be extracted and shared, both in the API lifecycles and authenticated clients.

Expect some follow up 😉

antoineco:
@odacremolbap before I merge, would `auth.PermanentCredentialsError` sound better than `auth.Fatal...`?
It would be a quick change.

odacremolbap:
> @odacremolbap before I merge, would `auth.PermanentCredentialsError` sound better than `auth.Fatal...`? It would be a quick change.

I think so (not sure if this is too much my opinion).

antoineco:
I like it too, changing.

PR #260, opened 2021-11-13
Merge adapters of Azure Service Bus sources

### LINKED ISSUE #240: Merge adapters of Azure Service Bus sources
Service Bus supports [both messaging Queues and Topics](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-messaging-overview), and we currently have a source for each of them with their own respective receive adapter.

But because both use the same SDK and underlying protocol (AMQP), I believe it should be possible, fairly easily, to merge the two receive adapters.

For example, the `adapter` struct from the Service Bus Topic source contains a generic [`servicebus.Receiver`](https://pkg.go.dev/github.com/Azure/azure-service-bus-go#Receiver) that can be used to consume messages from either a Queue or a Topic:

https://github.com/triggermesh/triggermesh/blob/c13be0932d4608455edf31cfb6da07ce96ba492f/pkg/sources/adapter/azureservicebustopicsource/adapter.go#L56-L64

There is one unique [`*servicebus.Namespace.NewReceiver`](https://pkg.go.dev/github.com/Azure/azure-service-bus-go#Namespace.NewReceiver) method available to obtain such `Receiver`, and it does the right thing based on the "entity path" passed as argument (`name` for a Queue, `topicName/Subscriptions/subscriptionName` for a Topic).

https://github.com/triggermesh/triggermesh/blob/c13be0932d4608455edf31cfb6da07ce96ba492f/pkg/sources/adapter/azureservicebustopicsource/adapter.go#L91-L95

The identity of the resource to consume messages from is passed as a generic Azure Resource ID and, although the name of the corresponding environment variable currently contains the word "subscription", we could easily rename it:

https://github.com/triggermesh/triggermesh/blob/c13be0932d4608455edf31cfb6da07ce96ba492f/pkg/sources/adapter/azureservicebustopicsource/adapter.go#L46-L47

### LINKED ISSUE #261: Source receives always the same message at several minutes of interval
Messages aren't marked as complete with `msg.Complete()` after they have been processed, so the behaviour explained in the title can be observed in both these sources.

https://github.com/triggermesh/triggermesh/blob/c123453683ad9967f4900bbb318df9d789aeeedc/pkg/sources/adapter/azureservicebusqueuesource/adapter.go#L117-L125

https://github.com/triggermesh/triggermesh/blob/c123453683ad9967f4900bbb318df9d789aeeedc/pkg/sources/adapter/azureservicebustopicsource/adapter.go#L171-L177

A simple fix would be to return from the handler with `return msg.Complete(ctx)`.

antoineco:
Closes #240
Fixes #261

Both sources for Azure Service Bus (queues + topics) use the exact same AMQP client to listen to messages, so I merged those two adapters into one.

- I verified that it was still possible to receive messages from either a Queue or a Topic ✔️
- I verified that both SAS Token and AAD (Service Principal) authentication modes were working ✔️

As a bonus:

- I added a JSON schema for events of type `com.microsoft.azure.servicebus.message` and checked some test payloads against it using https://www.jsonschemavalidator.net/
- I enabled AAD (Service Principal) as an alternative authentication mode in the Service Bus Queue source, since the merged adapter now supports both.


config/300-azureservicebusqueuesource.yaml
```diff
@@ -52,19 +52,147 @@ spec:
             description: Desired state of the event source.
             type: object
             properties:
+              queueID:
```
odacremolbap, line 55: Might `queueID` be missing from the required fields?

odacremolbap (approved):

config/300-azureservicebusqueuesource.yaml
antoineco: Nice catch

pkg/sources/adapter/azureservicebussource/adapter.go
```diff
@@ -111,33 +122,104 @@ func NewAdapter(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClie
...
+	// if a key is set explicitly, it takes precedence and is used to
+	// compose a new connection string
+	if keyName, keyValue := os.Getenv(envKeyName), os.Getenv(envKeyValue); keyName != "" || keyValue != "" {
+		azureEnv := &azure.PublicCloud
+		connStr = fmt.Sprintf("Endpoint=sb://%s.%s;SharedAccessKeyName=%s;SharedAccessKey=%s;EntityPath=%s",
+			namespace, azureEnv.ServiceBusEndpointSuffix, keyName, keyValue, entityPath)
+	}
+
+	return connStr
+}
```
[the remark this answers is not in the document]
antoineco: We can and I initially wanted to do it that way, but later had a feeling this might introduce some confusion, because the SDK itself accesses the env vars directly using `os.Getenv()`.

So for AAD auth, when we call `servicebus.NamespaceWithEnvironmentBinding()`, the function reads env vars one way, but for SAS auth, the function reads them a different way.

But I think you're right, the readability of our code shouldn't from using one SDK or another. I'll add the env vars to envconfig with a comment that explains exactly how they are being used.

Issue #32, opened 2021-04-08
Allow fan-out over SNS topic


antoineco:
### Problem

Currently, there is no possibility for _multiple_ subscribers to create a notification configuration in the _same_ bucket. AWS rejects the request with the error `Configurations on the same bucket cannot share a common event type.`.

Even if we enforced a mechanism such as allowing multiple subscribers only if they define non-overlapping event types, we would be limited by the feature of the SQS queue, which doesn't allow fanning messages out to multiple consumers.

### Proposal

We could enable an alternative mode of notification delivery where a user could select a **SNS topic** (which must exist) instead of a SQS queue, and the source would then create and subscribe one SQS queue to that topic, for each AWSS3Source instance which uses that mode.

Unlike the current SQS-based mode of delivery, external subscribers would now be able to consume bucket notifications as well (fan out).

### Possible API change

**edit**: API change implemented in triggermesh/triggermesh#269

```yaml
apiVersion: sources.triggermesh.io/v1alpha1
kind: AWSS3Source

# ...

spec:

  # ...

  destination:  # Optional. Default to 'sqs' with managed queue.

    # Only *one* possible choice per AWSS3Source instance:

    sqs:
      queueARN: arn:aws:sqs:us-west-2:123456789012:triggermeshtest  # Optional. Managed by the source if not provided.

    sns:
      topicARN: arn:aws:sns:us-west-2:123456789012:triggermeshtest
```

### External links

- [SNS - Fanout to Amazon SQS queues](https://docs.aws.amazon.com/sns/latest/dg/sns-sqs-as-subscriber.html)

PR #269, opened 2021-11-14
Add nested destination object to Amazon S3 source spec

antoineco:
Implements the API change that is required to enable #32.

From:

```yaml
spec:
  # ...
  queueARN: arn:aws:sqs:us-west-2:123456789012:triggermeshtest  # Optional. Managed by the source if not provided.
```

To:

```yaml
spec:
  # ...
  destination:  # Optional. Default to 'sqs' with managed queue.
    sqs:
      queueARN: arn:aws:sqs:us-west-2:123456789012:triggermeshtest  # Optional. Managed by the source if not provided.
```


pkg/apis/sources/v1alpha1/awss3_types.go
```diff
@@ -65,19 +65,29 @@ type AWSS3SourceSpec struct {
...
-	//
-	// When not provided, a SQS queue for receiving event notifications
-	// from the S3 bucket is automatically created and configured.
+	// The destination of notifications originating from the Amazon S3 bucket.
 	//
-	// +optional
-	QueueARN *apis.ARN `json:"queueARN,omitempty"`
+	// If omitted, a SQS queue is automatically created and associated with
+	// the bucket.
+	Destination *AWSS3SourceDestination `json:"destination,omitempty"`
```
cab105, line 72: Shouldn't the `// +optional` annotation be included here as well?

cab105 (approved):
Save for a minor annotation nit, lgtm.

pkg/apis/sources/v1alpha1/awss3_types.go
antoineco: Nice catch!

Issue #268, opened 2021-11-14
Switch from Event Hubs to Service Bus as event destination for Azure events


antoineco:
### Context

Two of our Azure sources rely on Event Grid to subscribe to events: the _Blob Storage source_ and the _Event Grid source_ itself.

These two currently use **Event Hubs** as event destination (Event Grid _endpoint_), and our Event Hubs adapter picks up from there.

### Proposal

I suggest we switch the default event destination to **Service Bus**.
At the end of the day this should be considered an implementation detail, so why?

 - Service Bus is both cheaper and easier to integrate with. It provides familiar queuing and pub/sub primitives where a message is delivered once to subscribers, then forgotten.

- Event Hubs, on the other end, is similar to Kafka. It is focused on message streaming and data is persisted by default. Clients are expected to deal with partitions, offsets, etc.

I believe Service Bus is simply a better fit for sourcing events into TriggerMesh. Once such event has been consumed and injected into the platform, it is unlikely that someone cares about persisting it into Azure. And if they do, it should be done with intention, so we could maybe offer the choice to keep Event Hubs as an option.

### Possible API change

Current API, for reference: [samples/sources/azureeventgridsource.yaml](https://github.com/triggermesh/triggermesh/blob/e79d151b57c6931083ed9d45c6bcc8095b2b18f4/config/samples/sources/azureeventgridsource.yaml#L32-L34)

In order to make it clear what the endpoint-related settings are, we could introduce a new `endpoint` sub-level to the spec:

```yaml
apiVersion: sources.triggermesh.io/v1alpha1
kind: AzureEventGridSource

# ...

spec:

  # ...

  endpoint:  # Optional. Default to 'serviceBus' with managed queue.

    # Only *one* possible choice per AzureEventGridSource instance:

    serviceBus:
      namespaceID: /subscriptions/[...]/resourceGroups/MyGroup/providers/Microsoft.ServiceBus/namespaces/MyNamespace
      queueName: MyQueue  # Optional. Managed by the source if not provided.

    eventHubs:
      namespaceID: /subscriptions/[...]/resourceGroups/MyGroup/providers/Microsoft.EventHub/namespaces/MyNamespace
      eventHubName: MyHub  # Optional. Managed by the source if not provided.
```

### External links

Pricing
- [Event Hubs](https://azure.microsoft.com/en-us/pricing/details/event-hubs/)
- [Service Bus](https://azure.microsoft.com/en-us/pricing/details/service-bus/)

PR #274, opened 2021-11-15
Add nested endpoint object to spec of Event Grid sources

antoineco:
Implements the API change that is required to enable #268.

From:

```yaml
spec:
  # ...
  eventHubID: /subscriptions/[...]/resourceGroups/MyGroup/providers/Microsoft.EventHub/namespaces/MyNamespace/...
```

To:

```yaml
spec:
  # ...
  endpoint:
    eventHubs:
      namespaceID: /subscriptions/[...]/resourceGroups/MyGroup/providers/Microsoft.EventHub/namespaces/MyNamespace
      eventHubName: MyHub  # Optional. Managed by the source if not provided.
```


pkg/apis/sources/v1alpha1/azureblobstorage_types.go
```diff
@@ -97,7 +88,7 @@ type AzureBlobStorageSourceStatus struct {
 
 	// Resource ID of the Event Hubs instance that is currently receiving
 	// events from the Azure Event Grid subscription.
-	EventHubID *EventHubResourceID `json:"eventHubID,omitempty"`
```
cab105, line 100: Should there be a `+optional` annotation?

pkg/apis/sources/v1alpha1/azureeventgrid_types.go
```diff
@@ -95,7 +106,7 @@ type AzureEventGridSourceStatus struct {
 
 	// Resource ID of the Event Hubs instance that is currently receiving
 	// events from the Azure Event Grid subscription.
-	EventHubID *EventHubResourceID `json:"eventHubID,omitempty"`
```
cab105, line 98: May want to have a `+optional` annotation here as well.

config/300-azureblobstoragesource.yaml
```diff
@@ -110,6 +98,31 @@ spec:
                   # Allowed for completeness, but users should be aware that those events use a different schema from
                   # the other Storage event types.
                   - Microsoft.Storage.BlobInventoryPolicyCompleted
+              endpoint:
+                description: The destination of events subscribed via Event Grid.
+                type: object
+                properties:
+                  eventHubs:
```
cab105, line 105: IIRC, Azure does support multiple event hubs within an eventhub namespace, but given the use case here, shouldn't it be singular?

cab105 (commented):
Left a couple of minor nits. Will hold off on your take with regard to singular/plural form of eventhub usage to bless the change.

pkg/apis/sources/v1alpha1/azureeventgrid_types.go
antoineco: No because this field was _removed_ 🙂 
And if I'm not mistaken the "+optional" convention applies only to Spec fields, because Status fields are read-only from a user's perspective.

cab105: Understood on the status portion. The placement of this comment was with respect to the addition of `AzureResourceID` and not the removal of `EventHubResourceID`.

config/300-azureblobstoragesource.yaml
antoineco: I never know whether this should be singular or plural!

The intended meaning here was `destination: <name of an Azure service>`.
Azure refers to their _service_ as "Event Hubs" (plural). A significant example from their docs

> An Event Hubs namespace is a management container ...
>
> -- https://docs.microsoft.com/en-us/azure/event-hubs/event-hubs-features

They seem to use "an Event Hub" to refer to an instance and not to the _service_.

I didn't have a strong opinion while coming up with this proposal, but now that I think about it, if I change the meaning to `destination: <type of instance>`, I will have to use `serviceBusQueue` and `serviceBusTopic` when I add support for Service Bus. But my original plan was to implement it as:

```yaml
destination:
  serviceBus:
    # stuff related to Service Bus: namespace, queue, topic, ...
```

Anyway, let me know if the intention is more clear after this explanation, and most importantly: whether you think I should revisit my original intention.

cab105: Yeah it does seem like they use `Event Hubs` for the name of the technology and to encompass the namespace, but use `Event Hub` for the specific component. While looking at that, I thought destination was pointing to an instance of the service and not the service name itself.

I see where you're going with it, and it does make sense so I'll retract my objection.

cab105 (approved):

pkg/apis/sources/v1alpha1/azureeventgrid_types.go
antoineco: I realized that only after and edited my comment, sorry for the misunderstanding!

config/300-azureblobstoragesource.yaml
antoineco: To provide more context, the spec of this source is meant to echo the direction I took with the Amazon S3 spec in #269, where `destination` is either `sqs` or `sns` (name of _services_) and the attributes related to the _instances_ within those services are nested (ARN, optional configs, API endpoints, etc.)

As long as things are consistent between specs of different source (current focus is on AWS, Azure and Google Cloud) and feel right for everyone, I'm OK with amending the original proposal.
We are exploring how to evolve the specs anyway, so no harm mentioning it when something feels off. It's still early enough to revisit!

PR #282, opened 2021-11-17
Remove Slack badge

antoineco:
The invitation link has expired, which makes this badge more misleading than anything else.

I suggest we remove it for now, and revert this commit once we have a new link to share.


sameersbn:
should we update the invite link instead? there's no other way of having people join the slack community. this is the [new invite link](https://join.slack.com/t/triggermesh-community/shared_invite/zt-yaydoq0y-ULniSgpqndaUa90wPsdD5g) (expires in 18 days)

antoineco:
Updating for now, but 18 days is really short. Without automation, this badge will expire constantly, and it's never a good look to show something that's expired on a front page.

PR #285, opened 2021-11-18
Add Azure Event Hubs target to kustomization

FranBarrera:
Azure Event Hubs Target was missing in kustomization.yaml


JeffNeff (approved):
lgtm. thank you!

PR #291, opened 2021-11-19
Move 'credentials' under 'auth' in spec of AWS sources

### LINKED ISSUE #290: Bring auth spec of AWS sources in alignment with the AWS SQS source
In #245, we performed a change to the spec of the `AWSSQSSource` type which allowed us to support more than one method of authentication (IAM roles in this case).

Although all AWS sources except SQS still only support the `credentials` mode (AWS security credentials, a.k.a.  key ID + secret key), we should update their API for consistency.

antoineco:
Closes #290

Big boring PR, identical to #245 but mirrored to _all_ AWS sources to remain consistent across the board.

tl;dr; Changes the spec from

```yaml
kind: AWSInfinidashSource
metadata:
  name: sample
spec:
  arn: arn:aws:infinidash:eu-central-1:000000000000:foo

  credentials:
    accessKeyID:
      value: foo
    secretAccessKey:
      value: foo
```

to

```yaml
kind: AWSInfinidashSource
metadata:
  name: sample
spec:
  arn: arn:aws:infinidash:eu-central-1:000000000000:foo

  # NEW - auth modes are nested under 'auth'
  auth:
    credentials:
      accessKeyID:
        value: foo
      secretAccessKey:
        value: foo
```


pkg/sources/reconciler/common/adapter.go
```diff
@@ -278,53 +278,13 @@ func MakeAWSAuthEnvVars(auth v1alpha1.AWSAuth) []corev1.EnvVar {
...
+		authEnvVars = MaybeAppendValueFromEnvVar(authEnvVars, EnvAccessKeyID, creds.AccessKeyID)
+		authEnvVars = MaybeAppendValueFromEnvVar(authEnvVars, EnvSecretAccessKey, creds.SecretAccessKey)
 	}
 
 	return authEnvVars
 }
 
-// MakeSecurityCredentialsEnvVars returns environment variables for the given
-// AWS security credentials.
-func MakeSecurityCredentialsEnvVars(creds v1alpha1.AWSSecurityCredentials) []corev1.EnvVar {
```
antoineco, line 289: This helper is now totally obsolete, so I removed it.

antoineco (commented):
Notes

antoineco:
Merging because:

- Code already reviewed and approved before (it's an exact copy-pasta)
- No new feature / new logic, just moving one field one level down so close to 0 chance to introduce issues or
- We shouldn't end up with a release where APIs are inconsistent, and this is the very last piece in the "refactor" AWS and Azure specs puzzle (for now at least)

PR #294, opened 2021-11-22
Targetgen's `main.go` refactoring

tzununbekov:
I had the pleasure to use targetgen to generate a new target that I'm working on right now. This PR introduces a bit of `main.go` refactoring:

1. Moved paths to the variables at the top of the file so it would be easier to read and update them if needed,
2. Replaced strings concatenations with `filepath.Join` because it does a better job in paths creation and the code looks a bit cleaner this way,
3. Removed unused code, added a missing error check.

### DIFF
--- a/hack/targetgen/main.go
+++ b/hack/targetgen/main.go
@@ -27,125 +27,128 @@ import (
 	"text/template"
 )
 
+const (
+	apiVersion = "v1alpha1"
+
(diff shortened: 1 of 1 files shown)


JeffNeff (approved):
LGTM!! thank so much for trying out the target gen!! legend

PR #296, opened 2021-11-22
Implements the Package Replier for the AWS DynamoDB Target

JeffNeff:
Moving this Target to utilize the Package replier.


antoineco:
@odacremolbap this PR has been open for a while but hasn't gotten attention, and it is not productive to leave it in that state.

Are there concerns with the way things are implemented, or do you think we shouldn't do this now?

odacremolbap:
Thanks for bringing this up, I'll take a look (unless @JeffNeff had a good reason to close this).

PR #300, opened 2021-11-23
Insert the requesting HTTP method and route in the CE metadata

JeffNeff:
This PR updates webhook source to add the incoming request method to the `method` CE extension. and the requesting path in the `route` CE extension. This change is being made In order to route events based upon the incoming request method and route

### DIFF
--- a/pkg/sources/adapter/webhooksource/webhook.go
+++ b/pkg/sources/adapter/webhooksource/webhook.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"strings"
 	"time"
 
 	cloudevents "github.com/cloudevents/sdk-go/v2"
@@ -125,7 +126,8 @@ func (h *webhookHandler) handleAll(w http.ResponseWriter, r *http.Request) {
 	event := cloudevents.NewEvent(cloudevents.VersionV1)
 	event.SetType(h.eventType)
 	event.SetSource(h.eventSource)
-	event.SetExtension("http.method", r.Method)
+	event.SetExtension("method", strings.ToLower(r.Method))
+	event.SetExtension("route", strings.ToLower(r.URL.String()))
 
 	if err := event.SetData(r.Header.Get("Content-Type"), body); err != nil {
 		h.handleError(fmt.Errorf("failed to set event data: %w", err), http.StatusInternalServerError, w)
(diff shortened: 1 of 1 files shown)


odacremolbap:
@JeffNeff can you elaborate on the case?

If I am not missing something the URL should be the one exposed at the ksvc status (always the same one).
The method should be a POST, although nothing prevents users from posting any other verbs but that is not very expected.

Are there external consumers that might use the webhook source in a different way?

JeffNeff:
The url is exposed. but you can hit it with different routes and they will appear in the CE extension. This is being implemented so that the Webhook source may act as an API gateway

JeffNeff:
See. If we `GET` `/test` we have this 
```
  specversion: 1.0
  type: test.event
  source: test.source
  id: 
Extensions,
  method: get
  route: /test
Data (binary),
```

JeffNeff:
With this update we can then place triggers/filters to point the request to other flows/transformations/targets/etc

cab105:
While I can see the usefulness of this when running the source standalone, as a proxy for an API gateway, this is not. If you look at what an API Gateway does, it routes incoming traffic to a target service ensuring that:
1. The target service is alive and functional
2. May inject or strip out headers relevant to the target service (or any other device listening on the wire such as a SIEM)
3. May reshape the URL such as awesome.co/coolapp -> awesome.co/hiddensvc/v2/coolapp
4. May verify the presence of tokens such as an oAuth2 access token (or jwt token), validate it, and determine access to the underlying resource.

The way traffic is routed to the Webhook Source is, as @odacremolbap mentioned, a `POST` method, and if a URL path is provided, it is stripped out.

Perhaps one idea for a source, would be to create a plug-in for some API Gateways (or those running NGINX with Lua) to have it produce an event upon each request/action and then allow for an integration to either store the results in Splunk or LogZ (think poor-man's SIEM) or Twilio in the case of services failing.

Thoughts?

JeffNeff:
i still think this is useful... it takes nothing away and adds more routing functionality.. no its not a full fledged API gateway. But I still think it will prove useful in the future.

JeffNeff:
@antoineco Thoughts?

JeffNeff:
@tzununbekov  thoughts?

tzununbekov:
Sorry for being late 😓 
I'd either leave everything as is until we face a use case that will show us the need for these attributes or, go for an all-in and put all request headers in the extensions 🤠 Probably not in the suggested form though.
The relevant paragraph in CE specs: https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#defining-extensions
Btw, I may be a bit out of the context of the webhook source usage but what if my arbitrary little [IoT] producer sends `PUT` instead of `POST`?

PR #301, opened 2021-11-24
AWS SNS Source:check nil address when finalizing

odacremolbap:
When finalizing an SNS Source check if the address has not been set.

### DIFF
--- a/pkg/sources/reconciler/awssnssource/subscription.go
+++ b/pkg/sources/reconciler/awssnssource/subscription.go
@@ -52,10 +52,9 @@ func (r *Reconciler) ensureSubscribed(ctx context.Context) error {
 	status := &src.(*v1alpha1.AWSSNSSource).Status
 
 	isDeployed := status.GetCondition(v1alpha1.ConditionDeployed).IsTrue()
-	url := status.Address.URL
 
 	// skip this cycle if the URL couldn't yet be determined
-	if !isDeployed || url == nil {
+	if !isDeployed {
 		status.MarkNotSubscribed(v1alpha1.AWSSNSReasonNoURL, "The receive adapter isn't ready yet")
 		return nil
 	}
@@ -69,6 +68,7 @@ func (r *Reconciler) ensureSubscribed(ctx context.Context) error {
 			"Error creating SNS client: %s", err))
 	}
 
+	url := status.Address.URL
 	topicARN := typedSrc.Spec.ARN.String()
 
 	subsARN, err := findSubscription(ctx, snsClient, topicARN, url.String())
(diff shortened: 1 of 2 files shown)


tzununbekov (approved):

pkg/sources/reconciler/awssnssource/subscription.go
```diff
@@ -129,8 +129,8 @@ func (r *Reconciler) ensureUnsubscribed(ctx context.Context) error {
 
 	src := v1alpha1.SourceFromContext(ctx)
 
-	url := src.GetStatusManager().Address.URL
-	if url == nil {
+	sm := src.GetStatusManager()
+	if sm == nil || sm.Address == nil || sm.Address.URL == nil {
```
antoineco, line 133: `sm` will never be nil, in case you want to shorten that conditional a little bit:

https://github.com/triggermesh/triggermesh/blob/bd69609b0c5d45a2a5c71f72239bc73d6e6904c5/pkg/apis/sources/v1alpha1/awssns_lifecycle.go#L53

antoineco (approved):
It's surprising to me that the adapter can be "deployed" (ready) and not have an address 🤔 
Have you checked whether that was actually the case in the live cluster?

pkg/sources/reconciler/awssnssource/subscription.go
```diff
@@ -52,14 +52,14 @@ func (r *Reconciler) ensureSubscribed(ctx context.Context) error {
 	status := &src.(*v1alpha1.AWSSNSSource).Status
 
 	isDeployed := status.GetCondition(v1alpha1.ConditionDeployed).IsTrue()
-	url := status.Address.URL
 
 	// skip this cycle if the URL couldn't yet be determined
-	if !isDeployed || url == nil {
+	if !isDeployed || status.Address == nil || status.Address.URL == nil {
```
tzununbekov, line 57: + for Anotine's comment - Isn't `IsDeployed` mean that Address is set?

tzununbekov:
Wouldn't it be more logical to update the failing `ensureUnsubscribed()` function with the `isSubscribed` condition check in the beginning? I believe it would solve the case

odacremolbap:
> It's surprising to me that the adapter can be "deployed" (ready) and not have an address thinking Have you checked whether that was actually the case in the live cluster?

It might be that the instances with the previous incompatible format are behind this.
Anyway, the `FinalizeKind` tries to `ensureUnsubscribed`no matter the readiness, and I think that is the correct behavior. A deployment can be down due to some other reason and we still want the subscription to be removed.

The problem might be that if the object is not ready but was ready before and a subscription was created, the status manager will not inform the address, and the subscription won't be found, hence not removed. That seems a different case to tackle out of this PR, should I open an issue on that?

pkg/sources/reconciler/awssnssource/subscription.go
odacremolbap: yes, removing that one.

odacremolbap:
> Wouldn't it be more logical to update the failing `ensureUnsubscribed()` function with the `isSubscribed` condition check in the beginning? I believe it would solve the case

I think you are very right but here

- Unsubscription operation needs the ARN and Address.
- Deployment true ensures the Address is informed.
- Subscribed true ensures the ARN is informed. 

I think I should be better checking both, which is pretty much the same as asking if the source is ready:

```golang
if sm.IsReady() {
	event.Warn(ctx, ReasonFailedUnsubscribe, "Subscription status incomplete, skipping finalization")
	return nil
}
```

This means that we could be leaking subscriptions if the user runs out of quotas and an SNS pod is evicted, but I think that's something to talk about in a different issue. In general SNS source has a good trade off of cleaning up resources and not dead locking finalizations.

If you don't have concerns on this, I'll go for that code snippet above.

tzununbekov (approved):

PR #313, opened 2021-12-02
Fix Tekton Target's Cryptic Logs

### LINKED ISSUE #242: TektonTarget reaper logs cryptic errors
Seen in the logs of the controller today (formatted for readability):

```json
{
  "level": "info",
  "ts": "2021-11-04T01:31:07.943Z",
  "logger": "triggermesh-controller",
  "caller": "tektontarget/reaper.go:90",
  "msg": "event sending results: {error 26 0  200: }",
  "commit": "b2dcbf2"
}
```

Notice how the following line prints the value returned by [`zap.Error()`](https://pkg.go.dev/go.uber.org/zap#Error) (a struct) instead of either
- passing the error directly as argument to print (`Info("error foo: ", err)`)
- using `Infow()` so that `zap.Error()` actually generates a new "err" key in the structured log entry

_Note: There are many occurrences of this in the same file._

https://github.com/triggermesh/triggermesh/blob/c13be0932d4608455edf31cfb6da07ce96ba492f/pkg/targets/reconciler/tektontarget/reaper.go#L90

See this Go playground for a demo: https://play.golang.org/p/mBWdP3cg1Y__n

JeffNeff:
By using `Infow` and `Fatalw` to correctly print the logs. 

closes #242

### DIFF
--- a/pkg/targets/reconciler/tektontarget/reaper.go
+++ b/pkg/targets/reconciler/tektontarget/reaper.go
@@ -60,7 +60,7 @@ func reaperThread(ctx context.Context, r *reconciler) {
 		for _, ns := range nsl.Items {
 			targets, err := r.targetLister(ns.Name).List(labels.Everything())
 			if err != nil {
-				log.Warnw("unable to retrieve target adapters", zap.Error(err), zap.String("namespace",ns.Name))
+				log.Warnw("unable to retrieve target adapters", zap.Error(err), zap.String("namespace", ns.Name))
 				continue
 			}
 
--- a/pkg/targets/reconciler/tektontarget/reaper.go
+++ b/pkg/targets/reconciler/tektontarget/reaper.go
@@ -40,12 +40,12 @@ func reaperThread(ctx context.Context, r *reconciler) {
 
 	p, err := cloudevents.NewHTTP()
 	if err != nil {
-		log.Fatalw("Failed to create new HTTP based Cloudevent: ", zap.Error(err))
+		log.Fatalw("Failed to create new HTTP based Cloudevent ", zap.Error(err))
 	}
 
 	client, err := cloudevents.NewClient(p)
 	if err != nil {
-		log.Fatalw("Unable to create Cloudevent client: ", zap.Error(err))
+		log.Fatalw("Unable to create Cloudevent client ", zap.Error(err))
 	}
 
 	for {
@@ -53,7 +53,7 @@ func reaperThread(ctx context.Context, r *reconciler) {
 		log.Debug("executing reaping")
 		nsl, err := k8sclient.Get(ctx).CoreV1().Namespaces().List(ctx, v1.ListOptions{})
 		if err != nil {
-			log.Fatalw("Unable to retrieve namespaces: ", zap.Error(err))
+			log.Fatalw("Unable to retrieve namespaces ", zap.Error(err))
 		}
 
 		// search for tektontargets across all namespaces
@@ -71,7 +71,7 @@ func reaperThread(ctx context.Context, r *reconciler) {
 					continue
 				}
 
-				log.Info("Found target: " + t.Namespace + "." + t.Name)
+				log.Info("Found target " + t.Namespace + "." + t.Name)
 				// Send the reap cloudevent
 				cloudCtx := cloudevents.ContextWithTarget(ctx, t.Status.Address.URL.String())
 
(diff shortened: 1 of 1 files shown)


pkg/targets/reconciler/tektontarget/reaper.go
```diff
@@ -71,7 +71,7 @@ func reaperThread(ctx context.Context, r *reconciler) {
 					continue
 				}
 
-				log.Info("Found target: " + t.Namespace + "." + t.Name)
+				log.Infow("Found target: " + t.Namespace + "." + t.Name)
```
cab105, line 74: Do we really need `Infow` here as there isn't a `zap.Error(err)` being passed in as an arg?

pkg/targets/reconciler/tektontarget/reaper.go
```diff
@@ -40,27 +40,27 @@ func reaperThread(ctx context.Context, r *reconciler) {
 
 	p, err := cloudevents.NewHTTP()
 	if err != nil {
-		log.Fatal("failed to create new HTTP based cloudevent: ", zap.Error(err))
+		log.Fatalw("failed to create new HTTP based cloudevent: ", zap.Error(err))
```
odacremolbap, line 43: When using structured logs the error will have its own _field_, we should remove the `: ` at the message.


minor nits: 
- When logging we should capitalize sentences (when passing errors around we do not capitalize).
- `CloudEvent` instead of `cloudevent`
- I am not convinced of the usage of `Fatal`, this code lives in the reconciler and will tear down the whole process. That said, that error is very unexpected to happen and it was there before your PR.

This also applies to the other log fixes in this PR.

```suggestion
		log.Fatalw("Failed to create new HTTP based CloudEvent", zap.Error(err))
```

odacremolbap: Unresolving just in case you didn't noticed the second suggestion.
All log writtings looks like this:

```go
log.Warnw("some message: ", zap.Error(err))
```

The colon should not be there, this is not `log.Warnf("failed to do something: %v", err)` but `log.Warnf` which will render an structured log, using the first string as the `msg` element, and the other parameters as structured details for the log, so `zap.Error(err)` will render as the `error` element. When using JSON it would render something like this:


```json
{"level":"warning","ts":"2021-12-14T13:35:42.152Z","msg":"failed to do something","error":"detailed reason why something failed",...}
```

As you can see, there is no need to add a colon to the `msg`

pkg/targets/reconciler/tektontarget/reaper.go
```diff
@@ -40,27 +40,27 @@ func reaperThread(ctx context.Context, r *reconciler) {
...
-			log.Fatal("Unable to retrieve namespaces: ", zap.Error(err))
+			log.Fatalw("Unable to retrieve namespaces: ", zap.Error(err))
 		}
 
 		// search for tektontargets across all namespaces
 		for _, ns := range nsl.Items {
 			targets, err := r.targetLister(ns.Name).List(labels.Everything())
 			if err != nil {
-				log.Warn("unable to retrieve targets for #{ns.Name}: ", zap.Error(err))
+				log.Warnw("unable to retrieve targets for #{ns.Name}: ", zap.Error(err))
```
odacremolbap, line 63: Not your code, but I am puzzled about this. What does "... #{ns.Name}" do?
Either there is something at zap that I am not aware of or this can be replaced with


```suggestion
				log.Warnw("unable to retrieve target adapters", zap.Error(err), zap.String("namespace",ns.Name))
```

cab105: The syntax inside the log string looks like a ruby in-line variable substitution that has long since been deprecated/removed. Your suggestion makes sense.

odacremolbap (approved):
I'm +1 this, although most logs have a trailing space that would be good to be removed, but I think this is getting too many comments.

Up to you if you want to remove those spaces before the PR gets merged.

pkg/targets/reconciler/tektontarget/reaper.go
```diff
@@ -40,27 +40,27 @@ func reaperThread(ctx context.Context, r *reconciler) {
 
 	p, err := cloudevents.NewHTTP()
 	if err != nil {
-		log.Fatal("failed to create new HTTP based cloudevent: ", zap.Error(err))
+		log.Fatalw("Failed to create new HTTP based Cloudevent ", zap.Error(err))
```
antoineco, line 43: I would still remove the trailing spaces (not just this one, all of them).
Let's ship code in a clean state.

antoineco (changes requested):

antoineco:
Taking over so we can merge this.

antoineco (approved):

PR #319, opened 2021-12-07
Add missing sdk initializer

cab105:
Follow-up to #318  add missing AWS SDK client initialization codebase.

### DIFF
--- a/pkg/targets/adapter/awslambdatarget/adapter.go
+++ b/pkg/targets/adapter/awslambdatarget/adapter.go
@@ -36,18 +36,20 @@ import (
 // NewTarget Adapter implementation
 func NewTarget(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClient cloudevents.Client) pkgadapter.Adapter {
 	env := envAcc.(*envAccessor)
-	config := env.GetAwsConfig()
 	logger := logging.FromContext(ctx)
 
 	a := MustParseARN(env.AwsTargetArn)
 
-	config = config.WithRegion(a.Region)
+	session := session.Must(session.NewSession(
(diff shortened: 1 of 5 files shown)


pkg/targets/adapter/awslambdatarget/adapter.go
```diff
@@ -60,8 +62,7 @@ type adapter struct {
 	awsArnString string
 	awsArn       arn.ARN
 	config       *aws.Config
```
JeffNeff, line 64:
```suggestion

```

pkg/targets/adapter/awssnstarget/adapter.go
```diff
@@ -59,7 +63,7 @@ type adapter struct {
 	awsArn       arn.ARN
 	config       *aws.Config
```
JeffNeff, line 64:
```suggestion

```

pkg/targets/adapter/awssnstarget/adapter.go
```diff
@@ -59,7 +63,7 @@ type adapter struct {
...
 	session      *session.Session
```
JeffNeff, line 65:
```suggestion

```

odacremolbap (approved):

odacremolbap (approved):
+1 when linter is happy

pkg/targets/adapter/awslambdatarget/adapter.go
cab105: Good catch!

pkg/targets/adapter/awssnstarget/adapter.go
cab105: Another good catch!

PR #330, opened 2021-12-13
Update Select AWS Source Adapters to Report Health-Check

JeffNeff:
This PR intends to add health-check capabilities to a few of the AWS Source adapters in order for them to run properly within a Google Cloud Run instance. 

The  includes:
- AWS Code Commit
- AWS Performance Insights
- AWS CloudWatch Logs
- AWS Kinesis 

These were the only ones that I could get working/test.

### DIFF
--- a/pkg/sources/adapter/awscloudwatchlogssource/adapter.go
+++ b/pkg/sources/adapter/awscloudwatchlogssource/adapter.go
@@ -36,6 +36,7 @@ import (
 
 	"github.com/triggermesh/triggermesh/pkg/apis/sources/v1alpha1"
 	"github.com/triggermesh/triggermesh/pkg/sources/adapter/common"
+	"github.com/triggermesh/triggermesh/pkg/sources/adapter/common/health"
 )
 
 // envConfig is a set parameters sourced from the environment for the source's
@@ -123,7 +124,7 @@ func ExtractLogDetails(details string) (string, string) {
 // Start implements adapter.Adapter.
 func (a *adapter) Start(ctx context.Context) error {
 	a.logger.Info("Enabling CloudWatchLog")
-
+	go health.Start(ctx)
 	// Setup polling to retrieve metrics
 	poll := time.NewTicker(a.pollingInterval)
 	defer poll.Stop()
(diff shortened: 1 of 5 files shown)


sebgoa:
How about doing it to all AWS sources that it applies to in this single PR ?

JeffNeff:
> How about doing it to all AWS sources that it applies to in this single PR ?

you got it!

antoineco:
I have a concern with this approach:

These components don't require a web server, they are pull-based. Yet, we are now going to embed a web server in each of them just to expose an endpoint that says "OK".

Even worse: that endpoint **doesn't actually reflect the health of the service**. It just says "OK" to every request, even if the source adapter isn't in an healthy state.

Is there _really_ no better way to run a long running service in Cloud Run? Making adjustments to accommodate a specific environment is one thing, but hacking our own software that way isn't a good look.

At the very least, the health endpoint should effectively reflect the health of the component.

JeffNeff:
@antoineco I agree.

sebgoa:
Why is that same line in the SQS source which is a pull source ?

https://github.com/triggermesh/triggermesh/blob/main/pkg/sources/adapter/awssqssource/adapter.go#L171

antoineco:
@sebgoa good point and yes you're definitely calling me out on the web server argument, but the rest of my comment remains true.

Here is the explanation: https://github.com/triggermesh/aws-event-sources/pull/245

The difference with this PR is that, in the SQS source, we actually check that we can authenticate against SQS (see `queueLookup()` below your quoted line) and _only then_ call `MarkReady()`. This is used for **readiness** purposes, and is not a hack to fake liveness like in this PR.

In the SQS source, if we can't authenticate:

1. The Pod is never marked Ready
2. The Pod fails quickly as it should

I see none of that here.

sebgoa:
Ok fair, it sounds like it would be a good addition to all the sources then. We can verify that we can authenticate with AWS and also maybe check that the ARN "responds".

antoineco:
Yes that would be useful to prevent sources from switching between "Ready" and "Not Ready" when the adapter keeps failing (e.g. due to missing permissions or bad credentials).

Ideally we should send a call to the cloud provider asking if the current credentials are allowed to perform a variety of actions (different for every source), then call `MarkReady()` or fail based on the outcome.

Potentially relevant: [AWS - ](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_testing-policies.html#policies-simulator-using-api)

antoineco:
Considering this is

- not currently being worked
- unfinished (probes don't return 200 right now)

I'm going to go ahead and close.

Feel free to reopen once this is

- reviewable
- properly tested

PR #332, opened 2021-12-14
Salesforce Source: JWT Audience must be sent as string

odacremolbap:
JWT Audience must be sent as string.

Audience field [interpretation is open to the implementer](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3), and in Salesforce case they require it to be a single string, not an array.

---

Also added a backoff mechanism to avoid flooding the Salesforce API on credential issues.

### DIFF
--- a/pkg/sources/adapter/salesforcesource/auth/jwt.go
+++ b/pkg/sources/adapter/salesforcesource/auth/jwt.go
@@ -58,6 +58,11 @@ type claims struct {
 
 // NewJWTAuthenticator creates an OAuth JWT authenticator for Salesforce.
 func NewJWTAuthenticator(certKey, clientID, user, server string, client *http.Client, logger *zap.SugaredLogger) (Authenticator, error) {
+	// JWT package marshals Audience as array even if there is only one element in it. This does not
+	// seem to be supported by Salesforce. By setting the following option to false we tell the imported
+	// library to marshal single item Audience array as a string.
+	jwt.MarshalSingleStringAsArray = false
+
 	audience := strings.TrimSuffix(server, "/")
 
 	signKey, err := jwt.ParseRSAPrivateKeyFromPEM([]byte(certKey))
--- a/pkg/sources/adapter/salesforcesource/auth/jwt.go
+++ b/pkg/sources/adapter/salesforcesource/auth/jwt.go
@@ -58,11 +58,6 @@ type claims struct {
 
 // NewJWTAuthenticator creates an OAuth JWT authenticator for Salesforce.
 func NewJWTAuthenticator(certKey, clientID, user, server string, client *http.Client, logger *zap.SugaredLogger) (Authenticator, error) {
-	// JWT package marshals Audience as array even if there is only one element in it. This does not
-	// seem to be supported by Salesforce. By setting the following option to false we tell the imported
-	// library to marshal single item Audience array as a string.
-	jwt.MarshalSingleStringAsArray = false
-
 	audience := strings.TrimSuffix(server, "/")
 
 	signKey, err := jwt.ParseRSAPrivateKeyFromPEM([]byte(certKey))
--- a/pkg/sources/adapter/salesforcesource/client/bayeux.go
+++ b/pkg/sources/adapter/salesforcesource/client/bayeux.go
@@ -166,16 +166,46 @@ func (b *bayeux) Start(ctx context.Context) error {
 	b.ctx = ctx
 	b.mutex.Unlock()
 
-	bom := wait.NewExponentialBackoffManager(time.Second, time.Minute*5, time.Minute*30, 2, 0, &clock.RealClock{})
+	bom := wait.NewExponentialBackoffManager(time.Second, time.Second*60, time.Second*100, 2, 0, &clock.RealClock{})
 
 	// Connect loop will run until context is done
 	go func() {
+		// helper variable for signaling channels
+		empty := struct{}{}
+
+		// poll channel receives a signal when it is ok to
+		// receive data from Salesforce. We won't send signals
+		// to this channel when an error occurs at connection and
+		// we need to temporarily retry using backoffs, or when
+		// the context is done.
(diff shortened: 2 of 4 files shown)


pkg/sources/adapter/salesforcesource/client/bayeux.go
```diff
@@ -162,6 +166,8 @@ func (b *bayeux) Start(ctx context.Context) error {
 	b.ctx = ctx
 	b.mutex.Unlock()
 
+	bom := wait.NewExponentialBackoffManager(time.Second, time.Minute*5, time.Minute*30, 2, 0, &clock.RealClock{})
```
tzununbekov, line 169: If I understand this right, there are can be scenarios when `ctx.Done` is fired but we're blocked for 5 minutes in the `default` branch with this backoff timer, leaving the pod unresponsive for say rolling updates? Or is it considered to be the edge case and we're okay with it?

pkg/sources/adapter/salesforcesource/auth/jwt.go
```diff
@@ -58,6 +58,11 @@ type claims struct {
 
 // NewJWTAuthenticator creates an OAuth JWT authenticator for Salesforce.
 func NewJWTAuthenticator(certKey, clientID, user, server string, client *http.Client, logger *zap.SugaredLogger) (Authenticator, error) {
+	// JWT package marshals Audience as array even if there is only one element in it. This does not
+	// seem to be supported by Salesforce. By setting the following option to false we tell the imported
+	// library to marshal single item Audience array as a string.
+	jwt.MarshalSingleStringAsArray = false
```
antoineco, line 64: This is a **global** setting, so we shouldn't obfuscate it inside a constructor of some sort.

IMO global tweaks should be made obvious inside the `main` or `init`.

pkg/sources/adapter/salesforcesource/client/bayeux.go
odacremolbap: good point.

it is edgy, but also an anti-pattern. I cannot find  an ExponentialBackoffManager that accepts a context, but let me try to find a better solution.

pkg/sources/adapter/salesforcesource/auth/jwt.go
odacremolbap: I have mixed feelings since that global parameter is set for the Salesforce adapter binary only (no other component, nor the controller), and tied to the `github.com/golang-jwt/jwt/v4` library. Moving that to `cmd/salesforcesource-adapter/main` will move the dependency upwards which does not feel great, but will also bring awareness about the parameter.

That reminds me that I need to add this same parameter for the Salesforce target, which I will do in a minute.
Would it work for you adding the param setting to each of the main.go (source and target) but also making main aware of this import?

I don't have a strong opinion on this, this would be problematic if we create a single binary that does JWT and needs `aud` in different styles, not our case, so whatever you feel more intuitive is ok with me.

pkg/sources/adapter/salesforcesource/client/bayeux.go
odacremolbap: https://github.com/triggermesh/triggermesh/pull/332/commits/f4f6c40a387da83bd497674cec6adf12caed83fd

Can you take a look?

pkg/sources/adapter/salesforcesource/auth/jwt.go
odacremolbap: moved it to `main` at both, source and target.
https://github.com/triggermesh/triggermesh/pull/332/commits/b0a7dfd762a15c2a7c610847b33af5c5aa4b87c9

antoineco: If you're very worried about having this obscure line in the `main()`, we can always create a shared helper `salesforce.SetupJWT()` and call it wherever that makes sense.

The only thing which bothered me was the fact that it was obfuscated a level deeper. It's usually considered a good practice to initialize globals _once_ and _early_.

pkg/sources/adapter/salesforcesource/client/bayeux.go
```diff
@@ -201,6 +243,9 @@ func (b *bayeux) Start(ctx context.Context) error {
 					}
 					b.msgCh <- &crs[i]
 				}
+
+				// process next item.
+				poll <- empty
 			}
```
antoineco, line 249: I'm not sure I understand the purpose of this line. If we always re-poll when the loop complete, why can't we just have the whole thing inside a `case default` of the `select`?

pkg/sources/adapter/salesforcesource/auth/jwt.go
odacremolbap: agree, I think it is less prone to obscure errors now that it is at main

pkg/sources/adapter/salesforcesource/client/bayeux.go
odacremolbap: Falling back to default was the pattern before [Timur's comment.](https://github.com/triggermesh/triggermesh/pull/332#discussion_r768516258)

The default case was incompatible with the backoff. The 3 cases inside the loop are:
- context done, exit function
- error occurred when connecting, wait backoff and do not try to connect
- poll, connect

The error case will wait for the backoff in a go routine and return to the loop so that signaling the context also exit the loop immediately. During backoff we shouldn't try to connect, hence the `select` will wait until the go routine finishes and then it will put a signal at the poll channel.

The poll channel will put a message to re-queue itself if no error occurs during connection.

pkg/sources/adapter/salesforcesource/client/bayeux.go
tzununbekov: That's rather elegant channels management 🙂 
If you tested it and there are no blocking conditions occurred, then lgtm!

tzununbekov (approved):

antoineco (approved):

PR #338, opened 2021-12-20
Remove Azure EventHubResourceID type

antoineco:
The `EventHubResourceID` struct type — which is a structured representation of an Azure Resource ID for _Azure Event Hubs_ resources —  was replaced with an equivalent but more generic `AzureResourceID` type in https://github.com/triggermesh/event-sources/pull/276.

This PR replaces the last occurrence of `EventHubResourceID` with `AzureResourceID`, and removes the former.


odacremolbap (approved):
image
:shipit:

PR #340, opened 2021-12-20
Sanitize user input in log entries (CWE-117)

antoineco:
CodeQL brought to our attention that user input was being logged without sanitization in a few places, potentially allowing attackers to forge log entries, depending on the encoder ([CWE-117](https://cwe.mitre.org/data/definitions/117.html)).

The following Go playground very simply demonstrates the nature of the attack: https://play.golang.com/p/Fk_GxIM071r

This doesn't seem to be a concern when using `zap`'s structured logging (e.g. with `zap.String()`, `zap.Error()`, etc.), because as far as I can tell, both implementations of `zap`'s [Encoder](https://pkg.go.dev/go.uber.org/zap@v1.19.1/zapcore#Encoder) interface escape newline characters properly.

### DIFF
--- a/pkg/sources/adapter/awssnssource/handler/handler.go
+++ b/pkg/sources/adapter/awssnssource/handler/handler.go
@@ -100,7 +100,7 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
-		h.logger.Debug("Request body: ", string(body))
+		h.logger.Debugw("Received Notification request", zap.String("body", string(body)))
 
 		notif := &notification{}
 		if err := json.Unmarshal(body, notif); err != nil {
@@ -135,7 +135,7 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
-		h.logger.Debug("Request body: ", string(body))
+		h.logger.Debugw("Received SubscriptionConfirmation request", zap.String("body", string(body)))
 
 		subsConfirm := &subscriptionConfirmation{}
 		if err := json.Unmarshal(body, subsConfirm); err != nil {
--- a/pkg/sources/adapter/slacksource/slack.go
+++ b/pkg/sources/adapter/slacksource/slack.go
@@ -24,11 +24,13 @@ import (
 	"io/ioutil"
 	"net/http"
 	"strconv"
+	"strings"
 	"time"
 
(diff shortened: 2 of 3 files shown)


pkg/sources/adapter/awssnssource/handler/handler.go
```diff
@@ -100,7 +100,7 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
-		h.logger.Debug("Request body: ", string(body))
+		h.logger.Debugw("Received Notification request", zap.String("body", string(body)))
```
antoineco, line 103: I can only speak about `zap`'s **core** encoder implementations, `json` and `console`, but in both cases I observed that wrapping the request body that way effectively printed newline characters as literal `\n` and `\r`.

pkg/sources/adapter/slacksource/slack.go
```diff
@@ -160,7 +162,7 @@ func (h *slackEventAPIHandler) gracefulShutdown(stopCh <-chan struct{}, done cha
 }
 
 func (h *slackEventAPIHandler) handleError(err error, code int, w http.ResponseWriter) {
-	h.logger.Error("An error ocurred", zap.Error(err))
+	h.logger.Errorw("An error ocurred", zap.Error(err))
```
antoineco, line 165: Similarly, using `Errorw()` together with a structured error field seems to ensure the proper escape of newline characters in stringified errors.

pkg/sources/adapter/slacksource/slack.go
```diff
@@ -122,7 +124,7 @@ func (h *slackEventAPIHandler) handleAll(w http.ResponseWriter, r *http.Request)
 	// There are only 2 documented types to be received from the Events API
 	// - `event_callback`, See: https://api.slack.com/events-api#receiving_events
 	// - `event_callback`, See: https://api.slack.com/events-api#subscriptions
-	switch event.Type {
+	switch eventType := sanitizeUserInput(event.Type); eventType {
```
antoineco, line 127: This event type is eventually logged **inline**, like `"Content not supported: \"Foo\""`.
Considering the nature of the input (a single word), I don't think it is worth logging it as a structured field, and therefore decided to sanitize the input itself, instead of just the logged value.

antoineco (commented):
Notes to reviewers

odacremolbap (approved):

cab105 (approved):

PR #344, opened 2021-12-20
Add Support for CORS Headers to the Webhook Source

### LINKED ISSUE #41: Add support for CORS headers
I received a request for the ability have the HTTP Source send CORS headers during preflight requests where the invoker may require the header for verification.

JeffNeff:
This PR attaches a CORS response header to the response of the incoming POST request. 

If the user specifies the `corsOrigin` option in the spec, upon every request we will now attach the following header parameter: 

```
"Access-Control-Allow-Origin" : <spec.corsOrigin>
```

Closes #41

### DIFF
--- a/config/300-webhooksource.yaml
+++ b/config/300-webhooksource.yaml
@@ -57,6 +57,9 @@ spec:
                   context in which an event happened. Must be expressed as a URI-reference. Please refer to the
                   CloudEvents specification for more details: https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#source-1"
                 type: string
+              corsOrigin:
+                description: "Value of the CORS 'Access-Control-Allow-Origin' header to set on ingested requests."
+                type: string
               basicAuthUsername:
                 description: User name HTTP clients must set to authenticate with the webhook using HTTP Basic
                   authentication.
--- a/config/300-webhooksource.yaml
+++ b/config/300-webhooksource.yaml
@@ -57,7 +57,7 @@ spec:
                   context in which an event happened. Must be expressed as a URI-reference. Please refer to the
                   CloudEvents specification for more details: https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#source-1"
                 type: string
-              corsOrigin:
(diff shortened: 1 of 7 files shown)


config/300-webhooksource.yaml
```diff
@@ -57,6 +57,9 @@ spec:
                   context in which an event happened. Must be expressed as a URI-reference. Please refer to the
                   CloudEvents specification for more details: https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#source-1"
                 type: string
+              corsOrigin:
```
cab105, line 60: I'm sorry, but the name should reflect the full meaning of the header. How about `corsAllowOrigin`?

cab105 (changes requested):
Thank you for the PR. I don't see anything from a logic perspective, but would prefer a name to reflect the full header being set as AllowOrigin. This is to reduce conflict in the other CORS header that is sent by the browser called `Origin`.

pkg/sources/adapter/webhooksource/webhook.go
```diff
@@ -98,6 +97,10 @@ func runHandler(ctx context.Context, s *http.Server) error {
 // handleAll receives all webhook events at a single resource, it
 // is up to this function to parse event wrapper and dispatch.
 func (h *webhookHandler) handleAll(w http.ResponseWriter, r *http.Request) {
+	if h.corsAllowOrigin != "" {
```
JeffNeff, line 100: @cab105 is this right? Or should this be checking aginst a `nil`?

cab105: Much better, thank you!

config/300-webhooksource.yaml
cab105: Thank you!

pkg/sources/adapter/webhooksource/webhook.go
JeffNeff: okay! thank you

cab105 (approved):

Issue #339, opened 2021-12-20
Move e2e test suite and run it in KinD cluster


antoineco:
Create a GitHub Actions workflow that runs the test suite currently located at https://github.com/triggermesh/test-infra/tree/main/test/e2e in a KinD cluster.

The following GitHub Action is able to provision such cluster with Knative pre-installed: https://github.com/marketplace/actions/kubernetes-kind-cluster

antoineco:
Considering this closed by #357.

Tests that remain can't be migrated for now:

- Azure sources (no company account)
- Slack target (requires new token, and potentially racy if I remember correctly)

PR #354, opened 2021-12-24
Deploy TriggerMesh in KinD cluster in preparation for automating E2E tests

antoineco:
Part of #339

In preparation for running e2e tests as part of the standard CI checks, add a GitHub Actions workflow that provisions a local KinD cluster and deploys TriggerMesh in it using `ko`.

I decided to run this workflow on a GitHub Actions runner instead of CircleCI for the time being because [this action](https://github.com/marketplace/actions/kubernetes-kind-cluster) from the GitHub marketplace allows us to provision a KinD cluster **with Knative pre-installed**, which simplifies the setup a lot.

In case we decide to use runners with more resources in the future, we can always switch to CircleCI by creating our own [CircleCI Orb](https://circleci.com/docs/2.0/creating-orbs/) for KinD + Knative. I'm not too worried about resources for the time being though, because even on modest GitHub Actions runners it takes "only" 6 min to `ko apply` TriggerMesh with a warm Go build cache (versus 14 min with a cold cache):

image

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -49,9 +49,6 @@ jobs:
         # use for a container registry.
         registry: 'false'
 
-    - name: Apply Knative Eventing ClusterRoles for addressable-resolver
-      run: kubectl apply -f https://raw.githubusercontent.com/knative/eventing/knative-v1.1.0/config/core/roles/addressable-resolvers-clusterrole.yaml
-
     - name: Install ko
       run: go install github.com/google/ko@v0.9.3
 
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -1,0 +1,69 @@
+name: End-to-End Testing
+
+on:
+  push:
+    # NOTE(antoineco): To avoid exposing repository secrets to user supplied
+    # code in pull requests, we only allow this workflow to run on the 'main'
+    # branch, where all code changes are expected to have been reviewed by
+    # maintainers.
+    #
+    # We may want to relax this in the future, on condition that safeguards are
+    # added to prevent the workflow from running automatically when an external
+    # contributor submits code changes inside the .github/workflows/ or
+    # test/e2e/ directories.
+    #
+    # Ref. https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
+    branches: [ main ]
+
+jobs:
+
+  e2e-triggermesh:
+    name: Test TriggerMesh components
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-tests.yaml
```diff
@@ -0,0 +1,57 @@
...
+          ~/go/pkg/mod
+        key: ${{ github.job }}-${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ github.job }}-${{ runner.os }}-go-
+
+    - name: KinD Cluster
+      uses: container-tools/kind-action@v1
+      with:
+        knative_serving: v1.0.0
+        knative_kourier: v1.0.0
```
odacremolbap, line 37: should `knative_eventing` be also used?

antoineco: AFAIK none of the end-to-tests directly related to this repo rely on Eventing (direct connections only), so I wanted to save some resources.

But now that you mention it, it could be that the "addressable-resolver" role from Eventing is required to resolve `sink.ref`. Let me try to verify whether that is the case by running a real test.

antoineco: As I expected, TriggerMesh only requires the aforementioned ClusterRole to work, and none of the actual Eventing components. So instead of enabling the entire Eventing bundle (with the default Broker + Channel), I figured it was enough to do [this](https://github.com/triggermesh/triggermesh/compare/f5b3e8e25cc0a73dd3b465cdb414553aba270655..c66683529a0c2e21a913acaa7c80bc2a47654a0c).

Let me know whether that works for you.

Merry Christmas! 🎄

odacremolbap: sure thing!

odacremolbap (approved):

.github/workflows/e2e-tests.yaml
```diff
@@ -0,0 +1,66 @@
...
+    # branch, where all code changes are expected to have been reviewed by
+    # maintainers.
+    #
+    # We may want to relax this in the future, on condition that safeguards are
+    # added to prevent the workflow from running automatically when an external
+    # contributor submits code changes inside the .github/workflows/ or
+    # test/e2e/ directories.
+    #
+    # Ref. https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
+    branches: [ main ]
```
antoineco, line 16: Adding a note for myself about recommended security practices in GitHub Actions: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions

PR #357, opened 2021-12-26
Migrate remaining E2E tests for event sources from test-infra

antoineco:
Part of #339
Requires #356

Migrates the rest of the source tests from test-infra, with the exception of tests related to Microsoft Azure services because we don't yet have a company account.

---

A successful test run: https://github.com/triggermesh/triggermesh/actions/runs/1631370315

PR #379, opened 2021-12-29
Improve E2E test suite for Google Cloud Pub/Sub source

antoineco:
While migrating E2E tests in #357, I noticed a few issues / possible improvements with the test suite for the Google Cloud Pub/Sub source:

- The test case where the Pub/Sub subscription is provided by the user...
  - ... does not clean up the subscription at the end of the test ([code](https://github.com/triggermesh/triggermesh/blob/1d8390e2bbe5fb30ec9dbe62fc2b8f67747a477c/test/e2e/sources/googlecloudpubsub/main.go#L181-L185))
  - ... sets `spec.subscription` while the attribute is called `spec.subscriptionID` ([code](https://github.com/triggermesh/triggermesh/blob/1d8390e2bbe5fb30ec9dbe62fc2b8f67747a477c/test/e2e/sources/googlecloudpubsub/main.go#L284-L285), [spec](https://github.com/triggermesh/triggermesh/blob/1d8390e2bbe5fb30ec9dbe62fc2b8f67747a477c/config/300-googlecloudpubsubsource.yaml#L60-L65))

- The same setup steps and test logic were duplicated ([code](https://github.com/triggermesh/triggermesh/blob/1d8390e2bbe5fb30ec9dbe62fc2b8f67747a477c/test/e2e/sources/googlecloudpubsub/main.go#L91-L113), [code](https://github.com/triggermesh/triggermesh/blob/1d8390e2bbe5fb30ec9dbe62fc2b8f67747a477c/test/e2e/sources/googlecloudpubsub/main.go#L187-L211)).
  - _solution:_
    - I nested the two tests under a common test context, so they can use shared setup steps (`BeforeEach`).
    - I moved the test logic to a closure (`SendMessageAndAssertReceivedEvent`), as demonstrated in the [Ginkgo docs](https://onsi.github.io/ginkgo/#locally-scoped-shared-behaviors).

- Some variable names weren't explicit about what they represent.
  - _solution:_ append `Name` when something represents a _Resource Name_, `ID` when something represents an ID

- Every Google Cloud test was re-declaring common own environment variables ([code](https://github.com/triggermesh/triggermesh/blob/1d8390e2bbe5fb30ec9dbe62fc2b8f67747a477c/test/e2e/sources/googlecloudpubsub/main.go#L62-L63)).
  - _solution:_ I moved the variables to the framework's `gcloud` package, and simplified the contract with `ServiceAccountKeyFromEnv()` and `ProjectnameFromEnv()` helpers that do not require any parameter.

- Misc:
  - Some helpers were accepting a `Topic` struct (client) while only the topic ID was required.
  - Added more test cases for invalid specs (empty creds, invalid subscription ID, empty topic name)

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -367,7 +367,7 @@ The benefits of this approach become more obvious when the complexity of test sc
 [ginkgo-docs]: https://onsi.github.io/ginkgo/
 [ginkgo-issue222]: https://github.com/onsi/ginkgo/issues/222
 [ginkgo-struct]: https://onsi.github.io/ginkgo/#structuring-your-specs
-[optimized-test]: https://github.com/triggermesh/triggermesh/blob/main/test/e2e/sources/awscodecommit/main.go#L171-L187
+[optimized-test]: https://github.com/triggermesh/triggermesh/blob/1d8390e2/test/e2e/sources/awscodecommit/main.go#L171-L187
 
 [k8s-e2e]: https://godoc.org/k8s.io/kubernetes/test/e2e
 [sigtesting-howto]: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-testing/writing-good-e2e-tests.md
--- a/test/e2e/framework/gcloud/gcloud.go
+++ b/test/e2e/framework/gcloud/gcloud.go
@@ -23,6 +23,11 @@ import (
 	"github.com/triggermesh/triggermesh/test/e2e/framework"
 )
 
+const (
+	envServiceAccountKey = "GCLOUD_SERVICEACCOUNT_KEY"
+	envProjectName       = "GCLOUD_PROJECT"
+)
+
 const pubsubLabelOwnerResource = "io-triggermesh_owner-resource"
 
 // PubSubResourceID returns a deterministic Pub/Sub resource ID matching the given framework.Framework.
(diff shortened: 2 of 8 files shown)


test/e2e/README.md
```diff
@@ -367,7 +367,7 @@ The benefits of this approach become more obvious when the complexity of test sc
 [ginkgo-docs]: https://onsi.github.io/ginkgo/
 [ginkgo-issue222]: https://github.com/onsi/ginkgo/issues/222
 [ginkgo-struct]: https://onsi.github.io/ginkgo/#structuring-your-specs
-[optimized-test]: https://github.com/triggermesh/triggermesh/blob/main/test/e2e/sources/awscodecommit/main.go#L171-L187
+[optimized-test]: https://github.com/triggermesh/triggermesh/blob/1d8390e2/test/e2e/sources/awscodecommit/main.go#L171-L187
```
cab105, line 370: Why the change from main to a specific hash?

cab105 (approved):
I'm always impressed when reading your code. I do have one question more from a documentation standpoint, but not a blocker.

test/e2e/README.md
antoineco: So that the link doesn't start pointing at something else than originally intended if the code changes on `main` :) Typical

PR #337, opened 2021-12-20
IBM MQ source and target

tzununbekov:
In this PR:
- IBM MQ Source and Target
- Makefile updates to support custom-built components such as IBM MQ adapters

Not in this PR:
- Tests
- Validation webhooks for new CRDs
- Source's delivery backoff delay is exposed in the CRD but not handled by the adapter

The things that are missing in this PR will be filled in as follow-up issues to speed up the merge process.

PR #384, opened 2021-12-30
Revert "IBM MQ source and target"

antoineco:
Reverts triggermesh/triggermesh#337

This PR broke the release job because the CI runner doesn't have Docker installed.

Reverting to avoid blocking other PRs. Feel free to revert the revert once the CI changes have been tested.

cc. @tzununbekov


antoineco:
@tzununbekov I see your fix is on `main` now.

If it was tested on a fork and we are 100% sure container releases are now working, feel free to click "Revert" above ☝️

tzununbekov:
@antoineco I have to admit that I don't have a copy of the build environment in my personal fork to run the whole release cycle there. If there is a way to set up the test dependencies in the fork without the need to have all external services available, I'd gladly do that. Otherwise, I'd merge the update and see how it goes - I believe we must have at least a couple of days for fixes before the next release date.

antoineco:
What I usually do is keep only the part of the CircleCI config that I need, and push it to a test branch to see if it does what I expect.

For example

- keep only `publish-images` inside the CircleCI config
- make CircleCI run `publish-images` on **every commit** (not only merge to `main`)
- set the target container repo to `gcr.io/triggermesh/antoine` (or Docker Hub, or anything else you can easily throw away)
- push to `test-circleci` branch and open CircleCI to see if it completes

PR #389, opened 2021-12-31
Restructure Google Cloud Audit Logs E2E test suite

### LINKED ISSUE #377: Cloud Audit Logs E2E test asserts for single event, but might receive any number of events
The E2E test suite for the Google Cloud Audit Logs source asserts for a single event:

```go
// test/e2e/sources/googlecloudrepositories/main.go
Expect(receivedEvents).To(HaveLen(1))
```

This is incorrect in this context, because Audit Logs collects events for _the whole Google Cloud project_, not just the resources created by the test. It is very likely that another test / application/ developer ends up generating unrelated Pub/Sub audit logs during the lifecycle of this test.

For this reason, we often observe [failures in E2E tests](https://github.com/triggermesh/triggermesh/actions/runs/1631370315/attempts/1).

antoineco:
- Rename a few variables for consistency with other Google Cloud tests
- Don't enforce that we receive a single event (see #377)
- Add test cases to assert that Kubernetes properly rejects objects with invalid or missing attributes

Fixes #377

### DIFF
--- a/test/e2e/sources/googlecloudauditlogs/main.go
+++ b/test/e2e/sources/googlecloudauditlogs/main.go
@@ -60,17 +60,15 @@ const (
 	sourceResource = "googlecloudauditlogssources"
 )
 
-var _ = Describe("Google Cloud AuditLogs source", func() {
+var _ = Describe("Google Cloud Audit Logs source", func() {
 	f := framework.New("googlecloudauditlogssource")
 
 	var ns string
 
 	var srcClient dynamic.ResourceInterface
 
-	var serviceName string
-	var methodName string
-	var project string
-	var saKey string
+	var gcloudProject string
+	var serviceaccountKey string
 
 	var sink *duckv1.Destination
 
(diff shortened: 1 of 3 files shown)


test/e2e/sources/googlecloudauditlogs/main.go
```diff
@@ -135,7 +140,6 @@ var _ = Describe("Google Cloud AuditLogs source", func() {
 				readReceivedEvents := readReceivedEvents(f.KubeClient, ns, sink.Ref.Name, &receivedEvents)
 
 				Eventually(readReceivedEvents, receiveTimeout, pollInterval).ShouldNot(BeEmpty())
-				Expect(receivedEvents).To(HaveLen(1))
```
cab105, line 138: Wouldn't it be better to expect there to be at least one event received otherwise the test will panic in the next line attempting to obtain the first element of a zero length array?

cab105 (commented):
Sadly found one not-so minor nit, but should be an easy fix.

test/e2e/sources/googlecloudauditlogs/main.go
antoineco: That case is covered by `ShouldNot(BeEmpty())` above. We won't reach that point if no event is received.

cab105: Oh right. I was expecting a `> 0` check here, but that works too.

cab105 (approved):

PR #396, opened 2022-01-03
IBM MQ source and target

tzununbekov:
This is the clone of the previously reverted IBM MQ [PR](https://github.com/triggermesh/triggermesh/pull/337) but with one tiny addition that enables Docker's [Buildkit](https://github.com/triggermesh/triggermesh/compare/ibm-mq?expand=1#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R136) feature.


antoineco (approved):
Approving with a tiny comment on the usage of env vars in the CircleCI config.

Issue #407, opened 2022-01-07
Use tags to verify the ownership of a subscription prior to deleting it


antoineco:
Currently, the SNS subscription observed by the source is removed by the finalizer unconditionally when the source object is deleted.

We could make this operation safer by ensuring that we actually own the subscription prior to deleting it.

We already do something similar using tags in the source for S3:

https://github.com/triggermesh/triggermesh/blob/bd69609b0c5d45a2a5c71f72239bc73d6e6904c5/pkg/sources/reconciler/awss3source/queue.go#L245-L254

Issue #403, opened 2022-01-05
Pub/Sub resources aren't cleaned up at the end of a successful test


antoineco:
### Context

When a Google Cloud Storage source is created:

1. A Pub/Sub topic is created
1. A Pub/Sub subscription is attached to the topic
1. A notification configuration is added to the user-provided bucket
1. A receive adapter is deployed

When a Google Cloud Storage source is deleted:

1. The receive adapter is deleted
1. The notification configuration is removed from the user-provided bucket
1. The Pub/Sub subscription is detached from the topic
1. The Pub/Sub topic is deleted

### Observed problem

In our test suite, the bucket is deleted **before** the source object. This situation is expected, but likely to throw an error during the finalization of the source object, which might be why Pub/Sub resources aren't cleaned up by the source's finalizer.

### Extra notes

I suspect this is occurring in all Google Cloud Sources.

antoineco:
It seems like a 404 response (Not Found) inside `ensureNoNotificationConfig` throws an error instead of giving up with a warning. I suspect that the `isNotFound` case isn't handled properly:

https://github.com/triggermesh/triggermesh/blob/48a4a025fe6a4a208363f4f185a1215e7d681ddf/pkg/sources/reconciler/googlecloudstoragesource/notif_config.go#L131-L134

`"message":"Returned an event"` indicates that the function didn't return `nil`:

```json
{"severity":"INFO","timestamp":"2022-01-05T17:56:03.8075263Z","logger":"triggermesh-controller","caller":"googlecloudstoragesource/reconciler.go:287","message":"Returned an event","commit":"dcb880e","knative.dev/controller":"github.com.triggermesh.triggermesh.pkg.sources.reconciler.googlecloudstoragesource.Reconciler","knative.dev/kind":"sources.triggermesh.io.GoogleCloudStorageSource","knative.dev/traceid":"b00d2bee-9c96-4e34-853f-2b7b52119a62","knative.dev/key":"e2e-gcloudstoragesource-4467/test-kc7rl","targetMethod":"ReconcileKind","event":"failed to reconcile bucket notification configuration: Error obtaining notification configuration for bucket \"e2e-gcloudstoragesource-4467\": looking up notification configuration: googleapi: Error 404: The specified bucket does not exist., notFound"}
```

Indeed, the reconciliation is failed (`"message":"Reconcile error"`):

```json
[...]

PR #408, opened 2022-01-07
Delete SNS subscription even if the topic was already deleted

### LINKED ISSUE #392: Dangling SNS subscriptions left behind
e2e tests are leaving a bunch of subscriptions behind

image

I suspect this happens because the topic is removed (in the test's `AfterEach`) before the event source (due to namespace termination), so the finalizer doesn't get a chance to match subscriptions to a topic.

image

antoineco:
Fixes #392
Remotely related to #403 (which is already fixed on `main`)

In the finalizer, we now start by checking whether or not the user-provided topic exists.

- If YES, ask AWS for the subscription's ARN _(the existing behavior)_
- If NO, fallback to reading the subscription's ARN from the source's status _(⭐ added in this PR)_

**Note:** we could/should make the deletion safer by setting/verifying tags on the reconciled subscription (like we do in the [source for S3](https://github.com/triggermesh/triggermesh/blob/bd69609b0c5d45a2a5c71f72239bc73d6e6904c5/pkg/sources/reconciler/awss3source/queue.go#L245-L254)), but since this feature didn't exist prior to this PR, I'm going to consider it off-scope and opened #407 to tackle it.

---

The fact that it works as intended should be surfaced by E2E tests.

Meanwhile, here is a manual test which demonstrates that a subscription to a topic that doesn't exist (anymore) can nevertheless be deleted by the reconciler:

```console
$ aws sns list-topics
{
    "Topics": []
}
```

```console
$ aws sns list-subscriptions
{
    "Subscriptions": [
        {
            "SubscriptionArn": "arn:aws:sns:eu-central-1:123456789012:e2e-awssnssource-5385:ce6922a6-f988-4b53-bacd-90678f7ecc49",
            "Owner": "123456789012",
            "Protocol": "https",
            "Endpoint": "https://awssnssource-adapter-antoineco.dev.triggermesh.io/antoineco/sample",
            "TopicArn": "arn:aws:sns:eu-central-1:123456789012:e2e-awssnssource-5385"
        }
    ]
}
```

```console
$ kubectl delete awssnssources.sources.triggermesh.io/sample
awssnssource.sources.triggermesh.io "sample" deleted
```

```console
$ aws sns list-subscriptions
{
    "Subscriptions": []
}
```

### DIFF
--- a/pkg/sources/reconciler/awssnssource/subscription.go
+++ b/pkg/sources/reconciler/awssnssource/subscription.go
@@ -48,8 +48,8 @@ func (r *Reconciler) ensureSubscribed(ctx context.Context) error {
 		return nil
 	}
 
-	src := v1alpha1.SourceFromContext(ctx)
-	status := &src.(*v1alpha1.AWSSNSSource).Status
+	src := v1alpha1.SourceFromContext(ctx).(*v1alpha1.AWSSNSSource)
+	status := &src.Status
 
 	isDeployed := status.GetCondition(v1alpha1.ConditionDeployed).IsTrue()
 
@@ -59,9 +59,7 @@ func (r *Reconciler) ensureSubscribed(ctx context.Context) error {
 		return nil
 	}
 
-	typedSrc := src.(*v1alpha1.AWSSNSSource)
-
-	snsClient, err := r.snsCg.Get(typedSrc)
+	snsClient, err := r.snsCg.Get(src)
 	if err != nil {
 		status.MarkNotSubscribed(v1alpha1.AWSSNSReasonNoClient, "Cannot obtain SNS client")
 		return fmt.Errorf("%w", reconciler.NewEvent(corev1.EventTypeWarning, ReasonFailedSubscribe,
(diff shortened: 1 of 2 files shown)


pkg/sources/reconciler/awssnssource/subscription.go
```diff
@@ -129,17 +129,8 @@ func (r *Reconciler) ensureUnsubscribed(ctx context.Context) error {
...
 
-	sm := src.GetStatusManager()
-
-	// Note: requiring readiness to unsubscribe might leak subscriptions when the service
-	// is not ready. The reason for requiring IsReady is because we require the status address
-	// when looking for the SNS subscription and it is informed when the service is ready.
-	if !sm.IsReady() {
-		event.Warn(ctx, ReasonFailedUnsubscribe, "SNS status information incomplete, skipping finalization")
-		return nil
-	}
```
antoineco, line 140: This code (which I wrote) was wrong.

A Knative Service will in most cases have a URL even if it's not ready, so this check is not only unnecessary, but can also lead to leaking SNS subscriptions, which this PR tries to prevent.

pkg/sources/reconciler/awssnssource/subscription.go
```diff
@@ -156,30 +147,63 @@ func (r *Reconciler) ensureUnsubscribed(ctx context.Context) error {
 
 	topicARN := typedSrc.Spec.ARN.String()
 
-	subsARN, err := findSubscription(ctx, snsClient, topicARN, sm.Address.URL.String())
+	// We need to check the existence of the topic prior to attempting to
+	// list its subscriptions, because AWS returns the same NotFound error
+	// whether the topic or the subscription is missing, and we want to
+	// handle both cases differently.
+	topicExists, err := checkTopicExists(ctx, snsClient, topicARN)
```
antoineco, line 154: In fact, we should also do that inside `ensureSubscribe()`. Currently, we end up in a reconciliation loop if the topic doesn't exist.

Let me know if you're fine with me adding it in this PR, even though it's not directly related to the original issue.

antoineco (commented):
Notes to reviewers

pkg/sources/reconciler/awssnssource/subscription.go
```diff
@@ -228,6 +252,23 @@ func findSubscription(ctx context.Context, cli snsiface.SNSAPI, topicARN, endpoi
 	return "", awserr.New(sns.ErrCodeNotFoundException, "", nil)
 }
 
+// checkTopicExists returns whether the topic with the given ARN exists.
+func checkTopicExists(ctx context.Context, cli snsiface.SNSAPI, topicARN string) (bool, error) {
+	in := &sns.GetTopicAttributesInput{
+		TopicArn: &topicARN,
+	}
+
+	_, err := cli.GetTopicAttributesWithContext(ctx, in)
```
antoineco, line 261: Note to self: implement this method in the mocked SNS client. Tests are currently failing:

```
=== RUN   TestReconcileSubscription/Deletion_while_subscribed
    --- FAIL: TestReconcileSubscription/Deletion_while_subscribed (0.03s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
...
```

pkg/sources/reconciler/awssnssource/subscription.go
cab105: Do what you've got to do.

cab105 (approved):

PR #417, opened 2022-01-11
IBM MQ build fixes

tzununbekov:
In this PR:

- Buildkit cache disabled to prevent building IBM images with the same binary for the source and the target adapters. The reason for that behavior is not completely clear since we're using cache for go modules only, but the result is reproducible. An example of the CI job with the "Cached" build:
  https://app.circleci.com/pipelines/github/triggermesh/triggermesh/994/workflows/b26c1c56-0d3e-465d-ab4a-7e54d54c6990/jobs/2633?invite=true#step-112-4838
  Ideally, we should bring this cache back as it drastically improves the build time, but for now, to get the working adapter images in the registry we can get rid of it

- Added IBM MQ redis client installation to CodeQL workflow to run analysis in the IBM adapters code

- Golangci's Linter job uses `codeanalysis` build tag to be able to skip analysis on the certain packages

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -11,6 +11,8 @@ linters:
 
 run:
   timeout: 10m
+  build-tags:
+    - codeanalysis
 
 linters-settings:
   stylecheck:
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -12,7 +12,7 @@ linters:
 run:
   timeout: 10m
   build-tags:
-    - codeanalysis
+  - noclibs
 
 linters-settings:
   stylecheck:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -137,7 +137,6 @@ jobs:
           environment:
             KO_DOCKER_REPO: gcr.io/triggermesh
             DIST_DIR: /tmp/dist/
-            DOCKER_BUILDKIT: 1
       - persist_to_workspace:
           root: /tmp/
           paths:
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -42,6 +42,7 @@ jobs:
       with:
         languages: go
 
+    # Install IBM MQ redis client required by IBM MQ source and target adapters
     - name: Install IBM MQ client
       run: |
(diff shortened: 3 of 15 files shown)


.golangci.yaml
```diff
@@ -11,6 +11,8 @@ linters:
 
 run:
   timeout: 10m
+  build-tags:
+    - codeanalysis
```
antoineco, line 15: nit: The indentation doesn't match the rest of the file.

About the tag: `codeanalysis` feels obscure to me. How about something that says exactly what this is about, such as `noclibs` ("no C libraries"). It's more versatile and can be used with anything (not just analysis tools).

antoineco (approved):

antoineco (changes requested):
Oh, before I forget, this (and the corresponding `cmd/`) is also breaking the linter:

```console
$ golangci-lint run ./pkg/flow/adapter/xslttransform
WARN [runner] Can't run linter goanalysis_metalinter: buildir: failed to load package help: could not load export data: no export data for "github.com/jbowtie/gokogiri/help"
```

antoineco (approved):
edit: Everything looks OK in the CI, I must be missing some library. Feel free to ignore!

image

tzununbekov:
> Oh, before I forget, this (and the corresponding `cmd/`) is also breaking the linter:
> 
> ```
> $ golangci-lint run ./pkg/flow/adapter/xslttransform
> WARN [runner] Can't run linter goanalysis_metalinter: buildir: failed to load package help: could not load export data: no export data for "github.com/jbowtie/gokogiri/help"
> ```

I cannot reproduce it locally, could you point me to the build job with that issue?

antoineco:
@tzununbekov don't forget to pass the `noclibs` build tag here:

https://github.com/triggermesh/triggermesh/blob/f703e7bb7a852bdf738d2949318fd3c6f294c8a2/.github/workflows/codeql.yaml#L45-L49

tzununbekov:
> @tzununbekov don't forget to pass the `noclibs` build tag here:
> 
> https://github.com/triggermesh/triggermesh/blob/f703e7bb7a852bdf738d2949318fd3c6f294c8a2/.github/workflows/codeql.yaml#L45-L49

Wait, but I added IBM MQ client installation to make CodeQL workflow pass

antoineco:
Right, silly me, I got confused.

Let's see how this goes: https://github.com/triggermesh/triggermesh/actions/runs/1682833900

PR #419, opened 2022-01-11
Remove unnecessary interface check on `runtime.Object`

antoineco:
`runtime.Object` is already part of the `EventSource` interface, so we have a double assertion.

### DIFF
--- a/pkg/apis/sources/v1alpha1/awscloudwatch_types.go
+++ b/pkg/apis/sources/v1alpha1/awscloudwatch_types.go
@@ -18,8 +18,6 @@ package v1alpha1
 
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
-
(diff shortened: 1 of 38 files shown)


pkg/apis/sources/v1alpha1/awscloudwatch_types.go
```diff
@@ -40,8 +39,7 @@ type AWSCloudWatchSource struct {
 
 // Check the interfaces the event source should be implementing.
 var (
-	_ runtime.Object = (*AWSCloudWatchSource)(nil)
-	_ EventSource    = (*AWSCloudWatchSource)(nil)
+	_ EventSource = (*AWSCloudWatchSource)(nil)
```
antoineco, line 42: `runtime.Object` is part of the `EventSource` interface already.

antoineco (commented):
tl;dr for reviewers

The comments below were simply applied to all Source and Router types.

cab105 (approved):

PR #421, opened 2022-01-11
Remove Azure `StorageAccountResourceID` type

### LINKED ISSUE #420: Get rid of `StorageAccountResourceID` custom type
This is similar to #338.

We use a dedicated type for the resource ID of an Azure Storage Account in this code location:

https://github.com/triggermesh/triggermesh/blob/e110e986f4ce357567f116aec5c8c2ab6c19fb66/pkg/apis/sources/v1alpha1/azureblobstorage_types.go#L62

We already have the `AzureResourceID` type that can represent _any_ Azure resource ID, so we should switch to that:

https://github.com/triggermesh/triggermesh/blob/393c94eb73118bd7fc0302eec332136d522e0448/pkg/apis/sources/v1alpha1/azure_common_types.go#L55-L65

antoineco:
Closes #420


cab105 (approved):
Don't see any major issues here.

PR #423, opened 2022-01-11
Do not set core CE attributes via `ceOverrides`

### LINKED ISSUE #373: Do not override CloudEvents core attributes via `ceOverrides`
We abuse a bug of the CloudEvents SDK for Go in several places, by using Knative's `ceOverrides` mechanism to hardcode the value of CloudEvents core attributes such as `source` and `type`.

As far as I can tell, the affected components are:

- The source for _Azure Activity Logs_ (to override the [default CE source](https://github.com/triggermesh/triggermesh/blob/48a4a025fe6a4a208363f4f185a1215e7d681ddf/pkg/sources/adapter/azureeventhubsource/adapter.go#L99), which is the resource ID of the backing Event Hubs instance)
- Most sources for _Google Cloud_ (to override the [default CE source](https://github.com/triggermesh/triggermesh/blob/48a4a025fe6a4a208363f4f185a1215e7d681ddf/pkg/sources/adapter/googlecloudpubsubsource/adapter.go#L85-L90), which is the fully-qualified resource name of the backing Pub/Sub topic)

Example taken from the Google Cloud Source Repositories source:

```yaml
    env:
    - name: K_CE_OVERRIDES
      value: '{"extensions":{"source":"projects/cebuk-01/repos/e2e-googlecloudrepositoriessource-3685","type":"com.google.cloud.repositories.notification"}}'
```

This is no longer working since the v2.7.0 release of the SDK, which means the behaviour got broken by #361.

Example of error:

```json
{
  "severity": "ERROR",
  "timestamp": "2021-12-28T16:23:24.5878306Z",
  "logger": "googlecloudpubsubsource",
  "caller": "googlecloudpubsubsource/adapter.go:137",
  "message": "Failed to send CloudEvents",
  "commit": "406b7f3",
  "error": "[\"failed to send event with ID 3816218250973425: extension:source: bad key \\\"source\\\": CloudEvents spec attribute MUST NOT be overwritten by extension\\nextension:type: bad key \\\"type\\\": CloudEvents spec attribute MUST NOT be overwritten by extension\\n\"]",
  "stacktrace": "github.com/triggermesh/triggermesh/pkg/sources/adapter/googlecloudpubsubsource.(*adapter).handleMessage\n\tgithub.com/triggermesh/triggermesh/pkg/sources/adapter/googlecloudpubsubsource/adapter.go:137\ncloud.google.com/go/pubsub.(*Subscription).Receive.func2.2\n\tcloud.google.com/go/pubsub@v1.17.1/subscription.go:1003\ncloud.google.com/go/pubsub/internal/scheduler.(*ReceiveScheduler).Add.func1\n\tcloud.google.com/go/pubsub@v1.17.1/internal/scheduler/receive_scheduler.go:75"
}
```

Notice the content of the error message:

```
failed to send event with ID 3816218250973425:
  extension:source: bad key "source": CloudEvents spec attribute MUST NOT be overwritten by extension
  extension:type: bad key "type": CloudEvents spec attribute MUST NOT be overwritten by extension
```

antoineco:
Fixes #373

This no more permitted as of cloudevents/sdk-go@v2.7.0:
"CloudEvents spec attribute MUST NOT be overwritten by extension"

Instead, introduce the `CE_SOURCE` and `CE_TYPE` (optional) environment variables for overriding the default source/type context attributes.

Currently supported by the following receive adapters:
- Azure Event Hubs
- Google Cloud Pub/Sub

We have no use for such override in any other source.

### DIFF
--- a/pkg/sources/adapter/azureeventhubsource/adapter.go
+++ b/pkg/sources/adapter/azureeventhubsource/adapter.go
@@ -34,6 +34,8 @@ import (
 	pkgadapter "knative.dev/eventing/pkg/adapter/v2"
 	"knative.dev/pkg/logging"
 
+	"github.com/triggermesh/triggermesh/pkg/apis/sources"
+	"github.com/triggermesh/triggermesh/pkg/apis/sources/v1alpha1"
 	"github.com/triggermesh/triggermesh/pkg/sources/adapter/azureeventhubsource/trace"
 )
 
@@ -54,6 +56,10 @@ type envConfig struct {
 	//
 	// Supported values: [ default eventgrid ]
 	MessageProcessor string `envconfig:"EVENTHUB_MESSAGE_PROCESSOR" default:"default"`
+
+	// Allows overriding common CloudEvents attributes.
+	CEOverrideSource string `envconfig:"CE_SOURCE"`
+	CEOverrideType   string `envconfig:"CE_TYPE"`
 
 	// The environment variables below aren't read from the envConfig struct
 	// by the Event Hubs SDK, but rather directly using os.Getenv().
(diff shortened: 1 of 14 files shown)


cab105 (commented):
I maybe missing something, but wouldn't the AWS Kinesis/SNS/SQS sources be applicable here as well?

antoineco:
@cab105 the sources you mentioned have their own adapter, which has knowledge of what the "source" and "type" mean for these particular AWS services (spoiler alert: the source is always the ARN; the type is sometimes based on the payload, sometimes static). [Example](https://github.com/triggermesh/triggermesh/blob/9fa816d801979128852ec1ce10eabe32527047a7/pkg/sources/adapter/awskinesissource/adapter.go#L199-L201)

The adapters for Event Hubs and Pub/Sub are special because they back respectively _all_ the sources for Azure and Google Cloud, so each of these sources needs a way to say "you may have received a message from Event Hubs or Pub/Sub, but the data it contains originates from XYZ".

cab105:
Fair, I was thinking of the targets when I realized what this would do.

cab105 (approved):

PR #425, opened 2022-01-12
Add AWS SQS Target e2e test

### LINKED ISSUE #414: Add E2E tests for Amazon SQS target

FranBarrera:
Closes #414

### DIFF
--- a/test/e2e/framework/cloudevents/cloudevents.go
+++ b/test/e2e/framework/cloudevents/cloudevents.go
@@ -42,8 +42,8 @@ const (
 	eventSenderNamePrefix = "eventsender-"
 )
 
-// NewEvent generates a CloudEvent with dummy values.
-func NewEvent(f *framework.Framework) *cloudevents.Event {
+// NewHelloEvent generates a CloudEvent with dummy values.
+func NewHelloEvent(f *framework.Framework) *cloudevents.Event {
 	event := cloudevents.NewEvent()
 	event.SetID("0000")
 	event.SetType("e2e.test")
--- a/test/e2e/framework/cloudevents/cloudevents.go
+++ b/test/e2e/framework/cloudevents/cloudevents.go
@@ -1,0 +1,109 @@
+/*
+Copyright (c) 2022 TriggerMesh Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package cloudevents contains helpers to interact with CloudEvents.
+package cloudevents
+
+import (
+	"context"
+	"encoding/json"
+
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	"github.com/cloudevents/sdk-go/v2/protocol/http"
+
+	"github.com/triggermesh/triggermesh/test/e2e/framework"
+)
+
+// E2EInstanceCeExtension is a CloudEvent extension that is used to propagate
+// the name of the test that generated the event.
+const E2ECeExtension = "iotriggermeshe2e"
+
+const (
+	curlContainerName     = "curl"
+	curlContainerImage    = "curlimages/curl:latest"
+	eventSenderNamePrefix = "eventsender-"
+)
+
+// NewEvent generates a CloudEvent with dummy values.
+func NewEvent(f *framework.Framework) *cloudevents.Event {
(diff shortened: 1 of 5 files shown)


test/e2e/targets/awssqs/main.go
```diff
@@ -0,0 +1,295 @@
...
+	var awsSecret *corev1.Secret
+
+	BeforeEach(func() {
+		ns = f.UniqueName
+
+		gvr := targetAPIVersion.WithResource(targetResource)
+		trgtClient = f.DynamicClient.Resource(gvr).Namespace(ns)
+	})
+
+	Context("a target watches an existing queue", func() {
```
antoineco, line 86: A target doesn't really "watch" anything. I would keep the test context generic, something along the lines of "a target is deployed". Especially since the subcontext are describing quite well what we are testing ("When an event is sent to the target ... It puts a message onto the queue")

test/e2e/targets/awssqs/main.go
```diff
@@ -0,0 +1,295 @@
...
+			BeforeEach(func() {
+				By("sending an event", func() {
+					sentEvent = e2ece.NewEvent(f)
+
+					job := e2ece.RunEventSender(f.KubeClient, ns, trgtURL.String(), sentEvent)
+					apps.WaitForCompletion(f.KubeClient, job)
+				})
+			})
+
+			It("forwards the event", func() {
```
antoineco, line 134: Instead of "forwards the event", I'd try to be more specific like "puts a message onto the queue", just to make it extra clear what is supposed to happen from the user's perspective.

antoineco (commented):
Looking super clean as usual 👍

I only have one request: could we please include a test case where [discardCloudEventContext](https://github.com/triggermesh/triggermesh/blob/b54f436c05457af4f26a3944c488b5dab1d233c8/config/301-awssqstarget.yaml#L91-L93) is switched to `true`? (`When("the CloudEvent context is discarded") ... It("only puts the event's data onto the queue")`)

test/e2e/framework/cloudevents/cloudevents.go
```diff
@@ -0,0 +1,109 @@
...
+const E2ECeExtension = "iotriggermeshe2e"
+
+const (
+	curlContainerName     = "curl"
+	curlContainerImage    = "curlimages/curl:latest"
+	eventSenderNamePrefix = "eventsender-"
+)
+
+// NewEvent generates a CloudEvent with dummy values.
+func NewEvent(f *framework.Framework) *cloudevents.Event {
```
cab105, line 46: This strikes me as more specific to the SQS target than something we can use in general.

cab105 (commented):
Left one minor nit, and do agree with Sire Antoine's comment on checking the discardCloudEventContext as that has been bug prone in the past.

test/e2e/framework/cloudevents/cloudevents.go
antoineco: Why that? Literally every target needs to be tested by sending a CloudEvent to it, and this helper accepts an arbitrary event to make it versatile.

In fact, this helper comes directly from test-infra.

antoineco: Correction: my comment was related to `RunEventSender()` (oops).

Regardless, every target that isn't an "integration" target will happily accept an arbitrary payload and store it (Elasticsearch, anything Kafka-based, Kinesis, SNS, SQS, S3, Event Hubs, Pub/Sub, ... the list goes on and on).

cab105: I do see your point, and can give you the same number of SaaS specific integrations that are looking for a specific target to use. Perhaps renaming this as `NewHelloEvent` to differentiate between the two use cases. Especially since some targets such as the S3 target can take a generic event _and_ a specific event of `io.triggermesh.awss3.object.put` (as per https://github.com/triggermesh/triggermesh/blob/bd69609b0c5d45a2a5c71f72239bc73d6e6904c5/pkg/apis/targets/v1alpha1/aws_s3_lifecycle.go#L27).

antoineco: I agree, `NewHelloEvent` is more descriptive.

The only purpose of this helper is really to avoid having to repeat `NewEvent ... SetType 
...` in every test when all we need is _some_ event. It's not meant to support _all_ target tests.

test/e2e/targets/awssqs/main.go
```diff
@@ -163,6 +163,51 @@ var _ = Describe("AWS SQS target", func() {
...
+						withDiscardCEContext(),
+					)
+					Expect(err).ToNot(HaveOccurred())
+
+					trgt = ducktypes.WaitUntilReady(f.DynamicClient, trgt)
+
+					trgtURL = ducktypes.Address(trgt)
+					Expect(trgtURL).ToNot(BeNil())
+				})
+			})
```
antoineco, line 182: The way this test suite is structured will lead to two `AWSSQSTarget` objects being created inside the same namespace.

```go
Context("a target is deployed", func() {

    BeforeEach(func() {

        // first object created here

    })

    When("the CloudEvent context is discarded", func() {

        BeforeEach(func() {

            // second object created here

        })
    })
})
```

To avoid that, we should move the first object creation inside its own nested `Context/When/Describe`:

```go
Context("a target is deployed", func() {

    BeforeEach(func() {

        // queue creation can stay here, since all tests require it

    })

    When("the spec contains default settings", func() {

        BeforeEach(func() {

            // only one object created here

        })
    })

    When("the CloudEvent context is discarded", func() {

        BeforeEach(func() {

            // only one object created here

        })
    })
})
```

antoineco (approved):

cab105 (approved):

PR #436, opened 2022-01-13
Document roles (IAM) used in Azure E2E tests

antoineco:
Adds the definition of all the roles that were created in our test Azure account to enable running E2E tests.

There is one role for the test orchestration itself ("TriggerMesh E2E test suite"), and one role per event source component.

**Note:** Each individual source role can be used independently from the E2E test suite. For example, simply assign the role "Blob Storage event source" to be able to run the TriggerMesh event source for Azure Blob Storage, etc.


antoineco:
Marking as DRAFT.

The test suite for Azure IoT Hub runs without issue, up to the point where we wait for a message to arrive and we never get one.
I have to investigate whether the test is flaky or if we're missing some permission.

antoineco:
Marking as Ready.

The issue with the IoT Hub test was in fact due to a race (#442).

test/e2e/iam/azure_iam_roles.jsonc
```diff
@@ -0,0 +1,177 @@
...
+      "/subscriptions/52e4a8f0-1968-40a4-b38b-f19fac1e89ce"
+    ]
+  },
+  {
+    "roleName": "Activity Logs event source",
+    "description": "Role suitable for use with the TriggerMesh event source for Activity Logs.",
+    "permissions": [
+      {
+        "actions": [
+          "Microsoft.EventHub/namespaces/authorizationRules/listkeys/action",
```
cab105, line 47: Is this still required as I had removed the listkeys on the EventHub test.

cab105 (approved):
Just one question based on my previous PR, but lgtm.

test/e2e/iam/azure_iam_roles.jsonc
antoineco: It is. This is the `Activity Logs event source` role, it contains everything required to run the **source**, not the E2E test orchestration.

Azure requires this permission while interacting with the Activity Logs "diagnostic settings" API to be able to perform the necessary wiring on the Event Hubs side.
In API calls, you pass the ID of the rule that should be used, and Activity Logs uses the aforementioned permission to read it:

https://github.com/triggermesh/triggermesh/blob/06a157cadcd1098d7d5daa3a22a94fa090f3d6bb/pkg/sources/reconciler/azureactivitylogssource/diagsettings.go#L163-L164

The requirement is also documented here:

https://github.com/triggermesh/triggermesh/blob/06a157cadcd1098d7d5daa3a22a94fa090f3d6bb/pkg/sources/reconciler/azureactivitylogssource/diagsettings.go#L62-L66

cab105: Gotcha. Was under the impression this was more for the test framework itself.

antoineco: I tried to separate them to make it clear what is used by the test logic, and what is used by each TriggerMesh components, but it's not easy because you can't [group permissions like in AWS](https://github.com/triggermesh/triggermesh/blob/06a157cadcd1098d7d5daa3a22a94fa090f3d6bb/test/e2e/aws_iam_policy.jsonc#L10-L12), so overlaps aren't obvious.

PR #440, opened 2022-01-14
Refactor e2e test suite for Azure Activity Logs source

antoineco:
I would like to propose the following improvements to the test suite for the Azure Activity Logs source:

1. Reorganize `Context` blocks to have a more logical flow and more targeted locations of setup/cleanup steps.
   - Now the test flow reads as:
       - "a source subscribes to activities from an Azure subscription"
           - "an activity is logged inside the subscription"
               - "the source generates an event
    - To me, it reads more logically than the previous flow, which was:
        - "a source watches an EventHub publishing Activity Log data"
            (_Note: at a high level, the source **watches and reports activities**, the rest is implementation detail._)
            - "an event flows"
                (_Note: "event" can be a confusing way to refer to an Azure activity, due to our own definition of "event"._)
                - "should create an azure eventhub"
                    (_Note: confusing for the reader, that's not what the source does at all._)

1. Add some `By()` blocks inside setup/cleanup to make Ginkgo print the different steps in verbose mode as the test unfolds.

1. Re-scope variables to ensure variables are declared only where necessary.
    - For example, some of the variables are only local to `BeforeEach` blocks, some others aren't accessed by any "bad spec" `Context`, etc.

1. Rename a few variables and functions to ensure the difference between names, resource IDs, etc. is clear.
    - For example: `ns` -> `eventHubsNamespaceID`


antoineco:
A test run in verbose mode:

image

test/e2e/sources/azureactivitylogs/main.go
```diff
@@ -50,11 +48,11 @@ import (
   - AZURE_TENANT_ID - Azure tenant to create the resources against
   - AZURE_CLIENT_ID - The Azure ServicePrincipal Client ID
   - AZURE_CLIENT_SECRET - The Azure ServicePrincipal Client Secret
-  - AZURE_REGION - Define the Azure region to run the test (default uswest2)
+  - AZURE_REGION - Define the Azure region to run the test (default "westus2")
```
cab105, line 51: Good catch! Knew I had missed one.

cab105 (approved):
Does seem a little more concise than my first cut at an Azure test.

PR #447, opened 2022-01-17
Document IAM permissions for Kinesis target test

antoineco:
The target test introduced in #439 requires some [extra permissions](https://github.com/triggermesh/triggermesh/runs/4843697215?check_suite_focus=true), which are documented in this PR.

### DIFF
--- a/test/e2e/iam/aws_iam_policy.jsonc
+++ b/test/e2e/iam/aws_iam_policy.jsonc
@@ -57,7 +57,10 @@
                 "kinesis:PutRecord",
                 "kinesis:DeleteStream",
                 "kinesis:CreateStream",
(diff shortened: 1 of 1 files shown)


test/e2e/iam/aws_iam_policy.jsonc
```diff
@@ -129,8 +132,7 @@
             "Action": [
                 "sqs:GetQueueUrl",
                 "sqs:ReceiveMessage",
-                "sqs:DeleteMessage",
-                "sqs:DeleteMessageBatch"
```
antoineco, line 133: AWS was complaining that this permission doesn't exist, so I removed it here too.

PR #471, opened 2022-01-19
Add AWS S3 Target e2e test 

### LINKED ISSUE #470: Add E2E tests for Amazon S3 target

FranBarrera:
Closes #470

### DIFF
--- a/test/e2e/iam/aws_iam_policy.jsonc
+++ b/test/e2e/iam/aws_iam_policy.jsonc
@@ -97,12 +97,12 @@
             "Action": [
                 "s3:CreateBucket",
                 "s3:GetObject",
-                "s3:ListObjects",
+                "s3:ListBucket",
                 "s3:DeleteBucket",
                 "s3:DeleteObject"
             ],
             "Resource": [
-                "arn:aws:s3:*:043455440429:e2e-*"
+                "arn:aws:s3:::e2e-*"
             ]
         },
         {
@@ -197,7 +197,7 @@
             "Action": [
                 "s3:PutObject"
             ],
-            "Resource": "arn:aws:s3:*:043455440429:e2e-*"
+            "Resource": "arn:aws:s3:::e2e-*"
         }
     ]
 }
--- a/test/e2e/iam/aws_iam_policy.jsonc
+++ b/test/e2e/iam/aws_iam_policy.jsonc
@@ -92,6 +92,20 @@
             ]
         },
         {
+            "Sid": "E2EFrameworkS3",
+            "Effect": "Allow",
+            "Action": [
+                "s3:CreateBucket",
+                "s3:GetObject",
+                "s3:ListObjects",
+                "s3:DeleteBucket",
+                "s3:DeleteObject"
+            ],
+            "Resource": [
+                "arn:aws:s3:*:043455440429:e2e-*"
+            ]
+        },
+        {
             "Sid": "AWSCodeCommitSourceReceiveAdapter",
             "Effect": "Allow",
             "Action": [
@@ -176,6 +190,14 @@
                 "dynamodb:GetRecords"
             ],
             "Resource": "arn:aws:dynamodb:*:043455440429:table/e2e-*/stream/*"
+        },
+        {
+            "Sid": "AWSS3TargetReceiveAdapter",
+            "Effect": "Allow",
+            "Action": [
+                "s3:PutObject"
+            ],
+            "Resource": "arn:aws:s3:*:043455440429:e2e-*"
         }
     ]
 }
--- a/test/e2e/targets/awss3/main.go
+++ b/test/e2e/targets/awss3/main.go
@@ -200,7 +200,7 @@ var _ = Describe("AWS S3 target", func() {
 					})
 				})
 
-				It("only puts the event's data onto the record", func() {
+				It("only puts the event's data onto the bucket object", func() {
 					var receivedObj []byte
 					var err error
 
@@ -213,7 +213,7 @@ var _ = Describe("AWS S3 target", func() {
 						receivedObj, err = ioutil.ReadAll(object.Body)
 						Expect(err).ToNot(HaveOccurred())
 					})
-					By("inspecting the record payload", func() {
+					By("inspecting the object payload", func() {
 						Expect(receivedObj).To(Equal(sentEvent.Data()))
 					})
 				})
--- a/test/e2e/targets/awss3/main.go
+++ b/test/e2e/targets/awss3/main.go
@@ -1,0 +1,364 @@
+/*
+Copyright 2022 TriggerMesh Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package awss3
+
+import (
+	"context"
+	"encoding/json"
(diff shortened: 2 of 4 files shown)


test/e2e/targets/awss3/main.go
```diff
@@ -0,0 +1,364 @@
...
+	f := framework.New("awss3target")
+	region := os.Getenv("AWS_REGION")
+
+	var ns string
+
+	var trgtClient dynamic.ResourceInterface
+
+	var bucketARN string
+	var bucketName string
+	var awsCreds credentials.Value
```
antoineco, line 78: These two are only used inside the scope of `Context("a target is deployed"`, so they could be moved there instead of at the top scope.

test/e2e/targets/awss3/main.go
```diff
@@ -0,0 +1,364 @@
...
+			})
+		})
+	})
+
+	When("a client creates a target object with invalid specs", func() {
+
+		// Those tests do not require a real bucketARN or awsSecret
+		BeforeEach(func() {
+			bucketARN = "arn:aws:s3:eu-central-1:000000000000:test"
+			awsSecret = createAWSCredsSecret(f.KubeClient, ns, awsCreds)
```
antoineco, line 229: (related to my previous comment because it involves the `awsCreds` variable)

Creating a secret isn't required here since we're just testing the spec _validation_. We could just initialize `awsSecret` with an empty secret:

```golang
awsSecret = &corev1.Secret{}
```

test/e2e/targets/awss3/main.go
```diff
@@ -0,0 +1,364 @@
...
+				)
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(ContainSubstring(
+					"spec.awsApiSecret: Required value, spec.awsApiKey: Required value"))
+			})
+		})
+	})
+})
+
+// createTarget creates an AWSS3 object initialized with the given options.
```
antoineco, line 270: AWSS3Target

test/e2e/targets/awss3/main.go
```diff
@@ -0,0 +1,364 @@
...
+
+					trgt = ducktypes.WaitUntilReady(f.DynamicClient, trgt)
+
+					trgtURL = ducktypes.Address(trgt)
+					Expect(trgtURL).ToNot(BeNil())
+				})
+			})
+
+			When("an event is sent to the target", func() {
+				var sentEvent *cloudevents.Event
```
antoineco, line 127: Instead of declaring this variable twice (here, and in the sibling Context), it could be moved under `Context("a target is deployed"`

test/e2e/targets/awss3/main.go
```diff
@@ -0,0 +1,364 @@
...
+						job := e2ece.RunEventSender(f.KubeClient, ns, trgtURL.String(), sentEvent)
+						apps.WaitForCompletion(f.KubeClient, job)
+					})
+				})
+
+				It("creates an object onto the bucket", func() {
+					var receivedObj []byte
+					var err error
+
+					By("polling the bucket objects", func() {
```
antoineco, line 142: nit: we aren't really polling anything. "listing" is more accurate.

test/e2e/targets/awss3/main.go
```diff
@@ -0,0 +1,364 @@
...
+						job := e2ece.RunEventSender(f.KubeClient, ns, trgtURL.String(), sentEvent)
+						apps.WaitForCompletion(f.KubeClient, job)
+					})
+				})
+
+				It("only puts the event's data onto the bucket object", func() {
+					var receivedObj []byte
+					var err error
+
+					By("polling the bucket objects", func() {
```
antoineco, line 207: Same comment about "polling" vs "listing".

antoineco (approved):
A few nits, but everything looks great.

antoineco (commented):
By the way, we might want to add a `E2EFrameworkS3` and a `AWSS3TargetReceiveAdapter` blocks somewhere in [test/e2e/iam/aws_iam_policy.jsonc](https://github.com/triggermesh/triggermesh/blob/60d2bcda6835921bebce0278bc41cb0f109bb36a/test/e2e/iam/aws_iam_policy.jsonc#L9-L22) so I can easily keep our IAM role up-to-date.

If you're unsure about the required permissions, you can check the names of [supported operations](https://docs.aws.amazon.com/AmazonS3/latest/API/API_Operations.html) and match them with the API calls performed inside the test and adapter.

test/e2e/targets/awss3/main.go
```diff
@@ -0,0 +1,364 @@
...
+						receivedObj, err = ioutil.ReadAll(object.Body)
+						Expect(err).ToNot(HaveOccurred())
+					})
+					By("inspecting the object payload", func() {
+						Expect(receivedObj).To(Equal(sentEvent.Data()))
+					})
+				})
+			})
+		})
+	})
```
cab105, line 222: There is a third mode of operation for the s3 target where an event of type `io.triggermesh.awss3.object.put` can also put data into a bucket without the CE metadata.

FranBarrera: @cab105 Can you describe a bit more this operation? Not sure If I'm understanding

antoineco: (answering because Chris is probably not up yet 😴)

When `discardCEContext` is `false` (the default), a sender may still indicate that a specific event should be trimmed of its CloudEvent context by sending an event with `event.SetType("io.triggermesh.awss3.object.put")`.

E.g. imagine the following event:

```
Context
  type: io.triggermesh.awss3.object.put
  source: myapp
  id: 00000
Data
  {"orderId": 432748, "item": "Nicolas Cage pillow case"}
```

Only the `Data` part will be stored, regardless of the value of `io.triggermesh.awss3.object.put`, because the event is using the "magic" event type.

FranBarrera: Gotcha! thanks Antoine!

cab105: Think @antoineco meant that the `Data` part will be stored regardless of the state of `discardCEContext` when the `io.triggermesh.awss3.object.put` event type is used.

test/e2e/iam/aws_iam_policy.jsonc
```diff
@@ -91,6 +91,20 @@
                 "arn:aws:dynamodb:*:043455440429:table/e2e-*"
             ]
         },
+        {
+            "Sid": "E2EFrameworkS3",
+            "Effect": "Allow",
+            "Action": [
+                "s3:CreateBucket",
+                "s3:GetObject",
+                "s3:ListObjects",
```
antoineco, line 100: I'm seeing a warning for that particular action:

image

The doc says:

> To use this action in an AWS Identity and Access Management (IAM) policy, you must have permissions to perform the `s3:ListBucket` action.
> -- https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html

test/e2e/iam/aws_iam_policy.jsonc
```diff
@@ -91,6 +91,20 @@
...
+                "s3:DeleteBucket",
+                "s3:DeleteObject"
+            ],
+            "Resource": [
+                "arn:aws:s3:*:043455440429:e2e-*"
```
antoineco, line 105: image

image


```suggestion
                "arn:aws:s3:::e2e-*"
```

test/e2e/iam/aws_iam_policy.jsonc
```diff
@@ -176,6 +190,14 @@
...
             ],
             "Resource": "arn:aws:dynamodb:*:043455440429:table/e2e-*/stream/*"
+        },
+        {
+            "Sid": "AWSS3TargetReceiveAdapter",
+            "Effect": "Allow",
+            "Action": [
+                "s3:PutObject"
+            ],
+            "Resource": "arn:aws:s3:*:043455440429:e2e-*"
```
antoineco, line 200:
```suggestion
            "Resource": "arn:aws:s3:::e2e-*"
```

Issue #475, opened 2022-01-20
Implement `K_SINK` support in XSLTTransformation


JeffNeff:
Currently, they do not. It is in my opinion that they should.

PR #476, opened 2022-01-20
Implement `K_SINK` in XMLToJSONTransformation

JeffNeff:
Closes part of issue #475


pkg/flow/adapter/xmltojsontransformation/adapter.go
```diff
@@ -104,9 +104,20 @@ func (a *Adapter) dispatch(ctx context.Context, event cloudevents.Event) (*cloud
 		return a.replier.Error(&event, targetce.ErrorCodeAdapterProcess, err, nil)
 	}
 
+	if err := a.emitToSink(ctx, event); err != nil {
+		return a.replier.Error(&event, targetce.ErrorCodeAdapterProcess, err, nil)
+	}
+
 	return &event, cloudevents.ResultACK
```
antoineco, line 111: Do I understand correctly that we both send the event to the sink _and_ respond with the event?

If yes, we probably want to short circuit the last `return` in case a sink was provided (I'm assuming the sink parameter is optional, not sure what the intention was?)

pkg/flow/reconciler/xmltojsontransformation/controller.go
```diff
@@ -56,7 +58,11 @@ func NewController(
 	}
 
 	impl := reconcilerv1alpha1.NewImpl(ctx, r)
-	targetInformer.Informer().AddEventHandler(controller.HandleAll(impl.Enqueue))
+	logging.FromContext(ctx).Info("Setting up event handlers")
```
antoineco, line 61: Logging this feels very superfluous.
If we start doing that in every component, expect a flood of unnecessary log entries on every startup.

pkg/flow/reconciler/xmltojsontransformation/reconciler.go
```diff
@@ -39,10 +44,29 @@ type Reconciler struct {
 var _ reconcilerv1alpha1.Interface = (*Reconciler)(nil)
 
 // ReconcileKind implements Interface.ReconcileKind.
-func (r *Reconciler) ReconcileKind(ctx context.Context, trg *v1alpha1.XMLToJSONTransformation) pkgreconciler.Event {
-	adapter, event := r.ksvcr.ReconcileKService(ctx, trg, makeTargetAdapterKService(trg, r.adapterCfg))
+func (r *Reconciler) ReconcileKind(ctx context.Context, s *v1alpha1.XMLToJSONTransformation) pkgreconciler.Event {
+	s.Status.InitializeConditions()
+	s.Status.ObservedGeneration = s.Generation
+
```
antoineco, line 50: These 2 steps are performed by Knative, we can remove them from here.

pkg/flow/adapter/xmltojsontransformation/adapter.go
JeffNeff: hey just saw this! This was the main reason i pulled this back as a draft. I wanted to discuss how to handle this in the eng. meeting this morning with @cab105 and @odacremolbap .. I have now updated it to respond if no sink is specified.

pkg/flow/reconciler/xmltojsontransformation/controller.go
JeffNeff: removed

JeffNeff: Followup issue created to remove them where they are https://github.com/triggermesh/triggermesh/issues/479

pkg/flow/reconciler/xmltojsontransformation/controller_test.go
```diff
@@ -1,39 +0,0 @@
-/*
```
cab105, line 1: Why delete this test?

pkg/flow/adapter/xmltojsontransformation/adapter.go
```diff
@@ -46,6 +47,10 @@ type envAccessor struct {
 	BridgeIdentifier string `envconfig:"EVENTS_BRIDGE_IDENTIFIER"`
 	// CloudEvents responses parametrization
 	CloudEventPayloadPolicy string `envconfig:"EVENTS_PAYLOAD_POLICY" default:"error"`
+	// Sink defines the target sink for the events. If no Sink is defined the
+	// events are replied back to the sender.
+	// +optional
+	Sink *string `envconfig:"K_SINK"`
```
cab105, line 53: Should make use of the `omitempty` annotation

pkg/flow/adapter/xmltojsontransformation/adapter.go
cab105: Looking at the reconciler logic, the controller will fail to bring up the adapter if `K_SINK` is not defined so unless it really is optional, then it is safe that `K_SINK` will always be present.

pkg/flow/reconciler/xmltojsontransformation/adapter.go
```diff
@@ -68,6 +68,10 @@ func makeAppEnv(o *v1alpha1.XMLToJSONTransformation) []corev1.EnvVar {
 			Name:  libreconciler.EnvBridgeID,
 			Value: libreconciler.GetStatefulBridgeID(o),
 		},
+		{
+			Name:  "K_SINK",
+			Value: o.Status.SinkURI.String(),
```
cab105, line 73: A nil check should be made as `o.Status.SinkURI` is defined as a `*String`, and can be empty.

cab105 (commented):
Did spot one inconsistency in how `K_SINK` is defined/used which will need to be resolved.

pkg/flow/adapter/xmltojsontransformation/adapter.go
antoineco: This is an envConfig struct, it doesn't get marshaled to JSON :)

pkg/flow/reconciler/xmltojsontransformation/controller_test.go
JeffNeff: I could not get it to work :/

cab105: What was failing for it? Note that if this test fails, then that means the adapter will fail to come up.

pkg/flow/adapter/xmltojsontransformation/adapter.go
JeffNeff: I have made updates to address this :)

pkg/flow/reconciler/xmltojsontransformation/controller_test.go
JeffNeff:
```    /Users/jeffreynaef/go/src/github.com/triggermesh/triggermesh/pkg/flow/reconciler/xmltojsontransformation/logger.go:130: 2022-01-21T09:32:38.863-0700	PANIC	addressable/addressable.go:56	Unable to fetch knative.dev/pkg/apis/duck.InformerFactory from context.
        /Users/jeffreynaef/go/src/github.com/triggermesh/triggermesh/pkg/flow/reconciler/xmltojsontransformation/controller.go:61: Unexpected panic: Unable to fetch knative.dev/pkg/apis/duck.InformerFactory from context.
```

pkg/apis/flow/v1alpha1/xmltojsontransformation_lifecycle.go
```diff
@@ -29,7 +29,18 @@ const (
...
-var xmlToJSONCondSet = apis.NewLivingConditionSet()
+const (
+	// ConditionReady has status True when the router is ready to send events.
+	ConditionReady = apis.ConditionReady
+	// ConditionSinkProvided has status True when the router has been configured with a sink target.
+	ConditionSinkProvided apis.ConditionType = "SinkProvided"
+)
+
+var XMLToJSONCondSet = apis.NewLivingConditionSet(
+	ConditionReady,
```
antoineco, line 40: Unnecessary, `Ready` is already the default condition embedded in any Knative `ConditionSet`.

Example from another API group (notice that we don't explicitly declare the "Ready" condition):
https://github.com/triggermesh/triggermesh/blob/a7b8255c31533c309b24bb1e5f56222f1add9c38/pkg/apis/sources/v1alpha1/common_lifecycle.go#L64-L67

pkg/apis/flow/v1alpha1/xmltojsontransformation_lifecycle.go
```diff
@@ -29,7 +29,18 @@ const (
...
 
-var xmlToJSONCondSet = apis.NewLivingConditionSet()
+const (
+	// ConditionReady has status True when the router is ready to send events.
+	ConditionReady = apis.ConditionReady
+	// ConditionSinkProvided has status True when the router has been configured with a sink target.
+	ConditionSinkProvided apis.ConditionType = "SinkProvided"
+)
+
+var XMLToJSONCondSet = apis.NewLivingConditionSet(
```
antoineco, line 39: Does this type need to be exported?

pkg/apis/flow/v1alpha1/xmltojsontransformation_lifecycle.go
```diff
@@ -29,7 +29,18 @@ const (
 	EventTypeXMLToJSONGenericResponse = "io.triggermesh.xmltojsontransformation.error"
 )
 
-var xmlToJSONCondSet = apis.NewLivingConditionSet()
+const (
+	// ConditionReady has status True when the router is ready to send events.
```
antoineco, line 33: ~router~ component

pkg/apis/flow/v1alpha1/xmltojsontransformation_lifecycle.go
```diff
@@ -29,7 +29,18 @@ const (
...
+	ConditionReady = apis.ConditionReady
+	// ConditionSinkProvided has status True when the router has been configured with a sink target.
```
antoineco, line 35: ~router~ component

pkg/apis/flow/v1alpha1/xmltojsontransformation_lifecycle.go
```diff
@@ -84,11 +95,39 @@ func (ts *XMLToJSONTransformationStatus) PropagateKServiceAvailability(ksvc *ser
...
+
+// MarkNoSink sets the SinkProvided condition to False.
+func (ts *XMLToJSONTransformationStatus) MarkNoSink() {
+	ts.SinkURI = nil
+	XMLToJSONCondSet.Manage(ts).MarkFalse(ConditionSinkProvided,
+		ReasonSinkNotFound, "The sink does not exist or its URI is not set")
+}
+
+// MarkNoKService sets the condition that the service is not ready
+func (ts *XMLToJSONTransformationStatus) MarkNoKService(reason, messageFormat string, messageA ...interface{}) {
```
antoineco, line 126: nit: This function is called `MarkKService` but sets a condition that's called `Deployed`.
For consistency, I'd suggest we call it `MarkAdapterDeployed` (or `MarkDeployed` if you like short names, no preference on my side).

pkg/apis/flow/v1alpha1/xmltojsontransformation_lifecycle.go
```diff
@@ -38,32 +49,32 @@ func (t *XMLToJSONTransformation) GetGroupVersionKind() schema.GroupVersionKind
...
 func (ts *XMLToJSONTransformationStatus) MarkServiceUnavailable(name string) {
-	xmlToJSONCondSet.Manage(ts).MarkFalse(
+	XMLToJSONCondSet.Manage(ts).MarkFalse(
 		apis.ConditionReady,
 		"ServiceUnavailable",
 		"Service %q is not ready.", name)
 }
 
 // MarkServiceAvailable sets XMLToJSONTransformation condition to ready.
 func (ts *XMLToJSONTransformationStatus) MarkServiceAvailable() {
```
antoineco, line 69: The Ready condition should never be set by us. Knative will set it automatically based on the other conditions in the ConditionSet.

pkg/apis/flow/v1alpha1/xmltojsontransformation_lifecycle.go
```diff
@@ -38,32 +49,32 @@ func (t *XMLToJSONTransformation) GetGroupVersionKind() schema.GroupVersionKind
...
 }
 
 // InitializeConditions sets the initial values to the conditions.
 func (ts *XMLToJSONTransformationStatus) InitializeConditions() {
-	xmlToJSONCondSet.Manage(ts).InitializeConditions()
+	XMLToJSONCondSet.Manage(ts).InitializeConditions()
 }
 
 // MarkServiceUnavailable marks XMLToJSONTransformation as not ready with ServiceUnavailable reason.
 func (ts *XMLToJSONTransformationStatus) MarkServiceUnavailable(name string) {
```
antoineco, line 61: Likewise, we should never set this condition ourselves. Cf. Knative libs.

pkg/flow/adapter/xmltojsontransformation/adapter.go
```diff
@@ -46,6 +47,10 @@ type envAccessor struct {
 	BridgeIdentifier string `envconfig:"EVENTS_BRIDGE_IDENTIFIER"`
 	// CloudEvents responses parametrization
 	CloudEventPayloadPolicy string `envconfig:"EVENTS_PAYLOAD_POLICY" default:"error"`
+	// Sink defines the target sink for the events. If no Sink is defined the
+	// events are replied back to the sender.
+	// +optional
```
antoineco, line 52: nit: This is a Kubernetes convention, but this struct is not related to a Kubernetes API.
In envConfig, things are always optional unless the `required` tag is set.

pkg/flow/adapter/xmltojsontransformation/adapter.go
antoineco, line 53: Someone could pass an empty string here and the check for `nil` below wouldn't work anymore. envConfig isn't as strict as Kubernetes' OpenAPI contracts.

I feel it would be simpler to just make that a string an check for `""`.

pkg/flow/adapter/xmltojsontransformation/adapter.go
```diff
@@ -104,6 +111,15 @@ func (a *Adapter) dispatch(ctx context.Context, event cloudevents.Event) (*cloud
 		return a.replier.Error(&event, targetce.ErrorCodeAdapterProcess, err, nil)
 	}
 
+	fmt.Println(a.sink)
```
antoineco, line 114: Leftover from dev.

antoineco (commented):
Partial review, I still need to review the tests more thoroughly.

pkg/apis/flow/v1alpha1/xmltojsontransformation_lifecycle.go
JeffNeff: you know... i wondered the same thing. I copied a lot of this from `Synchronizer` he had it exported. so I left it exported.

pkg/apis/flow/v1alpha1/xmltojsontransformation_lifecycle.go
JeffNeff: is this method even being used? Can it just be removed?

cab105 (approved):

pkg/apis/flow/v1alpha1/xmltojsontransformation_lifecycle.go
antoineco: Ok but, does it need to be? Is anything outside of this package using it?

JeffNeff: nope! I have fixed it .

PR #478, opened 2022-01-20
Introduce startup probe in AWS sources' adapters

### LINKED ISSUE #418: Introduce startup probe in source adapters
Second take on #413.

We want to ensure that the adapter container is able to interact with the AWS API before marking it as "Ready". This can be implemented using a health HTTP endpoint that is polled by Kubernetes (readiness/startup probe).

As mentioned in the aforementioned issue, we can't simulate or dry-run API calls to ensure that we have all sufficient permissions, but we can probe AWS when the application starts by asking it whether the observed resource exists. Besides confirming that the resource does exist, it indirectly validates that the provided credentials are valid.

Examples:
- SQS: check that the queue exists (already implemented)
- CodeCommit: check that the repo exists
- DynamoDB: check that a stream exists for the table (indirectly tells us that the table exists)
- Kinesis: check that the stream exists
- SNS: https://github.com/triggermesh/triggermesh/issues/29#issuecomment-930445386
- ...

antoineco:
Closes #418

Prevents sources from becoming ready if the provided credentials aren't valid, or if the provided resource doesn't exist.

As a side effect, allows running all these adapters as long-running Knative Services in Google Cloud Run (because poking port 8080 will now return a response).

### DIFF
--- a/pkg/routing/reconciler/common/resource/container.go
+++ b/pkg/routing/reconciler/common/resource/container.go
@@ -163,10 +163,6 @@ func Probe(path, port string) ObjectOption {
 					Port: intstrPort,
 				},
 			},
-			// TODO(antoineco): remove delay after switching to StartupProbe, which is enabled by default
-			// starting with Kubernetes 1.18.
-			// Ref. https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
-			InitialDelaySeconds: 2,
 		}
 	}
 }
--- a/pkg/routing/reconciler/common/resource/container_test.go
+++ b/pkg/routing/reconciler/common/resource/container_test.go
@@ -82,7 +82,6 @@ func TestNewContainer(t *testing.T) {
 					Port: intstr.FromString("health"),
 				},
 			},
-			InitialDelaySeconds: 2,
 		},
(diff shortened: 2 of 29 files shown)


JeffNeff:
@antoineco this is on my list for tonight/tomorrow to test in GCR

JeffNeff:
looks like we are missing Performance Insights source

JeffNeff:
SNS is intentionally excluded because :
```
The SNS source is multi-tenant, so the logic of the startup probe is different and more involved: it needs to read the informer's lister, and verify that a HTTP handler was registered for each AWSSNSSource in the namespace. I drafted the logic here, regardless that source won't run outside of Kubernetes (so, no Cloud Run),
```

antoineco:
@JeffNeff I added a probe to the Performance Insights source in [982902ab..617ce146](https://github.com/triggermesh/triggermesh/compare/982902abeb9b452776eaa82d90b076f49fa26747..617ce146dc48a6cb6a17c76c3823f93d507ba30d#diff-af6b3baaf04e430fd894e7e5ee4b8e804676ad4500e75d1faf86eab795dc7669).

JeffNeff (approved):
lgtm!! everything seems to be in order on cloudrun

PR #484, opened 2022-01-21
Create E2E Test for XMLToJSONTransformation

JeffNeff:


test/e2e/e2e_suite_test.go
```diff
@@ -28,8 +28,8 @@ import (
 	_ "github.com/triggermesh/triggermesh/test/e2e/framework"
 
 	// test suites
-	_ "github.com/triggermesh/triggermesh/test/e2e/sources"
-	_ "github.com/triggermesh/triggermesh/test/e2e/targets"
+	// _ "github.com/triggermesh/triggermesh/test/e2e/sources"
```
antoineco: There is a simpler way to work on a single test: use `FDescribe()` inside your Ginkgo test.
"F" stands for Focus, it runs only the focused series of tests.

antoineco:
Marking as draft. The test isn't reviewable yet. (Commented code, use of hacks instead of proposing a robust change to the test framework, etc.)

test/e2e/transformations/xmltojson/main.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+const (
+	transformationKind     = "XMLToJSONTransformation"
+	transformationResource = "xmltojsontransformations"
+
+	expectedResponseEvent = "{\"string\":\"\\u003cnote\\u003e\\u003cto\\u003eTove\\u003c/to\\u003e\\u003cfrom\\u003eJani\\u003c/from\\u003e\\u003cheading\\u003eReminder\\u003c/heading\\u003e\\u003cbody\\u003eDont forget me this weekend\\u003c/body\\u003e\\u003c/note\\u003e\"}"
+)
+
+// createTransformation creates an AWSKinesis object initialized with the given options.
```
antoineco, line 55: Copy-pasta.

Also as a general good practice it would be nice to move those helper functions _after_ the test specs, so that we see the tests first, and not the helpers.

test/e2e/transformations/xmltojson/main.go
```diff
@@ -0,0 +1,158 @@
...
+
+	return trnsClient.Create(context.Background(), trns, metav1.CreateOptions{})
+}
+
+// readReceivedEvents returns a function that reads CloudEvents received by the
+// event-display application and stores the result as the value of the given
+// `receivedEvents` variable.
+// The returned function signature satisfies the contract expected by
+// gomega.Eventually: no argument and one or more return values.
+func readReceivedEvents(c clientset.Interface, namespace, eventDisplayDeplName string,
```
antoineco, line 77: As a general good practice it would be nice to move those helper functions _after_ the test specs, so that we see the tests first, and not the helpers.

test/e2e/transformations/xmltojson/main.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+}
+
+var _ = Describe("XMLToJSON Transformation", func() {
+	f := framework.New("xmltojsontransformation")
+	var ns string
+	var sink *duckv1.Destination
+	var trnsClient dynamic.ResourceInterface
+	var trans *unstructured.Unstructured
+	var err error
```
antoineco, line 95: This `err` variable seems to only be used in one very specific `By`, and not across spec blocks, so it could be moved there instead of at the top scope.

test/e2e/transformations/xmltojson/main.go
```diff
@@ -0,0 +1,158 @@
...
+var transAPIVersion = schema.GroupVersion{
+	Group:   "flow.triggermesh.io",
+	Version: "v1alpha1",
+}
+
+const (
+	transformationKind     = "XMLToJSONTransformation"
+	transformationResource = "xmltojsontransformations"
+
+	expectedResponseEvent = "{\"string\":\"\\u003cnote\\u003e\\u003cto\\u003eTove\\u003c/to\\u003e\\u003cfrom\\u003eJani\\u003c/from\\u003e\\u003cheading\\u003eReminder\\u003c/heading\\u003e\\u003cbody\\u003eDont forget me this weekend\\u003c/body\\u003e\\u003c/note\\u003e\"}"
```
antoineco, line 52: This doesn't seem right. The transformation is supposed to convert structured XML data into structured JSON data, not stringify an entire XML and wrap it into a JSON.

antoineco (commented):
By looking at the "expected data" it doesn't seem like the transformation is working as expected.
If we have an XML document as input, we should have the same structured data represented as a JSON as output.

If the transformation is indeed broken, we should still have proper assertions inside the test and use them to verify that the transformation is fixed, not the other way around.

**Update**: see my comment in the review below (`xmltojson/main.go`)

test/e2e/transformations/xmltojson/main.go
antoineco: I think I know what happened.

`e.SetData()` assumes it receives structured data, and it tries to encode it to the given content type ([Godoc](https://pkg.go.dev/github.com/cloudevents/sdk-go/v2@v2.8.0/event#Event.SetData)).
However, you're giving it a string of pre-encoded data, so you don't want that behavior.

Try passing that data as `[]byte` instead of a string, it should help. (We might have to indicate that it's not base64 too, not 100% sure)

antoineco: Looks like passing the data as `[]byte` is working: https://go.dev/play/p/Dp9U_dJGiyJ

JeffNeff: I dont understand.. im not passing a string here I am passing in `readBuf` that is of type []byte


https://github.com/triggermesh/triggermesh/blob/d5d48a14c953d7ef1b218a42e3a519c091e7f767/pkg/flow/adapter/xmltojsontransformation/adapter.go#L97

https://github.com/triggermesh/triggermesh/blob/d5d48a14c953d7ef1b218a42e3a519c091e7f767/pkg/flow/adapter/xmltojsontransformation/adapter.go#L108

antoineco: I'm talking about the test and only the test.
Look at the `expectedResponseEvent`, I do hope you can see it's not the JSON representation of your XML, and the reason for that is the way you crafted your "hello" XML event.

test/e2e/transformations/xmltojson/main.go
```diff
@@ -0,0 +1,173 @@
...
+})
+
+// newXMLHelloEvent generates a CloudEvent with dummy values and an XML data payload.
+func newXMLHelloEvent(f *framework.Framework) *cloudevents.Event {
+	event := cloudevents.NewEvent()
+	event.SetID("0000")
+	event.SetType("e2e.test")
+	event.SetSource("e2e.triggermesh")
+	event.SetExtension("iotriggermeshe2e", f.UniqueName)
+	if err := event.SetData(cloudevents.ApplicationXML, `<note><to>Tove</to><from>Jani</from><heading>Reminder</heading><body>Dont forget me this weekend</body></note>`); err != nil {
```
antoineco, line 134: To my previous comment, the error is right here: a string is passed, but the SDK tries to convert that string to XML instead of accepting it as raw XML data.
Here is the fix: https://go.dev/play/p/Gg5gP15l240

test/e2e/transformations/xmltojson/main.go
JeffNeff: Thank you again!

test/e2e/transformations/xmltojson/main.go
JeffNeff: got it!

JeffNeff: Thank you!

antoineco (approved):

test/e2e/transformations/xmltojson/main.go
```diff
@@ -0,0 +1,264 @@
...
+				_ = transDst
+			})
+
+			By("creating a response intercepter", func() {
+				// TODO: create a helper that deploys a pre-configured
+				// Broker/Triggers/Event-Display bundle for intercepting event responses.
+				//
+				// When a message enters the Broker, it must be forwarded to the Transformation
+				// using a Trigger, and the response should be forwarded to the Event-Display
+				// using another Trigger.
```
antoineco, line 175: Note: I'm referring to https://github.com/triggermesh/triggermesh/issues/536

JeffNeff:
wow... amazing cleanup @antoineco.. this looks so much better. thank you and I will take notes.

PR #485, opened 2022-01-22
Filter AWS Performance Insights instance ID based on the ARN of the RDS instance

antoineco:
Instead of listing all instances inside the account, then iterating over them all, request just the instance we're interested in based on its ARN.

(On top of that, the current approach doesn't take pagination into account, so the DB instance we're looking for could be buried on another "page" and require subsequent API calls to retrieve.)

### DIFF
--- a/pkg/sources/adapter/awsperformanceinsightssource/adapter.go
+++ b/pkg/sources/adapter/awsperformanceinsightssource/adapter.go
@@ -107,17 +107,24 @@ func NewAdapter(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClie
 	}
 
 	r := rds.New(cfg)
-	var resourceID string
-	dbi, err := r.DescribeDBInstances(&rds.DescribeDBInstancesInput{})
+
+	dbi, err := r.DescribeDBInstances(&rds.DescribeDBInstancesInput{
+		Filters: []*rds.Filter{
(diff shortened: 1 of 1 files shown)


pkg/sources/adapter/awsperformanceinsightssource/adapter.go
```diff
@@ -107,18 +107,25 @@ func NewAdapter(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClie
...
 
 	r := rds.New(cfg)
-	var resourceID string
-	dbi, err := r.DescribeDBInstances(&rds.DescribeDBInstancesInput{})
+
+	dbi, err := r.DescribeDBInstances(&rds.DescribeDBInstancesInput{
+		Filters: []*rds.Filter{
+			{
+				Name:   aws.String("db-instance-id"),
+				Values: aws.StringSlice([]string{a.String()}),
```
antoineco, line 115: > `db-instance-id` - Accepts DB instance identifiers and DB instance Amazon Resource Names (ARNs). The results list will only include information about the DB instances identified by these ARNs.
>
> --  https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeDBInstances.html

JeffNeff (approved):
nice addition!

PR #487, opened 2022-01-23
Creates a Tool Enabling Debugging of the Package Replier 

JeffNeff:
This is just a thought on how we could create something simple to enable the debugging of our package cloudevent replier. Note that there is an implementation of this in [PR#484](https://github.com/triggermesh/triggermesh/pull/484)


antoineco:
This is a `curl` call, but coded in Go that prints the response using Printf. I personally don't see much value having this in the core of our codebase since `kubectl run ... -- curl -D- ...` would give you the exact same output.

Also, I see you dropped the code in the e2e framework, but as is I don't see how this could be useful to run e2e tests.

PR #489, opened 2022-01-24
AWS S3 Target - Standard ARN S3 Bucket

### LINKED ISSUE #473: AWS S3 Target - arn parameter is not an standard AWS S3 Bucket ARN
Right now we are requesting a bucket arn like `arn:aws:s3:<REGION>:<ACCOUNT_ID>:<BUCKET_NAME>`.  https://github.com/triggermesh/triggermesh/blob/main/config/301-awss3target.yaml#L91

Standard AWS S3 Bucket ARN: `arn:aws:s3:::bucket_name`
https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazons3.html

We should update our aws s3 target, removing the requirement for the `region` and `accountID` to allow the standard bucket arn

FranBarrera:
Closes #473


pkg/targets/adapter/awss3target/region.go
```diff
@@ -0,0 +1,47 @@
...
+// Per AWS conventions, a bucket which does not explicitly specify its location
+// is located in the "us-east-1" region.
+// https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketLocation.html
+const defaultS3Region = "us-east-1"
+
+// getBucketRegion retrieves the region the provided bucket resides in.
+func getBucketRegion(bucketName string, env *envAccessor) (string, error) {
+	sess := session.Must(session.NewSession(env.GetAwsConfig().
+		WithRegion(defaultS3Region).
+		WithMaxRetries(5)))
```
antoineco, line 33: nit: This is superfluous because the AWS SDK already has its own retry rules per service. I don't know if it's exactly "5", but unless care about it to be "5" I'd simply save one line of code.

config/301-awss3target.yaml
```diff
@@ -88,7 +88,7 @@ spec:
                 type: string
                 description: ARN of the S3 bucket that will receive events. The expected format is documented at
                   https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazons3.html
-                pattern: ^arn:aws(-cn|-us-gov)?:s3:[a-z]{2}(-gov)?-[a-z]+-\d:\d{12}:.+$
+                pattern: ^arn:aws(-cn|-us-gov)?:s3:::.+$
```
antoineco, line 91: nit: The actual naming requirement is slightly more restrictive than `.+`.
Being strict is a good way to prevent loops of errors early if a user passes input that's clearly not gonna work.

https://github.com/triggermesh/triggermesh/blob/3ce81f86e06a3d407c8b52029b133547c30b3234/config/300-awss3source.yaml#L72

antoineco (approved):
Approving with two minor comments.
I think having a correct regexp is beneficial here.
The retry is up to you to keep or remove.

PR #491, opened 2022-01-24
Document GetBucketLocation perm for S3 target adapter

antoineco:
This permission got omitted in #489 and the test failed as a result (I already fixed it).


FranBarrera (approved):
Thanks!

PR #499, opened 2022-01-25
Creates an adapter for the Dataweave Transformation

JeffNeff:
I wanted to post this just as an adapter for review prior to building out the rest of the boilerplate. I wanted to get input on how this is being implemented and if we should allow this into production or not.


antoineco (commented):
Binaries in Git are a big anti-pattern, and it dramatically increases the size of the repo.

I was suggesting in the implementations of `mq` components (which had a similar issue) that we have a separate repo where we maintain "special" base images (anything hybrid that requires external libs or binaries). @tzununbekov actually figured out another approach where we build and release those images inside our build pipeline, based on Dockerfiles, and it seems to be working well so far.

PR #503, opened 2022-01-25
Implement K_SINK in XSLTTransform

### LINKED ISSUE #475: Implement `K_SINK` support in XSLTTransformation

JeffNeff:
This PR implements K_SINK into the XSLTTransform object. Allowing users to define a sink to pass the transformed events along to. 

edit (@antoineco): closes #475


pkg/flow/reconciler/xslttransform/reconciler.go
```diff
@@ -35,17 +37,26 @@ type reconciler struct {
...
 }
 
 // Check that our Reconciler implements Interface
 var _ reconcilerv1alpha1.Interface = (*reconciler)(nil)
 
 // ReconcileKind implements Interface.ReconcileKind.
 func (r *reconciler) ReconcileKind(ctx context.Context, o *v1alpha1.XSLTTransform) pkgreconciler.Event {
-	o.Status.InitializeConditions()
-	o.Status.ObservedGeneration = o.Generation
+	var url string
```
antoineco, line 49: Why not passing the strongly typed `*apis.URL` returned by the resolver, untouched? It would allow you to simply compare against `nil`.

JeffNeff:
Im having the same issue I had with the XMLToJSONTransformation.. The controller test is failing due to 

```
Unexpected panic: Unable to fetch knative.dev/pkg/apis/duck.InformerFactory from context.
```

pkg/flow/adapter/xslttransform/adapter_test.go
```diff
@@ -256,6 +264,60 @@ func TestXSLTTransformEvents(t *testing.T) {
 	}
 }
 
+func TestXSLTTransformKSINK(t *testing.T) {
+	testCases := map[string]struct {
+		allowXSLTOverride bool
```
antoineco, line 269: Always `false` in the tests below, so it can be removed. A boolean is `false` by default.

pkg/flow/adapter/xslttransform/adapter_test.go
```diff
@@ -256,6 +264,60 @@ func TestXSLTTransformEvents(t *testing.T) {
...
+			expectedStringResponse: tExpectedResponseString,
+			expectCategory:         tSuccessAttribute,
+		},
+	}
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			svr := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				body, err := ioutil.ReadAll(r.Body)
+				assert.NoError(t, err)
+				assert.Equal(t, tc.expectedStringResponse, string(body))
```
antoineco, line 290: This looks clumsy to me.
The server should simply save the requests somewhere without handling the data (some slice that we append to), and assertions should be performed on these at the end of the test.

Even simpler: Knative has a mock CE client that records the events it "sends", and allows you to inspect these events. You can see it in action in tests for source adapters.

pkg/flow/adapter/xslttransform/adapter.go
```diff
@@ -139,6 +142,23 @@ func (a *xsltTransformAdapter) dispatch(ctx context.Context, event cloudevents.E
 			fmt.Errorf("eror processing XML with XSLT: %v", err), nil)
 	}
 
+	if isXML {
```
antoineco, line 145: Is this whole new conditional related to the K_SINK story?

JeffNeff: yes sir.

JeffNeff: yes. because we have to know if we are sending back an XML or JSON payload right?

pkg/flow/adapter/xslttransform/adapter_test.go
JeffNeff: I saw the mock CE Client in the sources.. but we have to set a K_SINK URL in order for them to send to the K_SINK.

JeffNeff: also.. the assertions are at the top.. however they are performed technically at the end, because that call is not made until the end of the test.

pkg/flow/adapter/xslttransform/adapter.go
antoineco: Ah, that's for the response 👍

antoineco:
Regarding the failing test, it's because you're missing something related to duck APIs in https://github.com/triggermesh/triggermesh/blob/54652bfd317467ece28dcf1f6f6a3269d1b94a3e/pkg/flow/reconciler/testing/factory.go

Compare with https://github.com/triggermesh/triggermesh/blob/54652bfd317467ece28dcf1f6f6a3269d1b94a3e/pkg/sources/reconciler/testing/factory.go maybe

pkg/flow/adapter/xslttransform/adapter_test.go
antoineco: But this test _is_ for the K_SINK scenario, right?
The mock client doesn't technically use the sink, because it's mocked.
I can't see why we couldn't use it here instead of starting a HTTP server.

JeffNeff: I got it figured out :) seems to be working okay.

pkg/flow/adapter/xslttransform/adapter_test.go
antoineco: Marking not resolved.

pkg/flow/adapter/xslttransform/adapter_test.go
```diff
@@ -256,6 +257,67 @@ func TestXSLTTransformEvents(t *testing.T) {
...
+
+		inEvent             cloudevents.Event
+		expectedEventString string
+		expectPanic         string
+		expectCategory      string
+	}{
+		"transform ok": {
+			xslt:                tXSLT,
+			inEvent:             newCloudEvent(tXML, cloudevents.ApplicationXML),
+			expectedEventString: "Context Attributes,\n  specversion: 1.0\n  type: ce.test.type\n  source: ce.test.source\n  id: ce-abcd-0123\n  datacontenttype: application/xml\nData (binary),\n  <?xml version=\"1.0\"?>\n<output>\n  <item>A1</item>\n  <item>B2</item>\n  <item>C3</item>\n</output>\n\n",
```
antoineco, line 273: That's very fragile. We should check attributes individually, and compare data to some fixed JSON payload, not stringify the entire event like that.
It might break on any update of the CloudEvent SDK. Even a single, unrelated character would break it.

pkg/flow/adapter/xslttransform/adapter_test.go
```diff
@@ -256,6 +257,67 @@ func TestXSLTTransformEvents(t *testing.T) {
...
 }
 
+func TestXSLTTransformKSINK(t *testing.T) {
+	testCases := map[string]struct {
+		allowXSLTOverride bool
+		xslt              string
+
+		inEvent             cloudevents.Event
+		expectedEventString string
+		expectPanic         string
```
antoineco, line 267: I do hope we don't expect panics to occur in production code.

pkg/flow/adapter/xslttransform/adapter_test.go
```diff
@@ -256,6 +257,67 @@ func TestXSLTTransformEvents(t *testing.T) {
...
+				logger: logtesting.TestLogger(t),
+
+				ceClient:     ceClient,
+				xsltOverride: false,
+				defaultXSLT:  style,
+				replier:      replier,
+				sink:         "http://localhost:8080",
+			}
+
+			_, r := a.Dispatch(ctx, tc.inEvent)
```
antoineco, line 310: Aren't we going to check the response while we're at it?

pkg/flow/reconciler/xslttransform/reconciler.go
```diff
@@ -35,17 +37,26 @@ type reconciler struct {
...
 var _ reconcilerv1alpha1.Interface = (*reconciler)(nil)
 
 // ReconcileKind implements Interface.ReconcileKind.
 func (r *reconciler) ReconcileKind(ctx context.Context, o *v1alpha1.XSLTTransform) pkgreconciler.Event {
-	o.Status.InitializeConditions()
-	o.Status.ObservedGeneration = o.Generation
+	var url *apis.URL
+
+	if o.Spec.Sink != nil {
+		uri, err := r.resolveDestination(ctx, o)
```
antoineco, line 52: nit: We could as well assign directly to `url` instead of having this strange `url = uri` below.

pkg/flow/adapter/xslttransform/adapter_test.go
```diff
@@ -256,6 +257,67 @@ func TestXSLTTransformEvents(t *testing.T) {
...
+			ctx := context.Background()
+			style, err := parseXSLT(tc.xslt)
+			assert.NoError(t, err)
+
+			replier, err := targetce.New(env.Component, logtesting.TestLogger(t),
+				targetce.ReplierWithStatefulHeaders(env.BridgeIdentifier),
+				targetce.ReplierWithStaticDataContentType(cloudevents.ApplicationXML),
+				targetce.ReplierWithStaticErrorDataContentType(*cloudevents.StringOfApplicationJSON()),
+				targetce.ReplierWithPayloadPolicy(targetce.PayloadPolicy(targetce.PayloadPolicyAlways)),
+				targetce.ReplierWithStaticResponseType(v1alpha1.EventTypeXSLTTransformError))
```
antoineco, line 297: How many of the options above are really relevant here?
As-is, the test doesn't even assert the response, so I'd keep the replier config to the bear minimum (no option) and focus on the test's main purpose instead: check the shape of the event delivered to K_SINK.

pkg/flow/adapter/xslttransform/adapter_test.go
JeffNeff: sorry! I removed the reference to this but never removed this guy.

pkg/flow/adapter/xslttransform/adapter_test.go
JeffNeff: yea i tried.. but i failed... it is supposed to return this:
```
*event.Event((*event.Event)(nil))
```

But i don't know how to make this to compare it against

JeffNeff:
I need some help with this linter error... there is clearly a newline at the end of the manifest.

```
Error: [new-line-at-end-of-file] no new line character at the end of file
Error: Process completed with exit code 1.
```

pkg/flow/adapter/xslttransform/adapter_test.go
JeffNeff: I removed all the ones except for what I think we need here.

pkg/flow/adapter/xslttransform/adapter_test.go
JeffNeff: Please review my update. I think i got what you are asking for.

pkg/flow/adapter/xslttransform/adapter_test.go
```diff
@@ -256,6 +257,62 @@ func TestXSLTTransformEvents(t *testing.T) {
...
+		xslt           string
+		inEvent        cloudevents.Event
+		expectedEvent  cloudevents.Event
+		expectCategory string
+	}{
+		"transform ok": {
+			xslt:           tXSLT,
+			inEvent:        newCloudEvent(tXML, cloudevents.ApplicationXML),
+			expectedEvent:  newCloudEvent(tOutXML, cloudevents.ApplicationXML),
+			expectCategory: tSuccessAttribute,
```
antoineco, line 271: Not used in the test logic.

pkg/flow/reconciler/xslttransform/adapter.go
```diff
@@ -84,5 +86,11 @@ func makeAppEnv(o *v1alpha1.XSLTTransform) ([]corev1.EnvVar, error) {
 		})
 	}
 
+	if sink != nil {
+		env = append(env, corev1.EnvVar{
+			Name:  envSink,
+			Value: sink.Path,
```
antoineco, line 92: Did you mean `sink.String()`? Otherwise this value is `/`, which is not going to work.

pkg/flow/reconciler/xslttransform/reconciler.go
antoineco: This wasn't resolved.
(Or rather, resolved in a different component unrelated to this PR)

pkg/flow/adapter/xslttransform/adapter_test.go
```diff
@@ -256,6 +257,62 @@ func TestXSLTTransformEvents(t *testing.T) {
...
+				sink:         "http://localhost:8080",
+			}
+
+			_, r := a.Dispatch(ctx, tc.inEvent)
+			assert.Equal(t, cloudevents.ResultACK, r)
+
+			events := ceClient.Sent()
+
+			assert.Equal(t, 1, len(events))
+			assert.Equal(t, tc.expectedEvent, events[0])
```
antoineco, line 311: We have to be careful with the usage of `assert` and `require`.
The first one should be `require`, otherwise the second one panics when `len(events` is 0.

pkg/flow/adapter/xslttransform/adapter_test.go
```diff
@@ -256,6 +257,62 @@ func TestXSLTTransformEvents(t *testing.T) {
...
+
+			ctx := context.Background()
+			style, err := parseXSLT(tc.xslt)
+			assert.NoError(t, err)
+
+			replier, err := targetce.New(env.Component, logtesting.TestLogger(t),
+				targetce.ReplierWithStatefulHeaders(env.BridgeIdentifier),
+				targetce.ReplierWithPayloadPolicy(targetce.PayloadPolicy(targetce.PayloadPolicyAlways)),
+				targetce.ReplierWithStaticResponseType(v1alpha1.EventTypeXSLTTransformError))
+			assert.NoError(t, err)
```
antoineco, line 293: Likewise, regarding `assert` vs `require`.
If you get an error here, we probably should interrupt the test immediately.

pkg/flow/adapter/xslttransform/adapter_test.go
antoineco: I performed the following deletion and the test still runs, so I can conclude with a 100% certainty that nothing of it is required in the context of this particular test. It just distracts from the actual test logic:

```diff
--- a/pkg/flow/adapter/xslttransform/adapter_test.go
+++ b/pkg/flow/adapter/xslttransform/adapter_test.go
@@ -275,30 +275,16 @@ func TestXSLTTransformKSINK(t *testing.T) {
                t.Run(name, func(t *testing.T) {
                        ceClient := adaptertest.NewTestClient()

-                       env := &envAccessor{
-                               EnvConfig: adapter.EnvConfig{
-                                       Component: tCloudEventSource,
-                               },
-                               BridgeIdentifier: tBridgeID,
-                       }
-
                        ctx := context.Background()
                        style, err := parseXSLT(tc.xslt)
                        assert.NoError(t, err)

-                       replier, err := targetce.New(env.Component, logtesting.TestLogger(t),
-                               targetce.ReplierWithStatefulHeaders(env.BridgeIdentifier),
-                               targetce.ReplierWithPayloadPolicy(targetce.PayloadPolicy(targetce.PayloadPolicyAlways)),
-                               targetce.ReplierWithStaticResponseType(v1alpha1.EventTypeXSLTTransformError))
-                       assert.NoError(t, err)
-
                        a := &xsltTransformAdapter{
                                logger: logtesting.TestLogger(t),

                                ceClient:     ceClient,
                                xsltOverride: false,
                                defaultXSLT:  style,
-                               replier:      replier,
                                sink:         "http://localhost:8080",
                        }
```

antoineco (dismissed):
Getting closer and closer but as-is the sink feature doesn't work (`K_SINK` resolves to `"/"`, see comment).

pkg/flow/adapter/xslttransform/adapter.go
```diff
@@ -86,10 +89,10 @@ func NewTarget(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClien
 // or the context is cancelled.
 func (a *xsltTransformAdapter) Start(ctx context.Context) error {
 	a.logger.Info("Starting XSLT transformer")
-	return a.ceClient.StartReceiver(ctx, a.dispatch)
+	return a.ceClient.StartReceiver(ctx, a.Dispatch)
 }
 
-func (a *xsltTransformAdapter) dispatch(ctx context.Context, event cloudevents.Event) (*cloudevents.Event, cloudevents.Result) {
+func (a *xsltTransformAdapter) Dispatch(ctx context.Context, event cloudevents.Event) (*cloudevents.Event, cloudevents.Result) {
```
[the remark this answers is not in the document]
JeffNeff: so that i can access it in the test

antoineco:
> I need some help with this linter error... there is clearly a newline at the end of the manifest.
> 
> ```
> Error: [new-line-at-end-of-file] no new line character at the end of file
> Error: Process completed with exit code 1.
> ```

@JeffNeff your IDE must be misconfigured, there is indeed no newline character at the end of the file.
If I save it in mine (without touching anything else) I see this diff:
image

pkg/flow/adapter/xslttransform/adapter.go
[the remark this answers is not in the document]
antoineco: I have good news for you: tests are in the same package as the production code (they don't have to, but in this case they are), so you don't need to export anything.

pkg/flow/adapter/xslttransform/adapter_test.go
antoineco: I didn't even know `assert` was able to perform this kind of nested `struct` comparisons. Looking good 👍

pkg/flow/adapter/xslttransform/adapter_test.go
antoineco: Marking not resolved.

pkg/flow/adapter/xslttransform/adapter.go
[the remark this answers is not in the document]
JeffNeff: mhm...okay! thank you!

pkg/flow/adapter/xslttransform/adapter.go
```diff
@@ -139,6 +142,23 @@ func (a *xsltTransformAdapter) dispatch(ctx context.Context, event cloudevents.E
...
 
+	if isXML {
+		if err := event.SetData(cloudevents.ApplicationXML, []byte(output)); err != nil {
+			return a.replier.Error(&event, targetce.ErrorCodeAdapterProcess, err, nil)
+		}
+	} else {
+		if err := event.SetData(cloudevents.ApplicationJSON, []byte(output)); err != nil {
+			return a.replier.Error(&event, targetce.ErrorCodeAdapterProcess, err, nil)
+		}
+	}
```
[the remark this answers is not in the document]
JeffNeff: I think its where it should be. no? We have to set the data properly either way (sink/replier). If i nest this under `if a.sink` then i would have to duplicate the code under that if statement in order to set the data for the replier

pkg/flow/adapter/xslttransform/adapter_test.go
JeffNeff: Ive removed all of the replier logic. so this one is no longer there.

pkg/flow/adapter/xslttransform/adapter.go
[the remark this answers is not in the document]
antoineco: @odacremolbap Can you please help here?
I don't understand why we'd need to suddenly SetData when the original version without sink didn't.
Why does the logic need to change when sink **isn't** set.

JeffNeff:
> > I need some help with this linter error... there is clearly a newline at the end of the manifest.
> > ```
> > Error: [new-line-at-end-of-file] no new line character at the end of file
> > Error: Process completed with exit code 1.
> > ```
> 
> @JeffNeff your IDE must be misconfigured, there is indeed no newline character at the end of the file. If I save it in mine (without touching anything else) I see this diff: image

you are right.. i don't know how i didn't have that guy toggled on till now. thank you

pkg/flow/reconciler/xslttransform/adapter.go
JeffNeff: good catch!

pkg/flow/adapter/xslttransform/adapter_test.go
JeffNeff: removed!

pkg/flow/adapter/xslttransform/adapter_test.go
JeffNeff: I removed what you have removed here. thank you

pkg/flow/reconciler/xslttransform/reconciler.go
JeffNeff: got it now!

pkg/flow/adapter/xslttransform/adapter_test.go
JeffNeff: how can i make this? @antoineco

pkg/flow/adapter/xslttransform/adapter.go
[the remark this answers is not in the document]
odacremolbap: I am not sure that I know what this is about.
The output of this transformation is an XML and should be typed as such, unless I am missing something.

Only returned errors should be JSON, and replier already takes care of that.

pkg/flow/adapter/xslttransform/adapter_test.go
```diff
@@ -256,6 +258,44 @@ func TestXSLTTransformEvents(t *testing.T) {
 	}
 }
 
+func TestXSLTTransformKSINK(t *testing.T) {
```
odacremolbap, line 261:
```suggestion
func TestXSLTTransformToSink(t *testing.T) {
```

config/304-xslttransform.yaml
```diff
@@ -89,14 +90,39 @@ spec:
...
             anyOf:
             - required: [xslt]
             - required: [allowPerEventXSLT]
-
+              sink:
+                description: The destination of events transformed by this component.
+                type: object
+                properties:
+                  ref:
+                    description: Reference to an addressable Kubernetes object to be used as the destination of events.
```
odacremolbap, line 104: Can we explain that in the absence of it, the transformed event is sent back to the caller?

odacremolbap (changes requested):
There seems to be some XML/JSON code here.
Is that expected to be at this PR?

pkg/flow/adapter/xslttransform/adapter_test.go
antoineco: `assert.Nil()` should work

config/samples/flows/xslttransform/200-xslttransform.yaml
```diff
@@ -37,3 +37,8 @@ spec:
           </item>
         </xsl:template>
       </xsl:stylesheet>
+  sink:
+    ref:
+      apiVersion: serving.knative.dev/v1
+      kind: Service
+      name: event-display
```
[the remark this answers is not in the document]
JeffNeff: got it!

pkg/flow/adapter/xslttransform/adapter.go
```diff
@@ -57,7 +58,8 @@ func NewTarget(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClien
 		targetce.ReplierWithStatefulHeaders(env.BridgeIdentifier),
 		targetce.ReplierWithStaticDataContentType(cloudevents.ApplicationXML),
 		targetce.ReplierWithStaticErrorDataContentType(*cloudevents.StringOfApplicationJSON()),
-		targetce.ReplierWithPayloadPolicy(targetce.PayloadPolicy(targetce.PayloadPolicyAlways)))
+		targetce.ReplierWithPayloadPolicy(targetce.PayloadPolicy(targetce.PayloadPolicyAlways)),
+		targetce.ReplierWithStaticResponseType(v1alpha1.EventTypeXSLTTransformError))
```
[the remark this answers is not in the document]
JeffNeff: If the event is successful, the incoming event type is used for the outgoing event. IF there is an error you get `io.triggermesh.xslt.transform.error`

config/304-xslttransform.yaml
```diff
@@ -89,14 +90,39 @@ spec:
...
                 - required: [valueFromConfigMap]
-
               allowPerEventXSLT:
                 description: Whether the XSLT informed at the spec can be overriden at each CloudEvent.
                 type: boolean
             anyOf:
             - required: [xslt]
             - required: [allowPerEventXSLT]
-
+              sink:
```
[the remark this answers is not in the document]
JeffNeff: idk... this is where i put it for the xmltojson, so i replicated that here.

pkg/flow/adapter/xslttransform/adapter.go
```diff
@@ -142,14 +142,8 @@ func (a *xsltTransformAdapter) dispatch(ctx context.Context, event cloudevents.E
...
 
-	if isXML {
-		if err := event.SetData(cloudevents.ApplicationXML, []byte(output)); err != nil {
-			return a.replier.Error(&event, targetce.ErrorCodeAdapterProcess, err, nil)
-		}
-	} else {
-		if err := event.SetData(cloudevents.ApplicationJSON, []byte(output)); err != nil {
-			return a.replier.Error(&event, targetce.ErrorCodeAdapterProcess, err, nil)
-		}
+	if err := event.SetData(cloudevents.ApplicationXML, []byte(output)); err != nil {
```
antoineco, line 145: I really still don't get how this is relevant _outside_ of `if sink` below, because now this affects _both_ modes and I don't think we want that.

Does it mean the transformer didn't work at all before the sink feature? Hard for me to believe.

Again, asking for @odacremolbap's help to clarify because this part of the code smells to me.

pkg/flow/adapter/xslttransform/adapter.go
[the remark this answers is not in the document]
odacremolbap: I would suggest to create tests to make sure that code is behaving as you say.
Also, a component replying the same type that it ingests would lead to loops when replying to sender, I don't think replying the same type should be the default behavior.

pkg/flow/adapter/xslttransform/adapter.go
JeffNeff: you are right!

pkg/flow/adapter/xslttransform/adapter.go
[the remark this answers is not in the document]
JeffNeff: you got it! I moved your example into a `replier` folder and created a new one in a `sink` folder. are you okay with this?

pkg/flow/adapter/xslttransform/adapter_test.go
JeffNeff: got it! that worked.

pkg/flow/adapter/xslttransform/adapter.go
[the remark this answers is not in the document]
JeffNeff: I have removed all of this. It was un-needed. I thought that there were cases it responded with JSON. I Was wrong

config/304-xslttransform.yaml
```diff
@@ -98,7 +98,8 @@ spec:
                 type: object
                 properties:
                   ref:
-                    description: Reference to an addressable Kubernetes object to be used as the destination of events. If left empty, the events will be sent back to the sender/source. 
+                    description: -|
```
antoineco, line 101: Tip: You simply need a carriage return before the 120 characters mark to keep the linter happy. No need for `: |` or any other alternative syntax.

JeffNeff: oh nice! thank you!

config/304-xslttransform.yaml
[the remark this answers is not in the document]
JeffNeff: fixed!

pkg/flow/adapter/xslttransform/adapter.go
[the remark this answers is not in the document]
odacremolbap: That line is setting a response type `v1alpha1.EventTypeXSLTTransformError` for all resquests, which I don't think is what you want to do.

This is how the component works:

- responses are typed with the incoming type plus the `.response` suffix
- on success they are added the attribute `category: success`.
- on error they are added the attribute `category: error`.

JeffNeff:
after the conversation in #Knative I have removed the static response type. @odacremolbap

odacremolbap (approved):
Once you fix the failing test this would be ok for me.

pkg/flow/adapter/xslttransform/adapter_test.go
```diff
@@ -256,6 +258,45 @@ func TestXSLTTransformEvents(t *testing.T) {
...
+				sink:         "http://localhost:8080",
+			}
+
+			e, r := a.dispatch(ctx, tc.inEvent)
+			assert.Nil(t, e)
+			assert.Equal(t, cloudevents.ResultACK, r)
+
+			events := ceClient.Sent()
+			require.Equal(t, 1, len(events))
+			assert.Equal(t, tc.expectedEvent, events[0])
```
antoineco, line 295: The test is failing here.

Both the data and context attributes seem to match, but I'm not positive you can just compare deep structs using the `assert` module, due to nested pointers, private fields, etc.

You're going to have to compare attributes and data separately. Example (without data comparison, but you can compare data the same way):
https://github.com/triggermesh/triggermesh/blob/9376615c7461c715eb7578d40ad1814324e8e7e5/pkg/sources/adapter/awssqssource/adapter_test.go#L163-L169

JeffNeff: cant i just do this? https://github.com/triggermesh/triggermesh/pull/503/commits/071681fcf7980174545706149f93df0eb3c5ea6d

antoineco (approved):

PR #507, opened 2022-01-27
Ensure all E2E namespaces have been finalized before completing CI job

### LINKED ISSUE #459: Diagnostic Settings aren't cleaned up at the end of Azure Activity Logs source E2E tests
E2E tests for the Azure Activity Logs source are currently failing consistently with one of the following messages:

```
Error waiting for resource azureactivitylogssources.sources.triggermesh.io "test-nnvsg" to become ready: timed out waiting for the condition
  In [BeforeEach] at: /home/runner/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.0.0/core_dsl.go:492
```

```
  Unable to create resource group: {"error":{"code":"SubscriptionNotFound","message":"The subscription '00000000-0000-0000-0000-000000000000' could not be found."}}
  In [BeforeEach] at: /home/runner/work/triggermesh/triggermesh/test/e2e/sources/azureactivitylogs/main.go:113
```

antoineco:
Fixes #459 (and similar issues, not limited to Azure Activity Logs).

Prevents leak of external resources when KinD cluster gets shut down before API objects have been finalized.

We could have also baked that feature into the E2E framework, but the issue is unlikely to occur outside of a CI pipeline (humans don't typically brutally delete clusters the second tests finish).

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -118,3 +118,27 @@ jobs:
         AZURE_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
         AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
       run: ginkgo -p -slow-spec-threshold=10m -randomize-all ./test/e2e/
+
+    # The KinD cluster can get shut down before all API objects created during
+    # E2E tests have been terminated, which leaks cloud resources depending on
+    # what test(s) are run last by Ginkgo.
+    # To prevent that, we delay the completion of the job until all namespaces
+    # labeled "e2e-framework" have been finalized.
+    - name: Wait for termination of E2E namespaces
+      if: always()
+      run: |
+        declare -i e2e_ns_count=-1
+
+        # retry for max 120s (60*2s)
+        for _ in $(seq 1 60); do
+            e2e_ns_count="$(kubectl get ns -l e2e-framework -o jsonpath='{.items}' | jq '. | length')"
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-tests.yaml
```diff
@@ -118,3 +118,27 @@ jobs:
...
+        declare -i e2e_ns_count=-1
+
+        # retry for max 120s (60*2s)
+        for _ in $(seq 1 60); do
+            e2e_ns_count="$(kubectl get ns -l e2e-framework -o jsonpath='{.items}' | jq '. | length')"
+            if ! ((e2e_ns_count)); then
+                break
+            fi
+
+            echo -n '.' >&2
```
antoineco, line 139: I added this `echo` so that we get some feedback on how long we've been waiting, otherwise it seems like we're just hanging while waiting:

```console
$ ./wait-for-ns-delete.sh
..........
(returned after 20 s)
```

.github/workflows/e2e-tests.yaml
```diff
@@ -118,3 +118,27 @@ jobs:
...
+    - name: Wait for termination of E2E namespaces
+      if: always()
+      run: |
+        declare -i e2e_ns_count=-1
+
+        # retry for max 120s (60*2s)
+        for _ in $(seq 1 60); do
+            e2e_ns_count="$(kubectl get ns -l e2e-framework -o jsonpath='{.items}' | jq '. | length')"
+            if ! ((e2e_ns_count)); then
+                break
```
antoineco, line 136: This condition is hit when the counter hits `0` (no more namespace).

PR #513, opened 2022-01-28
Add startup probe and improve logging in Azure Event Hubs source adapter

### LINKED ISSUE #512: E2E test for Azure Event Hubs source is flaky, fails to receive event
I suspect we send our event too early, immediately after the adapter has started but before the connection with the Hub instance has been established.

Two things could help here:
- a startup probe
- #69

antoineco:
Fixes #512

1. Adds a startup probe that only gets active if all message receivers have been successfully started
2. Improves the logging around startup, shut down, and failures.

### DIFF
--- a/pkg/sources/adapter/azureeventhubsource/adapter.go
+++ b/pkg/sources/adapter/azureeventhubsource/adapter.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 TriggerMesh Inc.
+Copyright 2022 TriggerMesh Inc.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -37,10 +37,13 @@ import (
 	"github.com/triggermesh/triggermesh/pkg/apis/sources"
 	"github.com/triggermesh/triggermesh/pkg/apis/sources/v1alpha1"
 	"github.com/triggermesh/triggermesh/pkg/sources/adapter/azureeventhubsource/trace"
+	"github.com/triggermesh/triggermesh/pkg/sources/adapter/common/health"
 )
 
-// Event Hub connection connTimeout
-const connTimeout = 20 * time.Second
+const (
+	connTimeout  = 20 * time.Second
+	drainTimeout = 1 * time.Minute
+)
 
 // envConfig is a set parameters sourced from the environment for the source's
 // adapter.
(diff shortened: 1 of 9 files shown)


pkg/sources/adapter/azureeventhubsource/adapter.go
```diff
@@ -37,10 +37,13 @@ import (
 	"github.com/triggermesh/triggermesh/pkg/apis/sources"
```
JeffNeff, line 37: Should we also update the date on this file?

pkg/sources/adapter/azureeventhubsource/adapter.go
```diff
@@ -146,28 +149,50 @@ func NewAdapter(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClie
...
 	connCtx, cancel := context.WithTimeout(ctx, connTimeout)
 	runtimeInfo, err := a.ehClient.GetRuntimeInformation(connCtx)
 	cancel()
 	if err != nil {
-		a.logger.Fatal("GetRuntimeInformation failed:", err)
+		return fmt.Errorf("getting Event Hub runtime information: %w", err)
 	}
 	a.runtimeInfo = runtimeInfo
 
+	a.logger.Info("Starting Event Hub message receivers for partitions ", runtimeInfo.PartitionIDs)
```
JeffNeff, line 162: (Question not a review) should this be `Infof()`?

pkg/sources/adapter/azureeventhubsource/adapter.go
```diff
@@ -146,28 +149,50 @@ func NewAdapter(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClie
...
+
 	// listen to each partition of the Event Hub
 	for _, partitionID := range runtimeInfo.PartitionIDs {
 		connCtx, cancel := context.WithTimeout(ctx, connTimeout)
 		_, err := a.ehClient.Receive(connCtx, partitionID, a.handleMessage, eventhub.ReceiveWithLatestOffset())
 		cancel()
 		if err != nil {
-			a.logger.Error("failed to start Event Hub listener:", err)
-			continue
+			a.logger.Errorw("An error occurred while starting message receivers. "+
```
JeffNeff, line 170: isnt it weird to put the wording `An error` in an error message? isn't that inferred?

JeffNeff, line 170:
```suggestion
			a.logger.Errorw("starting message receivers. "+
```

JeffNeff:
nit: I noticed the licence dates for the modified files were not updated.

JeffNeff (approved):
approving with nit comments :)

pkg/sources/adapter/azureeventhubsource/adapter.go
antoineco: Infof would assume there are formatting markers in the string (`%s`, etc.), but there are none here.

pkg/sources/adapter/azureeventhubsource/adapter.go
antoineco: That review would be more useful with a suggestion to reword the message.

antoineco: That's not a returned error, we're **logging** the error directly here. That message won't be wrapped into another error such as "Error running the adapter: ...", it's the entire message.
Another error is actually _returned_ below.

antoineco:
@JeffNeff updating the dates in modified files is our own invention, none of OSS projects that I know do that.

pkg/sources/adapter/azureeventhubsource/adapter.go
antoineco: Based on your other suggestion I think I understand what you mean now.

It's true that it is redundant to say it's an error when returning an error type.
However, this is not what we're doing here. We are logging that something happened in a particular partition, and the error that's actually returned below is more general about the fact that the adapter failed to start for X reason.

We are communicating in real time here something like this:
"oops something bad happened, we are now going to gracefully shut down and exit".
And because the shut down could take a while, we want the user to be notified as early as the problem occured.
Like I said, we are _also_ returning the error afterwards, but with a delay due to the shut down.

PR #514, opened 2022-01-28
Update Google Cloud Storage Target

FranBarrera:
Update Google Cloud Storage Target:

- Remove the possibility to select the file
- By default store the event and not only the event data
- Have a special type to save only the event data
- Add support for DiscardCeContext

### DIFF
--- a/pkg/targets/adapter/googlecloudstoragetarget/adapter.go
+++ b/pkg/targets/adapter/googlecloudstoragetarget/adapter.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 TriggerMesh Inc.
+Copyright 2022 TriggerMesh Inc.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -18,6 +18,7 @@ package googlecloudstoragetarget
 
 import (
 	"context"
+	"encoding/json"
 
 	"cloud.google.com/go/storage"
 	"google.golang.org/api/option"
@@ -55,9 +56,10 @@ func NewTarget(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClien
 		client: client,
 		bucket: client.Bucket(env.BucketName),
 
-		replier:  replier,
-		ceClient: ceClient,
-		logger:   logger,
+		discardCEContext: env.DiscardCEContext,
+		replier:          replier,
+		ceClient:         ceClient,
+		logger:           logger,
 	}
 }
 
@@ -67,9 +69,10 @@ type googlecloudstorageAdapter struct {
 	client *storage.Client
 	bucket *storage.BucketHandle
 
-	replier  *targetce.Replier
-	ceClient cloudevents.Client
-	logger   *zap.SugaredLogger
+	discardCEContext bool
+	replier          *targetce.Replier
+	ceClient         cloudevents.Client
+	logger           *zap.SugaredLogger
 }
 
 // Returns if stopCh is closed or Send() returns an error.
@@ -79,15 +82,14 @@ func (a *googlecloudstorageAdapter) Start(ctx context.Context) error {
 }
 
 func (a *googlecloudstorageAdapter) dispatch(ctx context.Context, event cloudevents.Event) (*cloudevents.Event, cloudevents.Result) {
-	switch typ := event.Type(); typ {
-	case v1alpha1.EventTypeGoogleCloudStorageObjectInsert:
+	if event.Type() == v1alpha1.EventTypeGoogleCloudStorageObjectInsert || a.discardCEContext {
+		return a.insertObjectData(ctx, event)
+	} else {
 		return a.insertObject(ctx, event)
-	default:
-		return a.instertArbitraryObject(ctx, event)
 	}
 }
 
-func (a *googlecloudstorageAdapter) instertArbitraryObject(ctx context.Context, event cloudevents.Event) (*cloudevents.Event, cloudevents.Result) {
+func (a *googlecloudstorageAdapter) insertObjectData(ctx context.Context, event cloudevents.Event) (*cloudevents.Event, cloudevents.Result) {
 
 	obj := a.bucket.Object(event.ID() + ".json")
 	w := obj.NewWriter(ctx)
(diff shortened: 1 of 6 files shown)


pkg/targets/adapter/googlecloudstoragetarget/adapter.go
```diff
@@ -79,15 +82,14 @@ func (a *googlecloudstorageAdapter) Start(ctx context.Context) error {
 }
 
 func (a *googlecloudstorageAdapter) dispatch(ctx context.Context, event cloudevents.Event) (*cloudevents.Event, cloudevents.Result) {
-	switch typ := event.Type(); typ {
-	case v1alpha1.EventTypeGoogleCloudStorageObjectInsert:
+	if event.Type() == v1alpha1.EventTypeGoogleCloudStorageObjectInsert || a.discardCEContext {
+		return a.insertObjectData(ctx, event)
```
antoineco, line 86: We probably don't need two separate functions, do we?
We could pass just the desired data as []byte and I'd assume the rest of the logic to be the same in both cases.

pkg/targets/adapter/googlecloudstoragetarget/adapter.go
```diff
@@ -79,15 +82,14 @@ func (a *googlecloudstorageAdapter) Start(ctx context.Context) error {
...
+	} else {
 		return a.insertObject(ctx, event)
-	default:
-		return a.instertArbitraryObject(ctx, event)
 	}
 }
 
-func (a *googlecloudstorageAdapter) instertArbitraryObject(ctx context.Context, event cloudevents.Event) (*cloudevents.Event, cloudevents.Result) {
+func (a *googlecloudstorageAdapter) insertObjectData(ctx context.Context, event cloudevents.Event) (*cloudevents.Event, cloudevents.Result) {
```
JeffNeff, line 93: should not be a return here.

pkg/targets/adapter/googlecloudstoragetarget/adapter.go
```diff
@@ -104,15 +106,15 @@ func (a *googlecloudstorageAdapter) instertArbitraryObject(ctx context.Context,
 }
 
 func (a *googlecloudstorageAdapter) insertObject(ctx context.Context, event cloudevents.Event) (*cloudevents.Event, cloudevents.Result) {
-	ep := &EventPayload{}
-	if err := event.DataAs(ep); err != nil {
-		return a.replier.Error(&event, targetce.ErrorCodeRequestParsing, err, nil)
+	d, err := json.Marshal(event)
+	if err != nil {
+		return a.replier.Error(&event, "error marshalling CloudEvent", err, nil)
```
JeffNeff, line 111:
```suggestion
		return a.replier.Error(&event, "error marshalling CloudEvent", err, "marshalling the data")
```

pkg/targets/adapter/googlecloudstoragetarget/adapter.go
```diff
@@ -104,15 +106,15 @@ func (a *googlecloudstorageAdapter) instertArbitraryObject(ctx context.Context,
...
 	}
 
-	obj := a.bucket.Object(ep.FileName)
+	obj := a.bucket.Object(event.ID() + ".json")
 	w := obj.NewWriter(ctx)
 
-	if _, err := w.Write(ep.Data); err != nil {
+	if _, err := w.Write(d); err != nil {
 		return a.replier.Error(&event, targetce.ErrorCodeAdapterProcess, err, nil)
```
JeffNeff, line 118:
```suggestion
		return a.replier.Error(&event, targetce.ErrorCodeAdapterProcess, err, "writing the data")
```

pkg/targets/adapter/googlecloudstoragetarget/adapter.go
```diff
@@ -104,15 +106,15 @@ func (a *googlecloudstorageAdapter) instertArbitraryObject(ctx context.Context,
...
-	ep := &EventPayload{}
-	if err := event.DataAs(ep); err != nil {
-		return a.replier.Error(&event, targetce.ErrorCodeRequestParsing, err, nil)
+	d, err := json.Marshal(event)
+	if err != nil {
+		return a.replier.Error(&event, "error marshalling CloudEvent", err, nil)
 	}
 
-	obj := a.bucket.Object(ep.FileName)
+	obj := a.bucket.Object(event.ID() + ".json")
```
JeffNeff, line 114: Why are we taking out the option to specify the filename?

JeffNeff:
I dont understand some of the changes.. the original intention of the two function definitions was to allow the user to pass either specific event type, "com.google.cloud.storage.object.insert", containing a filename and the data. Or arbitrary data.

pkg/targets/adapter/googlecloudstoragetarget/adapter.go
JeffNeff: the reason we had two functions was to allow either arbitrary events to be passed in. or to allow the user to specify the filename and data explicitly via the "com.google.cloud.storage.object.insert" event.. I prefer the original implementation..

JeffNeff:
I made some suggested adapter changes here -> https://github.com/triggermesh/triggermesh/blob/35e4fd837900d1e37d7064fd4f6afebfa5bd9606/pkg/targets/adapter/googlecloudstoragetarget/adapter.go

This target defiantly needed some cleanup. I love the addition of the e2e tests & DiscardCeContext, but I really prefer the original design allowing the user to pass either a crafted event containing a specified filename and data or allowing arbitrary events. I would like to see that still supported moving forward.

antoineco:
The reasons why this was suggested are:

- It is likely to feel very counter-intuitive (and technically involved) for users to have to perform this kind of transformation on the data, and turn it into some kind of REST API call in a CloudEvent shape, just to be able to select the name of the file their data should be saved as.
- Setting the filename based on the CE subject is the most terrible thing we could probably think of. The subject attribute isn't supposed to be unique, it's perfectly legit for multiple events to have the same subject if they relate to the same resource, and this will totally go boom with the current implementation.

There is a very simple way to have the filename feature without compromising on the usability of this target: introduce some kind of convention based on a CE extension that can be applied to any type (e.g. `gcloudstoragefilename: myfile.json`).
A CE extension is easy to stamp on an incoming event. Mangling the entire data? Not so much.

Also, more generally, we've already demonstrated countless times that the "imperative" way of writing targets that require strange transformations doesn't work. Events are mostly for data consumption, and functions + API gateways + connectors are what we need to turning this data into imperative actions.

pkg/targets/adapter/googlecloudstoragetarget/adapter.go
antoineco: We should avoid performing unrelated changes in this PR.

If we want to improve the usage of the replier, let's just open a "Improve the usage of the replier" PR.

cab105 (changes requested):
Thank you for the update. Would agree with @JeffNeff 's take in that the CE-Subject should be used as an alternate way to stream an event as the cloudevent id, when defined as a GUID, does make it harder to identify where the event was written to.

FranBarrera:
I've removed the two functions in favor to have less code that does the same. The only thing that is removed is the ability to select the filename based in the CE subject. I totally agree with the Antoine comments on that. I think we could think in a better way to support it, that could works on multiple targets.

cab105:
Now that I've had some time away and am rereading @antoineco's comment, this is a much larger issue that should be hashed out so that we have some consistency across the other event based storage targets. I'll lift my objection for now so we can get this change in, but we'll definitely need to address this.

@antoineco you do bring up a good point, and boils down to the one question that I had when using `Ce-Subject` as the one optional attribute that could be utilized for defining a filename: what real world problem are we trying to accomplish by storing the event? I'll note that my use of  `Ce-Subject` was to allow for multiple events to share the same subject whether in an overwrite or append capacity.

Perhaps an extension is the right place for it, or perhaps it is something that can be addressed in the spec as a pattern to expect to place incoming events.

I've opened: #540 to track the discussion.

cab105 (approved):

PR #516, opened 2022-01-29
Introduce Azure EventHubs Target e2e test

cab105:
Introduce an E2E test designed specifically for the Azure EventHubs Target.

### DIFF
--- a/test/e2e/targets/azureeventhubs/main.go
+++ b/test/e2e/targets/azureeventhubs/main.go
@@ -1,0 +1,278 @@
+/*
+Copyright 2022 TriggerMesh Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package azureeventhubs
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+	"os"
+	"time"
+
+	"github.com/Azure/azure-event-hubs-go/v3/persist"
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	. "github.com/onsi/ginkgo/v2" //nolint:stylecheck
+	. "github.com/onsi/gomega"    //nolint:stylecheck
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	eventhubs "github.com/Azure/azure-event-hubs-go/v3"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+
+	"github.com/triggermesh/triggermesh/test/e2e/framework"
+	"github.com/triggermesh/triggermesh/test/e2e/framework/apps"
+	"github.com/triggermesh/triggermesh/test/e2e/framework/azure"
+	e2ece "github.com/triggermesh/triggermesh/test/e2e/framework/cloudevents"
+	"github.com/triggermesh/triggermesh/test/e2e/framework/ducktypes"
+)
+
+var targetAPIVersion = schema.GroupVersion{
+	Group:   "targets.triggermesh.io",
+	Version: "v1alpha1",
+}
+
+const (
+	targetKind = "AzureEventHubsTarget"
+)
+
+/*
+  This test requires:
+  - Azure Service Principal Credentials with the Azure Event Hubs Data Owner role assigned at the subscription level
+  The following environment variables _MUST_ be set:
+  - AZURE_SUBSCRIPTION_ID - Common subscription for the test to run against
+  - AZURE_TENANT_ID - Azure tenant to create the resources against
+  - AZURE_CLIENT_ID - The Azure ServicePrincipal Client ID
+  - AZURE_CLIENT_SECRET - The Azure ServicePrincipal Client Secret
+  - AZURE_REGION - Define the Azure region to run the test (default westus2)
+
+  These will be done by the e2e test:
+  - Create an Azure Resource Group, EventHubs Namespace, and EventHub
+  - Send an event to the Azure EventHub via the AzureEventHub Target and verify the event was sent
+*/
+
+var _ = FDescribe("Azure EventHubs target", func() {
+	ctx := context.Background()
+	subscriptionID := os.Getenv("AZURE_SUBSCRIPTION_ID")
+	region := os.Getenv("AZURE_REGION")
+
+	if region == "" {
+		region = "westus2"
+	}
+
+	f := framework.New("azureeventhubstarget")
+
+	var ns string
+	var tgtURL *url.URL
+	var tgtClient dynamic.ResourceInterface
+
+	var rg armresources.ResourceGroup
+	var hub *eventhubs.Hub
+
+	BeforeEach(func() {
+		ns = f.UniqueName
+
+		gvr := targetAPIVersion.WithResource("azureeventhubstargets")
+		tgtClient = f.DynamicClient.Resource(gvr).Namespace(ns)
+	})
+
+	Context("a target is deployed", func() {
+		BeforeEach(func() {
+			By("creating an azure resource group", func() {
+				rg = azure.CreateResourceGroup(ctx, subscriptionID, ns, region)
+				DeferCleanup(func() {
+					By("deleting the azure resource group", func() {
+						_ = azure.DeleteResourceGroup(ctx, subscriptionID, *rg.Name)
+					})
+				})
+			})
+
+			By("creating an azure eventhub", func() {
+				hub = azure.CreateEventHubComponents(ctx, subscriptionID, ns, region, *rg.Name)
+			})
+		})
+
+		When("the spec contains default settings", func() {
+			var event *cloudevents.Event
+
+			BeforeEach(func() {
+				By("creating an AzureEventHubTarget object", func() {
+					tgt, err := createTarget(tgtClient, ns, "test-",
+						withServicePrincipal(),
+						withEventHubID(subscriptionID, *rg.Name, ns, ns))
+
+					Expect(err).ToNot(HaveOccurred())
+
+					tgt = ducktypes.WaitUntilReady(f.DynamicClient, tgt)
+					tgtURL = ducktypes.Address(tgt)
+					Expect(tgtURL).ToNot(BeNil())
+				})
+			})
+
+			It("receives an event on the Event Bus", func() {
+				var payload []byte
+				var partitionIDs []string
+				var eventHandler eventhubs.Handler
+				eventReceivedChannel := make(chan bool)
+				var evCtx context.Context       // Used to set a timeout for reading events
+				var evCancel context.CancelFunc // Used to cancel the reading context
+
+				By("retrieving eventhub partition details", func() {
+					info, err := hub.GetRuntimeInformation(ctx)
+					Expect(err).NotTo(HaveOccurred())
+
+					partitionIDs = info.PartitionIDs
+					Expect(len(partitionIDs) > 0).To(BeTrue())
+				})
+
+				By("creating a handler to verify the received event", func() {
+					eventHandler = func(ctx context.Context, ev *eventhubs.Event) error {
+						defer GinkgoRecover() // To circumvent being called from inside a goroutine
+
+						payload = ev.Data
+						Expect(len(payload) > 0).To(BeTrue())
+
+						// NOTE: The payload will be a stringified version of the cloudevent
+						Expect(payload).To(ContainSubstring(string(event.Data())))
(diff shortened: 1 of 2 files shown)


test/e2e/targets/azureeventhubs/main.go
```diff
@@ -0,0 +1,278 @@
...
+	BeforeEach(func() {
+		ns = f.UniqueName
+
+		gvr := targetAPIVersion.WithResource("azureeventhubstargets")
+		tgtClient = f.DynamicClient.Resource(gvr).Namespace(ns)
+	})
+
+	Context("a target is deployed", func() {
+		BeforeEach(func() {
+			By("creating an azure resource group", func() {
```
antoineco, line 96: Azure

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -0,0 +1,278 @@
...
+				rg = azure.CreateResourceGroup(ctx, subscriptionID, ns, region)
+				DeferCleanup(func() {
+					By("deleting the azure resource group", func() {
```
antoineco, line 99: Azure

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -0,0 +1,278 @@
...
+						_ = azure.DeleteResourceGroup(ctx, subscriptionID, *rg.Name)
+					})
+				})
+			})
+
+			By("creating an azure eventhub", func() {
```
antoineco, line 105: Azure Event Hub

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -0,0 +1,278 @@
...
+  - AZURE_CLIENT_ID - The Azure ServicePrincipal Client ID
+  - AZURE_CLIENT_SECRET - The Azure ServicePrincipal Client Secret
+  - AZURE_REGION - Define the Azure region to run the test (default westus2)
+
+  These will be done by the e2e test:
+  - Create an Azure Resource Group, EventHubs Namespace, and EventHub
+  - Send an event to the Azure EventHub via the AzureEventHub Target and verify the event was sent
+*/
+
+var _ = FDescribe("Azure EventHubs target", func() {
```
antoineco, line 69: Event Hubs

And careful with the Focused test, it will cause the suite to exit with a non-zero code when merged.

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -0,0 +1,278 @@
...
+  - Azure Service Principal Credentials with the Azure Event Hubs Data Owner role assigned at the subscription level
+  The following environment variables _MUST_ be set:
+  - AZURE_SUBSCRIPTION_ID - Common subscription for the test to run against
+  - AZURE_TENANT_ID - Azure tenant to create the resources against
+  - AZURE_CLIENT_ID - The Azure ServicePrincipal Client ID
+  - AZURE_CLIENT_SECRET - The Azure ServicePrincipal Client Secret
+  - AZURE_REGION - Define the Azure region to run the test (default westus2)
+
+  These will be done by the e2e test:
+  - Create an Azure Resource Group, EventHubs Namespace, and EventHub
```
antoineco, line 65: Event Hubs, Event Hub

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -0,0 +1,278 @@
...
+  - Send an event to the Azure EventHub via the AzureEventHub Target and verify the event was sent
```
antoineco, line 66: Event Hub

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -0,0 +1,278 @@
...
+						// Pass the bool to the channel to terminate the receiver
+						eventReceivedChannel <- true
+
+						return nil
+					}
+				})
+
+				By("invoking the handler", func() {
+					// Set a context with a timeout to ensure the event handler isn't waiting forever
+					evCtx, evCancel = context.WithDeadline(ctx, time.Now().Add(time.Second*15))
```
antoineco, line 165: context.Timeout() is what you're looking for.

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -0,0 +1,278 @@
...
+
+				By("retrieving eventhub partition details", func() {
+					info, err := hub.GetRuntimeInformation(ctx)
+					Expect(err).NotTo(HaveOccurred())
+
+					partitionIDs = info.PartitionIDs
+					Expect(len(partitionIDs) > 0).To(BeTrue())
+				})
+
+				By("creating a handler to verify the received event", func() {
```
antoineco, line 143: This test reads backwards. We read about the assertions before even understanding how the data got it.
In production code things like that are acceptable, but in tests we should really emphasize on clarity.

I'd strongly prefer to see these assertions in the last By(), where the event is received from the channel, and only limit the logic of the handler to sending to that channel.

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -0,0 +1,278 @@
...
+
+					Expect(err).ToNot(HaveOccurred())
+
+					tgt = ducktypes.WaitUntilReady(f.DynamicClient, tgt)
+					tgtURL = ducktypes.Address(tgt)
+					Expect(tgtURL).ToNot(BeNil())
+				})
+			})
+
+			It("receives an event on the Event Bus", func() {
```
antoineco, line 127: Hub

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -0,0 +1,278 @@
...
+			By("creating an azure eventhub", func() {
+				hub = azure.CreateEventHubComponents(ctx, subscriptionID, ns, region, *rg.Name)
+			})
+		})
+
+		When("the spec contains default settings", func() {
+			var event *cloudevents.Event
+
+			BeforeEach(func() {
+				By("creating an AzureEventHubTarget object", func() {
```
antoineco, line 114: EventHub **s** Target

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -0,0 +1,278 @@
...
+
+			It("receives an event on the Event Bus", func() {
+				var payload []byte
+				var partitionIDs []string
+				var eventHandler eventhubs.Handler
+				eventReceivedChannel := make(chan bool)
+				var evCtx context.Context       // Used to set a timeout for reading events
+				var evCancel context.CancelFunc // Used to cancel the reading context
+
+				By("retrieving eventhub partition details", func() {
```
antoineco, line 135: Event Hub

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -0,0 +1,278 @@
...
+					partitionIDs = info.PartitionIDs
+					Expect(len(partitionIDs) > 0).To(BeTrue())
+				})
+
+				By("creating a handler to verify the received event", func() {
+					eventHandler = func(ctx context.Context, ev *eventhubs.Event) error {
+						defer GinkgoRecover() // To circumvent being called from inside a goroutine
+
+						payload = ev.Data
+						Expect(len(payload) > 0).To(BeTrue())
```
antoineco, line 148: Could that help reading this more naturally?
https://pkg.go.dev/github.com/onsi/gomega#BeNumerically

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -0,0 +1,278 @@
...
+				})
+			})
+
+			It("receives an event on the Event Bus", func() {
+				var payload []byte
+				var partitionIDs []string
+				var eventHandler eventhubs.Handler
+				eventReceivedChannel := make(chan bool)
+				var evCtx context.Context       // Used to set a timeout for reading events
+				var evCancel context.CancelFunc // Used to cancel the reading context
```
antoineco, line 133: Most of the variables above do not need to carry between spec/by blocks, and shouldn't be declared here. They confuse the reader about what needs to be in the Ginkgo scope, and what doesn't 
Don't hesitate to declare these inside By() blocks, this is exactly where they belong.

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -0,0 +1,278 @@
...
+				})
+			})
+
+			By("creating an azure eventhub", func() {
+				hub = azure.CreateEventHubComponents(ctx, subscriptionID, ns, region, *rg.Name)
+			})
+		})
+
+		When("the spec contains default settings", func() {
+			var event *cloudevents.Event
```
antoineco, line 111: Another variable that much higher in the scope than needed.
If it's not initialized in a BeforeEach, it shouldn't be here per Ginkgo's definition of a test scope.

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -0,0 +1,278 @@
...
+				eventReceivedChannel := make(chan bool)
+				var evCtx context.Context       // Used to set a timeout for reading events
+				var evCancel context.CancelFunc // Used to cancel the reading context
+
+				By("retrieving eventhub partition details", func() {
+					info, err := hub.GetRuntimeInformation(ctx)
+					Expect(err).NotTo(HaveOccurred())
+
+					partitionIDs = info.PartitionIDs
+					Expect(len(partitionIDs) > 0).To(BeTrue())
```
antoineco, line 140: Could that help reading this more naturally?
https://pkg.go.dev/github.com/onsi/gomega#BeNumerically

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -0,0 +1,278 @@
...
+				})
+
+				By("creating a handler to verify the received event", func() {
+					eventHandler = func(ctx context.Context, ev *eventhubs.Event) error {
+						defer GinkgoRecover() // To circumvent being called from inside a goroutine
```
antoineco, line 145: Can you elaborate and/or link to the relevant documentation?

antoineco (commented):
Test logic looks great 👍

I have a few comments regarding the scope of the variables, which felt a bit all over the place to me while reading the logic.

I also have a concern with the way assertions are performed "before" the data is sent (in terms of the order in which a typical human will parse the logic).

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -0,0 +1,278 @@
...
+
+						payload = ev.Data
+						Expect(len(payload) > 0).To(BeTrue())
+
+						// NOTE: The payload will be a stringified version of the cloudevent
```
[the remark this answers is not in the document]
cab105: Could be interesting given the `cloudenvents.Event.String()` function serializes the string more into a human readable context instead of something that can be machine parseable. Would rather tackle this in a follow-up PR that corrects how the full event is sent to the Azure EventHub.

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -0,0 +1,278 @@
...
+
+  These will be done by the e2e test:
+  - Create an Azure Resource Group, EventHubs Namespace, and EventHub
+  - Send an event to the Azure EventHub via the AzureEventHub Target and verify the event was sent
+*/
+
+var _ = FDescribe("Azure EventHubs target", func() {
+	ctx := context.Background()
+	subscriptionID := os.Getenv("AZURE_SUBSCRIPTION_ID")
+	region := os.Getenv("AZURE_REGION")
```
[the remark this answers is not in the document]
cab105: You raise a good point, and will address this.

test/e2e/targets/azureeventhubs/main.go
cab105: Oooh good catch. Thought I removed the focused aspect of the test during development.

test/e2e/targets/azureeventhubs/main.go
cab105: In this case order really does matter as we need the handler configured before we can register it, and the handler needs to run before we send the event. I could nest the By to call another function that performs the actual test to fix the ordering if that will help.

test/e2e/targets/azureeventhubs/main.go
cab105: With the exception of `payload`, the rest are used by multiple `By()` blocks which is why they are located where they are.

test/e2e/targets/azureeventhubs/main.go
cab105: Created https://github.com/triggermesh/triggermesh/issues/517 to track this.

test/e2e/targets/azureeventhubs/main.go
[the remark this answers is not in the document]
antoineco: I was more thinking about about using `json.Unmarshal(<pointer to cloudevent.Event>)`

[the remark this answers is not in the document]
antoineco: Please ignore the above comment, I see what you mean now.

test/e2e/targets/azureeventhubs/main.go
antoineco: I really don't see why we couldn't tackle it in this PR, the data format is a different issue.

If
1. the handler writes to a channel and
2. the By() that performs the assertions reads from that same channel
you can have a test that reads naturally without compromising on concurrency

antoineco (approved):
One last comment (sorry to insist) on the order in which the logic is presented to the user.

Other than that I think this PR is good to go.

test/e2e/targets/azureeventhubs/main.go
cab105: Sorry, #517 comment was intended for the event deserialization. Will refactor the test to ensure the event is placed on a queue and then evaluated so things read in a more linear fashion.

PR #518, opened 2022-01-31
IBM MQ source TLS support

tzununbekov:
In this PR IBM MQ source CRD updated to introduce TLS connection security:

- CR's additional `TLS` section added, example:
  ```yaml
  ...
  tls:
    cipher: TLS_RSA_WITH_AES_128_CBC_SHA256
    clientAuthRequired: false
    keyRepository:
      kdbSecret:
        name: ibm-mq-certificate
        key: key.kdb
      sthSecret:
        name: ibm-mq-certificate
        key: key.sth
  ```
  TLS security parameters can be used together with basic authentication.
  More information about IBM MQ TLS is here:
  https://www.ibm.com/docs/en/ibm-mq/9.2?topic=mechanisms-tls-security-protocols-in-mq

- A couple of small fixes and updates for the source adapter, such as CE `source` attribute update and the plain text MQ message support.

Follow-up issues:
- add TLS support in the IBM MQ target,
- make `config/samples` sub-directories structure consistent.


config/300-ibmmqsource.yaml
```diff
@@ -162,9 +162,60 @@ spec:
...
+                            - name
+                            - key
+                        required:
+                        - kdbSecret
+                        - sthSecret
+                    required:
+                    - cipher
+                    - clientAuthRequired
+                    - keyRepository
+                anyOf:
```
odacremolbap, line 215: would this be `oneOf` ?

config/300-ibmmqsource.yaml
```diff
@@ -162,9 +162,60 @@ spec:
...
+                        description: Parameter to specify whether TLS client authentication is required. For more
+                          information, please refer to the IBM MQ documentation at
+                          https://www.ibm.com/docs/en/ibm-mq/9.2?topic=fields-sslclientauth-mqlong.
+                        type: boolean
+                      keyRepository:
+                        description: Key repository for (m)TLS connection. For more information, please refer to the
+                          IBM MQ documentation at https://www.ibm.com/docs/en/ibm-mq/9.2?topic=mq-ssltls-key-repository.
+                        type: object
+                        properties:
+                          kdbSecret:
```
odacremolbap, line 186: opinion, dismiss if you think otherwise:

I am more used to see the `secretKeyRef` name when referencing secrets. Something like

```yaml
keyDatabase:
  secretKeyRef:
     ...
```

pkg/apis/sources/v1alpha1/ibmmq_types.go
```diff
@@ -53,19 +54,34 @@ type IBMMQSourceSpec struct {
...
+// TLSSpec holds the IBM MQ TLS authentication parameters.
+type TLSSpec struct {
+	Cipher             string   `json:"cipher"`
+	ClientAuthRequired bool     `json:"clientAuthRequired"`
+	KeyRepository      Keystore `json:"keyRepository"`
+}
+
+// Keystore represents Key Database components.
+type Keystore struct {
+	KeyDatabase   *corev1.SecretKeySelector `json:"kdbSecret"`
```
antoineco, line 83: This is neither compliant with the way targets nor sources have secrets exposed in specs, and we already have trouble agreeing on a common way across TriggerMesh, so we should at least stick to one of these two:

https://github.com/triggermesh/triggermesh/blob/6dc00197a0ea827840b501d1c9cd16ae536c1c5b/config/samples/targets/confluent/200-target.yaml#L30-L32

https://github.com/triggermesh/triggermesh/blob/6dc00197a0ea827840b501d1c9cd16ae536c1c5b/config/samples/sources/awscognitoidentitysource.yaml#L32-L34

tzununbekov: `valueFromSecret` added in 86af0b1

config/300-ibmmqsource.yaml
tzununbekov: I believe that it should be
```
anyOf:
- required: [username, password]
- required: [tls]
```
meaning either the `username` and `password` must exist, or `tls`, or both, but it does not seem to have any effect.

antoineco: I think `anyOf` is correct, my understanding is that either `tls` or `username, password` or `tls, username, password` can be used.

antoineco (approved):
Again, I'm lacking a lot of context and knowledge about MQ, so stamping without a deep review.

config/300-ibmmqsource.yaml
tzununbekov: Decided to go with the `valueFromSecret` in 86af0b1 to have all secret references in the same style in this CRD.

odacremolbap (approved):

PR #520, opened 2022-01-31
Add handler to the Amazon SNS source for HTTP startup probe

### LINKED ISSUE #29: Duplicate subscriptions for the same source instance
We often notice in our end-to-end tests that more than one subscription are created for a single instance of a AWSSNSSource. Not only this leads to **dangling subscriptions**, but it also causes **multiple message deliveries**.

It appears that attempts to subscribe fail for a little while because the receive adapter isn't yet ready to accept HTTP requests, then, after a few minutes (!), the error `Subscription is pending confirmation, will retry` is reported multiple times, before the subscription eventually succeeds:

```
  Normal   CreateAdapter    3m43s                  awssnssource-controller  Created adapter Service "awssnssource-adapter"
  Warning  FailedSubscribe  3m35s (x3 over 3m36s)  awssnssource-controller  Error subscribing endpoint "http://awssnssource-adapter.e2e-awssnssource-1787.k.triggermesh.io/e2e-awssnssource-1787/test-nsbrz" to SNS topic "arn:aws:sns:eu-central-1:123456789012:e2e-awssnssource-1787": InvalidParameter: Invalid parameter: Unreachable Endpoint
  Warning  FailedSubscribe  14s (x5 over 16s)      awssnssource-controller  Subscription is pending confirmation, will retry
  Normal   Subscribed       13s                    awssnssource-controller  Subscribed to SNS topic "arn:aws:sns:eu-central-1:123456789012:e2e-awssnssource-1787"
```

During these rapid retries, multiple subscriptions are occasionally created:

```console
$ aws sns list-subscriptions | jq -r '.Subscriptions[] | select(.SubscriptionArn | test("e2e-awssnssource")) | .SubscriptionArn'
arn:aws:sns:eu-central-1:123456789012:e2e-awssnssource-1787:9aeab061-932c-49c9-8ef0-9d1746240a1a
arn:aws:sns:eu-central-1:123456789012:e2e-awssnssource-1787:c0ba1d42-b2d3-434b-ac47-d60f05b99016
```

Of course, the source reports only one subscription in its status:

```yaml
subscriptionARN: arn:aws:sns:eu-central-1:123456789012:e2e-awssnssource-1787:c0ba1d42-b2d3-434b-ac47-d60f05b99016
```

antoineco:
Returns an error until a handler has been registered for _all_ source objects cached in the informer.

Prevents the reconciler from attempting to subscribe new instances before the HTTP handler is actually ready to serve.
Avoids multi-registrations as a result.

Fixes #29

---

Note: this is really meant to be used as a **startup** probe, but because Knative doesn't support it, it is set as a readiness probe instead.
This should have the same result though, because Knative doesn't actually set a `readinessProbe` on the user container. Instead, it passes the probe definition to the `queue-proxy` container as an environment variable, which uses it to perform a series of quick requests when the pod starts.

```console
$ kubectl get deployments/awssnssource-adapter-00001-deployment -o jsonpath='{.spec.template.spec.containers[?(@.name=="adapter")].readinessProbe}'
null
```

```console
$ kubectl get deployments/awssnssource-adapter-00001-deployment -o jsonpath='{.spec.template.spec.containers[?(@.name=="queue-proxy")].env[?(@.name=="SERVING_READINESS_PROBE")].value}'
{"httpGet":{"path":"/health","port":8080,"host":"127.0.0.1","scheme":"HTTP","httpHeaders":[{"name":"K-Kubelet-Probe","value":"queue"}]},"successThreshold":1}
```

Then, if I log every request to the internal Router for debugging purposes:

```console
$ kubectl logs deployments/awssnssource-adapter-00001-deployment adapter -f
2022/01/31 14:52:59 GET "/health" map[Accept-Encoding:[gzip] Connection:[close] K-Kubelet-Probe:[queue] User-Agent:[kube-probe//]]
2022/01/31 14:53:00 GET "/health" map[Accept-Encoding:[gzip] Connection:[close] K-Kubelet-Probe:[queue] User-Agent:[kube-probe//]]
2022/01/31 14:53:01 GET "/health" map[Accept-Encoding:[gzip] Connection:[close] K-Kubelet-Probe:[queue] User-Agent:[kube-probe//]]
2022/01/31 14:53:02 GET "/health" map[Accept-Encoding:[gzip] Connection:[close] K-Kubelet-Probe:[queue] User-Agent:[kube-probe//]]
2022/01/31 14:53:03 GET "/health" map[Accept-Encoding:[gzip] Connection:[close] K-Kubelet-Probe:[queue] User-Agent:[kube-probe//]]
2022/01/31 14:53:04 GET "/health" map[Accept-Encoding:[gzip] Connection:[close] K-Kubelet-Probe:[queue] User-Agent:[kube-probe//]]
2022/01/31 14:53:05 GET "/health" map[Accept-Encoding:[gzip] Connection:[close] K-Kubelet-Probe:[queue] User-Agent:[kube-probe//]]
2022/01/31 14:53:06 GET "/health" map[Accept-Encoding:[gzip] Connection:[close] K-Kubelet-Probe:[queue] User-Agent:[kube-probe//]]
2022/01/31 14:53:07 GET "/health" map[Accept-Encoding:[gzip] Connection:[close] K-Kubelet-Probe:[queue] User-Agent:[kube-probe//]]
2022/01/31 14:53:08 GET "/health" map[Accept-Encoding:[gzip] Connection:[close] K-Kubelet-Probe:[queue] User-Agent:[kube-probe//]]
2022/01/31 14:53:09 GET "/health" map[Accept-Encoding:[gzip] Connection:[close] K-Kubelet-Probe:[queue] User-Agent:[kube-probe//]]
...
```

### DIFF
--- a/pkg/sources/adapter/awssnssource/probe/probe.go
+++ b/pkg/sources/adapter/awssnssource/probe/probe.go
@@ -1,0 +1,126 @@
+/*
+Copyright 2022 TriggerMesh Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package probe contains facilities for asserting the readiness of a
+// multi-tenant receive adapter.
+package probe
+
+import (
+	"fmt"
+	"sync"
+
+	"k8s.io/apimachinery/pkg/labels"
+
+	"github.com/triggermesh/triggermesh/pkg/apis/sources/v1alpha1"
+	listersv1alpha1 "github.com/triggermesh/triggermesh/pkg/client/generated/listers/sources/v1alpha1"
+	"github.com/triggermesh/triggermesh/pkg/sources/adapter/common/router"
+)
+
+// ReadinessChecker can assert the readiness of a component.
+type ReadinessChecker interface {
+	IsReady() (bool, error)
+}
+
+// AdapterReadyChecker asserts the readiness of a receive adapter.
+type AdapterReadyChecker struct {
+	srcLister listersv1alpha1.AWSSNSSourceLister
+
+	srcRouter *router.Router
+
+	sync.RWMutex
+	isReady bool
+}
+
+var _ ReadinessChecker = (*AdapterReadyChecker)(nil)
+
+// NewAdapterReadyChecker returns an AdapterReadyChecker initialized with the
+// given Lister and Router.
+func NewAdapterReadyChecker(ls listersv1alpha1.AWSSNSSourceLister, r *router.Router) *AdapterReadyChecker {
+	return &AdapterReadyChecker{
+		srcLister: ls,
+		srcRouter: r,
+	}
+}
+
+// IsReady implements ReadinessChecker.
+// It checks whether the adapter has registered a handler for all observed sources.
+//
+// NOTE(antoineco): we might want to revisit this in the future, because the
+// following implementation details are currently leaking into this
+// ReadinessChecker implementation:
+//
+// - The fact that the adapter exposes a health endpoint via the HTTP router.
+// - The exact URL path of this health endpoint.
+// - The fact that a handler isn't registered for sources which don't have a valid sink.
+//
+// We consider this tight coupling acceptable for the time being, because this
+// is a sub-package of the main adapter package for the SNS source
+// specifically.
+func (c *AdapterReadyChecker) IsReady() (bool, error) {
+	// readiness already observed at an earlier point in time,
+	// short-circuit the check
+	if c.readLockedIsReady() {
+		return true, nil
+	}
+
+	c.Lock()
+	defer c.Unlock()
+
+	// double-checked lock to ensure we don't write the value of "isReady"
+	// twice if multiple goroutines called IsReady() simultaneously.
+	if c.isReady {
+		return true, nil
+	}
+
+	sources, err := c.srcLister.List(labels.Everything())
+	if err != nil {
+		return false, fmt.Errorf("listing cached sources: %w", err)
+	}
+	// informer cache wasn't populated yet
+	if len(sources) == 0 {
+		return false, nil
+	}
+
+	numCachedSrcsWithSink := len(filterSourcesWithSink(sources))
+	numSrcHandlers := c.srcRouter.HandlersCount(ignoreHealthEndpoint)
+
+	c.isReady = numSrcHandlers >= numCachedSrcsWithSink
+
+	return c.isReady, nil
+}
+
+// readLockedIsReady returns the value of c.isReady while guaranteeing that
+// this value isn't currently being written by another goroutine.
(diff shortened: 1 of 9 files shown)


pkg/sources/adapter/awssnssource/probe/probe.go
```diff
@@ -0,0 +1,126 @@
...
+	numCachedSrcsWithSink := len(filterSourcesWithSink(sources))
+	numSrcHandlers := c.srcRouter.HandlersCount(ignoreHealthEndpoint)
+
+	c.isReady = numSrcHandlers >= numCachedSrcsWithSink
+
+	return c.isReady, nil
+}
+
+// readLockedIsReady returns the value of c.isReady while guaranteeing that
+// this value isn't currently being written by another routing.
```
antoineco, line 106:
```suggestion
// this value isn't currently being written by another goroutine.
```

pkg/sources/adapter/awssnssource/probe/probe.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	c.Lock()
+	defer c.Unlock()
+
+	// double-checked lock to ensure we don't write the value of "isReady"
+	// twice if multiple goroutines called IsReady() simultaneously.
+	if c.isReady {
+		return true, nil
+	}
```
antoineco, line 86: Just in case that part isn't clear, the double check on `c.isReady` guarantees that we avoid continuing unnecessarily in case the following scenario occurs:

```golang
/* Two concurrent calls to IsReady() cause two goroutines to land here simultaneously: */

	// <-------- goroutine1
	// <-------- goroutine2
	c.Lock()
	defer c.Unlock()
```

```golang
/* The mutex lets *one* routine through while the other waits until we call Unlock(): */

	// <-------- goroutine2
	c.Lock()
	defer c.Unlock()
	// <-------- goroutine1
```

```golang
/* The function called by goroutine1 returned, isReady is now true, but goroutine2 is still
    inside IsReady() and gets unblocked: */

	c.Lock()
	defer c.Unlock()

	// <-------- goroutine2

	// we save goroutine2 the effort of doing again what goroutine1 just did
	if c.isReady {
		return true, nil
	}
```

pkg/sources/adapter/awssnssource/probe/probe.go
```diff
@@ -0,0 +1,126 @@
...
+// following implementation details are currently leaking into this
+// ReadinessChecker implementation:
+//
+// - The fact that the adapter exposes a health endpoint via the HTTP router.
+// - The exact URL path of this health endpoint.
+// - The fact that a handler isn't registered for sources which don't have a valid sink.
+//
+// We consider this tight coupling acceptable for the time being, because this
+// is a sub-package of the main adapter package for the SNS source
+// specifically.
```
antoineco, line 71: After sending the PR, I realized that the coupling between the adapter and the checker are so tight that it might make no sense at all to have the checker inside a separate package.

Comments/concerns welcome.

pkg/sources/reconciler/common/resource/container.go
```diff
@@ -173,19 +173,27 @@ func Probe(path, port string) ObjectOption {
...
+		case *appsv1.Deployment, *servingv1.Service:
 			sp = &firstContainer(o).StartupProbe
 		}
 
+		switch object.(type) {
+		case *corev1.Container, *appsv1.Deployment:
+			intstrPort = intstr.FromString(port)
+		case *servingv1.Service:
+			// setting port explicitly is invalid in a Knative Service
+		}
```
antoineco, line 190: Although Knative doesn't yet support startup probes, I know there are plans to make it happen in the near future so I decided to include support for it in this PR.

I'm fine deleting that code if this feels unrelated to the main purpose of the PR.

PR #524, opened 2022-01-31
Enable support for discardCEContext and send CloudEvent data as a JSON payload for the Azure Event Hubs target

### LINKED ISSUE #517: Azure Event Hubs payload needs to be machine readable
While working on the E2E test for #516 I observed two issues that will need to be resolved to fully utilize the target:
1. The current event payload that is streamed into the Event Hub is a human readable stringification of the event instead of a machine readable JSON version of the event. To be usable by future consumers, and standardize on the other streaming targets such as AWS, this needs to emit the JSON representation.
2. The adapter logic appears to handle `discardCEContext`, but there is no mention of it in the target yaml configuration.

cab105:
While working on the E2E test for the Azure Event Hubs target, noticed that the option to discard CloudEventwas not included with the config, and therefore the target defaulted to sending the full CloudEvent payload. Additionally, the CloudEvent payload was being sent as a human readable string instead of a marshalled JSON payload that is machine readable.

Closes #517

### DIFF
--- a/config/301-azureeventhubstarget.yaml
+++ b/config/301-azureeventhubstarget.yaml
@@ -66,7 +66,7 @@ spec:
                 type: string
                 pattern: ^\/subscriptions\/[a-z0-9-]+\/resourceGroups\/[\w.()-]+\/providers\/Microsoft.EventHub\/namespaces\/[A-Za-z0-9-]{6,50}\/event[Hh]ubs\/[a-zA-Z0-9][\w.-]{0,49}$
               discardCloudEventContext:
-                description: Whether to omit CloudEvent context attributes in objects created in S3.
+                description: Whether to omit CloudEvent context attributes in objects created in Azure Event Hub.
                   When this property is false (default), the entire CloudEvent payload is included.
                   When this property is true, only the CloudEvent data is included.
                 type: boolean
--- a/config/301-azureeventhubstarget.yaml
+++ b/config/301-azureeventhubstarget.yaml
@@ -65,6 +65,11 @@ spec:
                     /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.EventHub/namespaces/{namespaceName}/eventhubs/{eventHubName}
                 type: string
                 pattern: ^\/subscriptions\/[a-z0-9-]+\/resourceGroups\/[\w.()-]+\/providers\/Microsoft.EventHub\/namespaces\/[A-Za-z0-9-]{6,50}\/event[Hh]ubs\/[a-zA-Z0-9][\w.-]{0,49}$
+              discardCloudEventContext:
+                description: Whether to omit CloudEvent context attributes in objects created in S3.
+                  When this property is false (default), the entire CloudEvent payload is included.
+                  When this property is true, only the CloudEvent data is included.
+                type: boolean
               auth:
                 description: Authentication method to interact with the Azure Event Hubs REST API.
                 type: object
--- a/pkg/apis/targets/v1alpha1/azureeventhubs_types.go
+++ b/pkg/apis/targets/v1alpha1/azureeventhubs_types.go
@@ -17,9 +17,10 @@ limitations under the License.
 package v1alpha1
 
 import (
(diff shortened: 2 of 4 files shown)


config/301-azureeventhubstarget.yaml
```diff
@@ -65,6 +65,11 @@ spec:
                     /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.EventHub/namespaces/{namespaceName}/eventhubs/{eventHubName}
                 type: string
                 pattern: ^\/subscriptions\/[a-z0-9-]+\/resourceGroups\/[\w.()-]+\/providers\/Microsoft.EventHub\/namespaces\/[A-Za-z0-9-]{6,50}\/event[Hh]ubs\/[a-zA-Z0-9][\w.-]{0,49}$
+              discardCloudEventContext:
```
antoineco, line 68: Not S3 :)

pkg/apis/targets/v1alpha1/azureeventhubs_types.go
```diff
@@ -61,7 +62,7 @@ type AzureEventHubsTargetSpec struct {
 	// EventOptions for targets
 	EventOptions *EventOptions `json:"eventOptions,omitempty"`
 
-	DiscardCEContext bool `json:"discardCEContext"`
+	DiscardCEContext bool `json:"discardCloudEventContext"`
```
antoineco, line 65: Nice catch, thanks for fixing that one.

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -193,9 +292,9 @@ var _ = Describe("Azure Event Hubs target", func() {
 
 					// NOTE: The payload will be a stringified version of the CloudEvent
```
antoineco, line 293: This shouldn't be true anymore, right? The message should be serializable to `cloudevent.Event` based on the change proposed in this PR.

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -193,9 +292,9 @@ var _ = Describe("Azure Event Hubs target", func() {
...
 					Expect(payload).To(ContainSubstring(string(event.Data())))
-					Expect(payload).To(ContainSubstring("type: " + event.Type()))
-					Expect(payload).To(ContainSubstring("source: " + event.Source()))
-					Expect(payload).To(ContainSubstring("id: " + event.ID()))
+					Expect(payload).ToNot(ContainSubstring("type: " + event.Type()))
```
antoineco, line 295: Echoing my previous comment: considering the purpose of this PR we should be able to assume there is no need to compare to whole payload to substrings anymore.
We could compare the Event Hub message to `helloEvent.Data()` though.

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -276,3 +375,11 @@ func withEventHubID(subscriptionID, resourceGroup, eventHubNS, eventHub string)
 		}
 	}
 }
+
+func withoutCloudEvents() targetOption {
```
antoineco, line 379: I'm sure we can come up with a better name, e.g.

https://github.com/triggermesh/triggermesh/blob/fb11bdeb11dbb3d1ce38c07a646886fd5cb55d43/test/e2e/targets/awss3/main.go#L327

config/301-azureeventhubstarget.yaml
cab105: Damn copy function.

test/e2e/targets/azureeventhubs/main.go
cab105: In this case I'd like to ensure that none of the CloudEvent data made it through without deserializing the string. Granted could just still to a full `Equals` instead of `ContainsSubstring`.

antoineco: If you compare Message.Data to HelloEvent.Data you can be sure there are no leftovers of context attributes.

antoineco (approved):

PR #535, opened 2022-02-01
Enable support for discardCEContext and send CloudEvent data as a JSON payload for the Azure Event Hubs target (resubmit)

cab105:
Resubmit with missing environment variable to indicate whether cloudevents should be omitted.

### DIFF
--- a/pkg/targets/reconciler/azureeventhubstarget/adapter.go
+++ b/pkg/targets/reconciler/azureeventhubstarget/adapter.go
@@ -17,6 +17,8 @@ limitations under the License.
 package azureeventhubstarget
 
 import (
+	"strconv"
+
 	corev1 "k8s.io/api/core/v1"
 	"knative.dev/eventing/pkg/reconciler/source"
 	"knative.dev/pkg/kmeta"
@@ -83,10 +85,6 @@ func makeTargetAdapterKService(target *v1alpha1.AzureEventHubsTarget, cfg *adapt
 		})
 	}
 
-	if target.Spec.DiscardCEContext {
-		envs = append(envs, corev1.EnvVar{Name: envDiscardCECtx, Value: "true"})
-	}
-
 	return resources.MakeKService(target.Namespace, name, cfg.Image,
 		resources.KsvcLabels(lbl),
 		resources.KsvcLabelVisibilityClusterLocal,
@@ -95,5 +93,6 @@ func makeTargetAdapterKService(target *v1alpha1.AzureEventHubsTarget, cfg *adapt
 		resources.KsvcPodEnvVars(envs),
 		resources.EnvVar(envHubNamespace, target.Spec.EventHubID.Namespace),
 		resources.EnvVar(envHubName, target.Spec.EventHubID.EventHub),
+		resources.EnvVar(envDiscardCECtx, strconv.FormatBool(target.Spec.DiscardCEContext)),
 	)
 }
--- a/pkg/targets/reconciler/azureeventhubstarget/adapter.go
+++ b/pkg/targets/reconciler/azureeventhubstarget/adapter.go
@@ -31,6 +31,7 @@ import (
 const (
 	adapterName = "azureeventhubstarget"
 
+	envDiscardCECtx    = "DISCARD_CE_CONTEXT"
 	envHubKeyName      = "EVENTHUB_KEY_NAME"
 	envHubNamespace    = "EVENTHUB_NAMESPACE"
 	envHubName         = "EVENTHUB_NAME"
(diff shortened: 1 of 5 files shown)


pkg/targets/reconciler/azureeventhubstarget/adapter.go
```diff
@@ -82,6 +83,10 @@ func makeTargetAdapterKService(target *v1alpha1.AzureEventHubsTarget, cfg *adapt
 		})
 	}
 
+	if target.Spec.DiscardCEContext {
+		envs = append(envs, corev1.EnvVar{Name: envDiscardCECtx, Value: "true"})
```
antoineco, line 87: nit: the conditional isn't even necessary. The user could have set this value explicitly to `false` in their spec, and expect this value to appear in the env vars of the adapter.

https://github.com/triggermesh/triggermesh/blob/12b2f1e86ba081390d82f1db852059777a0e1159/pkg/targets/reconciler/elasticsearchtarget/target_adapter.go#L80

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -190,12 +191,105 @@ var _ = Describe("Azure Event Hubs target", func() {
...
 
-					// NOTE: The payload will be a stringified version of the CloudEvent
-					Expect(payload).To(ContainSubstring(string(event.Data())))
-					Expect(payload).To(ContainSubstring("type: " + event.Type()))
-					Expect(payload).To(ContainSubstring("source: " + event.Source()))
-					Expect(payload).To(ContainSubstring("id: " + event.ID()))
+					err := json.Unmarshal(payload, &ce)
+					Expect(err).ToNot(HaveOccurred())
+
+					Expect(ce.Data()).To(ContainSubstring(string(event.Data())))
```
antoineco, line 199: I missed that one earlier, but it should be an `Equal` here.
The adapter isn't supposed to alter the events' data ("Hello, World!") while serializing it to raw bytes.

https://github.com/triggermesh/triggermesh/blob/12b2f1e86ba081390d82f1db852059777a0e1159/test/e2e/targets/awss3/main.go#L169

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -190,12 +191,105 @@ var _ = Describe("Azure Event Hubs target", func() {
...
+					Expect(ce.ID()).To(Equal(event.ID()))
+					Expect(ce.Type()).To(Equal(event.Type()))
+					Expect(ce.Subject()).To(Equal(event.Subject()))
```
antoineco, line 202: Subject isn't a mandatory context attribute, but `Source` is. We can keep the assertion of Subject, but Source shouldn't be missing.

https://github.com/triggermesh/triggermesh/blob/12b2f1e86ba081390d82f1db852059777a0e1159/test/e2e/targets/awss3/main.go#L168

It would also be nice to add an additional assertion to ensure that the rogue CE extension is preserved during serialization:

https://github.com/triggermesh/triggermesh/blob/12b2f1e86ba081390d82f1db852059777a0e1159/test/e2e/targets/awss3/main.go#L170-L171

test/e2e/targets/azureeventhubs/main.go
```diff
@@ -190,12 +191,105 @@ var _ = Describe("Azure Event Hubs target", func() {
...
+					// don't exit till event is received by handler or times out
+					select {
+					case <-eventReceivedChannel:
+					case <-evCtx.Done():
+						framework.FailfWithOffset(2, "timed out while waiting for event")
+					}
+				})
+
+				By("verifying the sent event", func() {
+					Expect(len(payload)).To(BeNumerically(">", 0))
```
antoineco, line 291: nit: Inferred by the Equal below (which will fail if we got no payload).

pkg/targets/reconciler/azureeventhubstarget/adapter.go
cab105: Good point.

test/e2e/targets/azureeventhubs/main.go
cab105: Thought I got that one too.

test/e2e/targets/azureeventhubs/main.go
cab105: Am aware Source is not mandatory, but it is being set by the hello event. Also good catch on the extension.

antoineco: What I mean is that, because the Hello event contains a Source attribute, this attribute should also be asserted.

Rationale: ensure that this field isn't being transformed by the target (intentionally or not), since the contract is that the target preserves all CE attributes.

antoineco (approved):
Approving, but it would be nice to include an assertion on `event.Source()`, since it is a mandatory CE attribute.

Right now, the assertion is on `event.Subject()`, which seems like a typo (?).

test/e2e/targets/azureeventhubs/main.go
cab105: Okay now I see what you mean. Added.

Issue #446, opened 2022-01-17
Wait for availability of Event Grid's System Topic before reporting readiness


antoineco:
Upon creation of an [Event Grid Event Subscription](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go@v61.2.0+incompatible/services/eventgrid/mgmt/2021-12-01/eventgrid#EventSubscription), Azure transparently provisions a [System Topic](https://docs.microsoft.com/en-us/azure/event-grid/system-topics) to back that subscription.

The topic is unavailable at first...

![image](https://user-images.githubusercontent.com/3299086/149808739-54eae405-a738-4f0b-b2de-fee86fbcc0d6.png)

...but eventually shows up in the console.

![image](https://user-images.githubusercontent.com/3299086/149808300-569edf88-c38c-4b9a-b8c9-3826fb0a6be3.png)

Our end-to-end tests surfaced that, during this time window, events produced by the source were discarded, which isn't intuitive because the source object reports "Ready".

We should consider checking the status of the associated System Topic before reporting readiness.

---

The code where the Event Subscription is created/reconciled:

https://github.com/triggermesh/triggermesh/blob/459d0a1068bb44ee5d0762fea677a01f0a6137fa/pkg/sources/reconciler/azureeventgridsource/event_subs.go#L91

`...`

https://github.com/triggermesh/triggermesh/blob/459d0a1068bb44ee5d0762fea677a01f0a6137fa/pkg/sources/reconciler/azureeventgridsource/event_subs.go#L132-L134

`...`

https://github.com/triggermesh/triggermesh/blob/459d0a1068bb44ee5d0762fea677a01f0a6137fa/pkg/sources/reconciler/azureeventgridsource/event_subs.go#L107

antoineco:
According to the Azure documentation, it is recommended to create the System Topic explicitly, and then attach a subscription to the topic:

> When you use CLI, REST, or Azure Resource Manager template, you can choose either of the above methods. We recommend **that you create a system topic first** and then create a subscription on the topic, as it's the latest way of creating system topics.
>
> -- https://docs.microsoft.com/en-us/azure/event-grid/system-topics#lifecycle-of-system-topics

This requires two Go clients:
- For the system topic: [SystemTopicsClient](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go@v61.2.0+incompatible/services/eventgrid/mgmt/2021-12-01/eventgrid#SystemTopicsClient)
[...]

PR #538, opened 2022-02-02
Mitigate flaky E2E test for Azure Event Grid source by allowing Azure some time for setting up the System Topic

### LINKED ISSUE #522: E2E test for Azure Event Grid source is flaky, fails to receive event after 10 min
This run, for example, shows two consecutive failures of that particular test:

https://github.com/triggermesh/triggermesh/runs/5009110595?check_suite_focus=true

Relevant code:
https://github.com/triggermesh/triggermesh/blob/fb11bdeb11dbb3d1ce38c07a646886fd5cb55d43/test/e2e/sources/azureeventgrid/main.go#L153-L165

Might be related to #446

antoineco:
Mitigates / closes #522

Temporary workaround until #446 is implemented.

I already included such workaround for the Azure Blob Storage source (\*) test in #445, but forgot to add it to the test for the Azure Event Grid source, which suffers from the same race issue.

_\* also based on Event Grid_

### DIFF
--- a/test/e2e/sources/azureblobstorage/main.go
+++ b/test/e2e/sources/azureblobstorage/main.go
@@ -163,6 +163,7 @@ var _ = Describe("Azure Blob Storage source", func() {
 				// topic upon creation of the Event Grid subscription by our reconciler (Blob Storage
 				// events are routed over Event Grid). The source shouldn't report Ready before this
 				// system topic is available, because events occuring prior to that are dropped.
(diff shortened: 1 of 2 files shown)


odacremolbap (approved):

antoineco:
Note: I ran this test locally with `ginkgo --until-it-fails ./test/e2e` and the test succeeded without failing (I stopped after 10 runs).

PR #556, opened 2022-02-14
Migrate Azure Service Bus to new SDK

### LINKED ISSUE #328: Migrate from `Azure/azure-service-bus-go` to `Azure/azure-sdk-for-go/sdk/messaging/azservicebus`
`Azure/azure-service-bus-go` was deprecated and archived: https://github.com/Azure/azure-service-bus-go.

Currently, we can't update any Azure AMQP-related package until we migrate ourselves (ref. #322 and #327), so we should do this rather soon.

FranBarrera:
Closes #328


go.mod
```diff
@@ -21,10 +21,11 @@ require (
 	cloud.google.com/go/pubsub v1.17.1
 	cloud.google.com/go/storage v1.19.0
 	cloud.google.com/go/workflows v1.1.1
-	github.com/Azure/azure-amqp-common-go/v3 v3.2.2
-	github.com/Azure/azure-event-hubs-go/v3 v3.3.16
+	github.com/Azure/azure-amqp-common-go/v3 v3.2.3
+	github.com/Azure/azure-event-hubs-go/v3 v3.3.17
 	github.com/Azure/azure-sdk-for-go v61.4.0+incompatible
-	github.com/Azure/azure-sdk-for-go/sdk/azidentity v0.12.0
+	github.com/Azure/azure-sdk-for-go/sdk/azidentity v0.13.0
```
antoineco, line 27: Updating this dependency is currently problematic because `azblob` — another of our imports — doesn't yet support it. The message thrown by `golangci-lint` is cryptic, but I think it is related to that updated.

I checked what version of `azservicebus` introduced this requirement, and it turns out to be quite recent. The following command should fix things:

```
go get github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus@v0.3.2
```

cab105: Is this a breaking change for azblob?

antoineco: Not really because we **can't** do anything with `azblob` right now. It hasn't been updated in a while and therefore doesn't support `azidentity v0.13.0`.

`azblob` was introduced when we migrated the E2E tests for Azure to this repo, but it has been causing some headaches because it's now blocking the update of `azidentity` and related packages (all the `sdk/resourcemanager/...` packages).

antoineco: Forget about it, things got resolved in #567

cab105 (approved):

antoineco:
Heads up: I have a few pending comments / questions but I'm not done with my review yet.

pkg/sources/adapter/azureservicebussource/adapter.go
```diff
@@ -180,27 +190,35 @@ func entityPath(entityID *v1alpha1.AzureResourceID) string {
...
-		sasErr := servicebus.NamespaceWithConnectionString(connStr)(ns)
-		if sasErr == nil {
-			return nil
+func clientFromEnvironment(entityID *v1alpha1.AzureResourceID) (*azservicebus.Client, error) {
+	// SAS authentication (token, connection string)
+	connStr := connectionStringFromEnvironment(entityID.Namespace, entityPath(entityID))
+	if connStr != "" {
+		client, err := azservicebus.NewClientFromConnectionString(connStr, nil)
+		if err != nil {
+			return nil, fmt.Errorf("creating client with connection string: %w", err)
```
antoineco, line 201: nit:
```suggestion
			return nil, fmt.Errorf("creating client from connection string: %w", err)
```

pkg/sources/adapter/azureservicebussource/adapter.go
```diff
@@ -180,27 +190,35 @@ func entityPath(entityID *v1alpha1.AzureResourceID) string {
 	}
 }
 
-// namespaceFromEnvironment mimics the behaviour of eventhub.NewHubFromEnvironment
```
antoineco, line 183: I see this comment was removed, but the behaviour still mimics `eventhub.NewHubFromEnvironment`. Keeping this specific part might help explain the design to readers who are unfamiliar with the code, since they know they should refer to the docs of the `eventhub` module to get the context of why we did it that way.

pkg/sources/adapter/azureservicebussource/adapter.go
```diff
@@ -180,27 +190,35 @@ func entityPath(entityID *v1alpha1.AzureResourceID) string {
...
-		}
+	// AAD authentication (service principal)
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create azure credentials: %w", err)
+	}
 
-		return fmt.Errorf("neither Azure Active Directory nor SAS token provider could be built - "+
-			"AAD error: %v, SAS error: %v", aadErr, sasErr)
+	if cred != nil {
```
antoineco, line 212: This check means that we could get both `err == nil` _and_ `cred == nil` after calling `azidentity.NewDefaultAzureCredential()`, which seems unlikely. The check looks safe to remove to me, since you shortcut on error above.

pkg/sources/adapter/azureservicebussource/adapter.go
```diff
@@ -180,27 +190,35 @@ func entityPath(entityID *v1alpha1.AzureResourceID) string {
...
 
-		// AAD authentication (service principal)
-		aadErr := servicebus.NamespaceWithEnvironmentBinding(entityID.Namespace)(ns)
-		if aadErr == nil {
-			return nil
-		}
+	// AAD authentication (service principal)
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create azure credentials: %w", err)
```
antoineco, line 209: nit:
```suggestion
		return nil, fmt.Errorf("unable to create Azure credentials: %w", err)
```

go.mod
```diff
@@ -25,6 +25,7 @@ require (
 	github.com/Azure/azure-event-hubs-go/v3 v3.3.16
 	github.com/Azure/azure-sdk-for-go v61.5.0+incompatible
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v0.13.1
+	github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus v0.3.1
```
antoineco, line 28: The current latest is `0.3.5`. Is anything stopping us from using that version?

pkg/sources/adapter/azureservicebussource/adapter.go
```diff
@@ -104,13 +107,20 @@ func NewAdapter(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClie
 		logger.Panicw("Unable to parse entity ID "+strconv.Quote(env.EntityResourceID), zap.Error(err))
 	}
 
-	ns, err := servicebus.NewNamespace(namespaceFromEnvironment(entityID))
+	entityPath := entityPath(entityID)
```
antoineco, line 110: This variable isn't used anywhere anymore, except in the error message at line 125. We could get rid of the variable, and just inline `strconv.Quote(entityPath(entityID))` inside the error message.

pkg/sources/adapter/azureservicebussource/adapter.go
```diff
@@ -230,22 +248,49 @@ func connectionStringFromEnvironment(namespace, entityPath string) string {
 //  Both (DataAction):
 //  - Microsoft.ServiceBus/namespaces/messages/receive/action
 func (a *adapter) Start(ctx context.Context) error {
+	maxMessages := 5
 	logging.FromContext(ctx).Info("Listening for messages")
 
-	handle := a.msgRcvr.Listen(ctx, servicebus.HandlerFunc(a.handleMessage))
-	<-handle.Done()
-	return handle.Err()
+	messages, err := a.msgRcvr.ReceiveMessages(ctx, maxMessages, nil)
```
antoineco, line 254: If we don't run this in a loop (cancelable by context like in e.g. the source for Amazon DynamoDB), won't the adapter terminate immediately after receiving and processing _one_ batch of messages?

pkg/sources/adapter/azureservicebussource/adapter_test.go
```diff
@@ -24,18 +24,10 @@ import (
...
-
 	adaptertest "knative.dev/eventing/pkg/adapter/v2/test"
 )
 
 func TestHandleMessage(t *testing.T) {
-	origMsgCompleteFn := messageCompleteFunc
-	messageCompleteFunc = testMessageCompleteFunc
-	t.Cleanup(func() {
-		messageCompleteFunc = origMsgCompleteFn
-	})
```
antoineco, line 37: Another good reason IMO to keep the entire processing logic inside `handleMessage()` is that we don't test whether messages are properly "completed" anymore, and that's is super critical.
This test was introduced in response to a bug which I discovered, and we should absolutely keep it to prevent regressions. Un-completed messages lead to messages being redelivered for a while, then dropped.

antoineco (changes requested):
I only reviewed the adapter but see 2 critical issues:
- The message completion is now excluded from tests.
- The adapter seems to return immediate after processing a single batch of messages.

pkg/sources/adapter/azureservicebussource/adapter.go
```diff
@@ -230,22 +248,49 @@ func connectionStringFromEnvironment(namespace, entityPath string) string {
...
+			return fmt.Errorf("error getting message data: %w", err)
+		}
+
+		msg := &Message{
+			Data:            body,
+			ReceivedMessage: *m,
+			LockToken:       stringifyLockToken((*uuid.UUID)(&m.LockToken)),
+		}
+
+		err = a.handleMessage(ctx, msg)
```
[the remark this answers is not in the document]
FranBarrera: If I add the `azservicebus.ReceviedMessage` in the func `handleMessage `we won't be able to have a test for it, because the `azservicebus.ReceviedMessage` doesn't have the `Data` parameter, to get it, we have to call the msg.Body() func

pkg/sources/adapter/azureservicebussource/adapter.go
```diff
@@ -230,22 +248,49 @@ func connectionStringFromEnvironment(namespace, entityPath string) string {
 //  Both (DataAction):
 //  - Microsoft.ServiceBus/namespaces/messages/receive/action
 func (a *adapter) Start(ctx context.Context) error {
+	maxMessages := 5
```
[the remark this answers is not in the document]
FranBarrera: We can put a higher number if we want, I wasn't sure what to put, so I decided to put 5, what is a good number to put?

pkg/sources/adapter/azureservicebussource/adapter.go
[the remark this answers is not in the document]
antoineco: OK fair enough, I can't think of a better way right now.

pkg/sources/adapter/azureservicebussource/adapter.go
```diff
@@ -180,27 +188,32 @@ func entityPath(entityID *v1alpha1.AzureResourceID) string {
 	}
 }
 
-// namespaceFromEnvironment mimics the behaviour of eventhub.NewHubFromEnvironment
-// by returning a servicebus.NamespaceOption that is suitable for the
+// clientFromEnvironment mimics the behaviour of eventhub.NewHubFromEnvironment
+// return a azservicebus.Client that is suitable for the
```
antoineco, line 192: The comment is cut. I think you meant to finish the sentence above and start a new one.

```suggestion
// clientFromEnvironment mimics the behaviour of eventhub.NewHubFromEnvironment.
// It returns a azservicebus.Client that is suitable for the
```

pkg/sources/adapter/azureservicebussource/adapter.go
```diff
@@ -180,27 +188,32 @@ func entityPath(entityID *v1alpha1.AzureResourceID) string {
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to create Azure credentials: %w", err)
+	}
 
-		return fmt.Errorf("neither Azure Active Directory nor SAS token provider could be built - "+
-			"AAD error: %v, SAS error: %v", aadErr, sasErr)
+	fqNamespace := entityID.Namespace + ".servicebus.windows.net"
+	client, err := azservicebus.NewClient(fqNamespace, cred, nil)
+	if err != nil {
+		return nil, fmt.Errorf("creating client with service principal: %w", err)
```
antoineco, line 214:
```suggestion
		return nil, fmt.Errorf("creating client from service principal: %w", err)
```

pkg/sources/adapter/azureservicebussource/adapter.go
```diff
@@ -230,22 +243,56 @@ func connectionStringFromEnvironment(namespace, entityPath string) string {
 //  Both (DataAction):
 //  - Microsoft.ServiceBus/namespaces/messages/receive/action
 func (a *adapter) Start(ctx context.Context) error {
+	maxMessages := 100
```
antoineco, line 246:
```suggestion
	const maxMessages = 100
```

pkg/sources/adapter/azureservicebussource/adapter.go
```diff
@@ -230,22 +243,56 @@ func connectionStringFromEnvironment(namespace, entityPath string) string {
...
+					return fmt.Errorf("error completing message: %w", err)
+				}
+			}
+		}
+	}
+
+	return nil
 }
 
 // handleMessage satisfies servicebus.HandlerFunc.
```
antoineco, line 287: This comment is outdated. Suggestion:

```suggestion
// handleMessage handles a single Service Bus message.
```

pkg/sources/adapter/azureservicebussource/adapter.go
```diff
@@ -317,3 +356,12 @@ func sanitizeEvent(validErrs event.ValidationError, origEvent *cloudevents.Event
 
 	return origEvent
 }
+
+// stringifyLockToken converts a UUID byte-array into its string representation.
```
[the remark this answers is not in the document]
antoineco: In this case, the issue is that we are processing some data inside `adapter.go`. In terms of organization of the code, it's hard for me to justify this move.

pkg/sources/adapter/azureservicebussource/adapter.go
```diff
@@ -230,22 +243,56 @@ func connectionStringFromEnvironment(namespace, entityPath string) string {
...
+				body, err := m.Body()
+				if err != nil {
+					return fmt.Errorf("error getting message data: %w", err)
+				}
+
+				msg := &Message{
+					Data:            body,
+					ReceivedMessage: *m,
+					LockToken:       stringifyLockToken((*uuid.UUID)(&m.LockToken)),
+				}
```
antoineco, line 269: (related to a previous conversation that is now marked a "Resolved").

Regardless of the signature of the `handleMessage()` function, it would be nice to keep the code organized.
The original intention was to have all the data processing inside `process.go`, otherwise the received message is mangled in different places, making the whole message processing logic hard to follow.

Keeping everything inside `process.go` is as simple as creating a function that performs the conversion from the native format to the custom `*Message`:

```go
func toMessage(rcvMsg *servicebus.ReceivedMessage) (*Message, err) {
	body, err := m.Body()
	if err != nil {
		return nil, fmt.Errorf("reading message data: %w", err)
	}

	return &Message{
		Data:            body,
		ReceivedMessage: rcvMsg,
		LockToken:       stringifyLockToken((*uuid.UUID)(&m.LockToken)),
	}, nil
}
```

antoineco (commented):
A few extra comments.

Also, the migration isn't complete until the reconciler is also using this new module.

pkg/sources/adapter/azureservicebussource/adapter.go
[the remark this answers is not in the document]
antoineco: Sounds better yes.

pkg/sources/adapter/azureservicebussource/adapter.go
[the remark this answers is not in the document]
antoineco: In this case, the issue is that we are processing some data inside `adapter.go`. In terms of organization of the code, it's hard for me to justify this move.

pkg/sources/adapter/azureservicebussource/adapter.go
FranBarrera: Should be done now, thanks!

antoineco (approved):

PR #593, opened 2022-02-24
Propagate bridge ID label to the adapters

tzununbekov:
Bridge components generated by the TIL have the label `bridges.triggermesh.io/id` with the bridge name. In this PR we're providing the way to propagate this label to the adapter pods. One of the use cases for this new thing - is matching pod logs with the bridges. To propagate labels other than bridge ID, the label should be added to the `labelsPropagationList` variable in `pkg/<API group>/reconciler/common/labels.go`.


pkg/flow/reconciler/reconcile_adapter.go
```diff
@@ -52,16 +53,25 @@ func MakeObsEnv(cfg source.ConfigAccessor) []corev1.EnvVar {
 }
 
 // MakeAdapterLabels adds generic label generation
-func MakeAdapterLabels(adapterName, name string) labels.Set {
+func MakeAdapterLabels(adapterName string, object kmeta.OwnerRefable) labels.Set {
```
antoineco, line 56: I think it would be cleaner if we kept a clear separation of concerns:
- generate a list of generic labels
- propagate labels from an external source

Those two things shouldn't be mixed an matched in a single function IMO.

(comment applies to all changed components)

tzununbekov: Well, then the function's name `MakeAdapterLabels` is quite confusing. Let me see how it can be changed.

antoineco: You right, I was mixing that up with the following code in the sources and routers:

https://github.com/triggermesh/triggermesh/blob/5499508d0f8f351d8b2f825f57b4e26e7e6113f2/pkg/sources/reconciler/common/adapter.go#L247-L256

Considering that we don't have a common way to reconcile things in TriggerMesh, I believe it's pointless to discuss style in this PR.
Feel free to ignore.

pkg/sources/reconciler/common/labels.go
```diff
@@ -37,3 +37,8 @@ const (
 	managedBy        = "triggermesh-controller"
 	componentAdapter = "adapter"
 )
+
+// LabelsPropagationList is the list of labels that should be propagated to the adapters.
```
antoineco, line 41: The linter didn't complain because this var is private, but this comment should start with a lowercase letter.

antoineco (approved):

pkg/targets/reconciler/deployment.go
```diff
@@ -43,6 +43,7 @@ type DeploymentReconciler interface {
 }
 
 // NewDeploymentReconciler creates the default implementation for Deployment reconciler.
+// Not used?
```
cab105, line 46: Quick search shows this should be safe to delete.

cab105 (approved):

PR #596, opened 2022-02-28
Set Iot Hub SKU to Basic in E2E test framework

### LINKED ISSUE #443: IoT Hub tier used in E2E tests doesn't allow multiple test workflows to run in parallel
It is quite frequent to see PRs being merged within a short interval. For example, dependabot updates, or multiple PRs reviewed by the same person in the same time frame.

When this occurs, GitHub Actions sometimes ends up running multiple E2E test suites in parallel (for different commits), and this breaks because the Free tier used in the Iot Hub source test doesn't permit creating multiple instances:

https://github.com/triggermesh/triggermesh/blob/1bd2f815f968e891be12d460bcfc2ff53a5e65c3/test/e2e/framework/azure/iothub.go#L55

The error:

```
Unable to create iothub: {"code":400019,"httpStatusCode":"BadRequest","message":"Max number of Iot Hubs exceeded for sku = Free, Max Allowed = 1, Current = 2 in the subscription Id: ***. If you contact a support representative please include this correlation identifier: <redacted>, timestamp: 2022-01-14 23:00:31Z, errorcode: IH400019."}
```

antoineco:
The free tier doesn't allow creating multiple instances, which causes failures while running multiple E2E test workflows in parallel.

Fixes #443

Proof that it works as expected, while running the same test twice in two different terminal windows:

image

### DIFF
--- a/test/e2e/framework/azure/iothub.go
+++ b/test/e2e/framework/azure/iothub.go
@@ -50,9 +50,8 @@ func CreateIOTHubComponents(ctx context.Context, subscriptionID, rg, region, nam
 		Location: &region,
 		Tags:     map[string]*string{E2EInstanceTagKey: to.StringPtr(name)},
 		SKU: &armiothub.SKUInfo{
-			Name:     armiothub.IotHubSKUF1.ToPtr(),
+			Name:     armiothub.IotHubSKUB1.ToPtr(),
 			Capacity: to.Int64Ptr(1),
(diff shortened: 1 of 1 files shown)


cab105:
Am in favor of this. Opted to keep the Free tier to minimize the cost to TriggerMesh for testing purposes. Although isn't there an additional limit for the basic tier on the number of IoT Hubs that can be created concurrently?

cab105 (approved):

PR #608, opened 2022-03-04
Add Tekton Target e2e tests

cab105:
Created a fairly straightforward e2e test for the Tekton Target.  The basic flow is:
1. Create the tekton task and pipeline citing the created task
2. Test task invocation via cloudevent and compare the "build" output (simple hello world greeting)
3. Test pipeline invocation via cloudevent and compare the "build" output (slightly different hello world greeting)
4. Ensure proper validation for the reaper attributes (not required by default)

### DIFF
--- a/test/e2e/framework/tekton/logs.go
+++ b/test/e2e/framework/tekton/logs.go
@@ -1,0 +1,44 @@
+/*
+Copyright 2022 TriggerMesh Inc.
+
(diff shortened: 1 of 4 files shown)


antoineco (changes requested):
Putting a hold because tests are currently failing, at least when executed in the CI environment.

PR #610, opened 2022-03-04
Relax e2e test invocation by introducing CODEOWNERS

cab105:
Based on a conversation with @antoineco relax some of the requirements for running the e2e tests by introducing a `CODEOWNERS` file to reflect who will be allowed to approve a PR affecting `.github/` for workflows and `test/e2e` for introducing additional tests.  By default, only members of the TriggerMesh organization can approve changes to these repos.

The action tweak is documented in: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-when-a-pull-request-is-approved

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -1,23 +1,17 @@
 name: End-to-End Testing
 
 on:
-  push:
-    # NOTE(antoineco): To avoid exposing repository secrets to user supplied
-    # code in pull requests, we only allow this workflow to run on the 'main'
-    # branch, where all code changes are expected to have been reviewed by
-    # maintainers.
-    #
-    # We may want to relax this in the future, on condition that safeguards are
-    # added to prevent the workflow from running automatically when an external
-    # contributor submits code changes inside the .github/workflows/ or
-    # test/e2e/ directories.
-    #
-    # Ref. https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
-    branches: [ main ]
+  pull_request_review:
+    types: [submitted]
+    # Run the e2e test once the PR has been approved to allow for a final check before
+    # merging the PR. One of the contributers metioned in CODEOWNERS must approve the PR
+    # if it touches .github/ or test/e2e/ to prevent potential credential leakage.
 
 jobs:
 
   e2e-triggermesh:
+    if: github.event.review.state == 'approved'
+
     name: Test TriggerMesh components
     runs-on: ubuntu-latest
 
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -6,6 +6,8 @@ on:
     # Run the e2e test once the PR has been approved to allow for a final check before
     # merging the PR. One of the contributers metioned in CODEOWNERS must approve the PR
     # if it touches .github/ or test/e2e/ to prevent potential credential leakage.
+  push:
+    branches: [main]
 
 jobs:
 
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -1,8 +1,10 @@
 name: End-to-End Testing
 
 on:
-  pull_request_review:
-    types: [submitted]
+  pull_request_target:
+    # Rely on default actions of opened, reopened, and syncrhonize to trigger a run
+  #pull_request_review:
+  #  types: [submitted]
     # Run the e2e test once the PR has been approved to allow for a final check before
     # merging the PR. One of the contributers metioned in CODEOWNERS must approve the PR
     # if it touches .github/ or test/e2e/ to prevent potential credential leakage.
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -3,13 +3,14 @@ name: End-to-End Testing
 on:
   pull_request_target:
     # Rely on default actions of opened, reopened, and syncrhonize to trigger a run
-  #pull_request_review:
-  #  types: [submitted]
+    # against the target repo to ensure the e2e tests function as intended.
+  pull_request_review:
+    types: [ submitted ]
     # Run the e2e test once the PR has been approved to allow for a final check before
     # merging the PR. One of the contributers metioned in CODEOWNERS must approve the PR
     # if it touches .github/ or test/e2e/ to prevent potential credential leakage.
   push:
-    branches: [main]
+    branches: [ main ]
 
 jobs:
 
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,5 @@
+# Reflect current maintainers allowed to approve pull requests that touch the .github
+# and test/e2e repos. This is the TriggerMesh engineering team
+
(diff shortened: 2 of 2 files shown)


CODEOWNERS
```diff
@@ -0,0 +1,5 @@
+# Reflect current maintainers allowed to approve pull requests that touch the .github
+# and test/e2e repos. This is the TriggerMesh engineering team
```
antoineco, line 2:
```suggestion
# and test/e2e directories. This is the TriggerMesh engineering team.
```

CODEOWNERS
```diff
@@ -0,0 +1,5 @@
...
+
+/.github/	@triggermesh/engineering
+/test/e2e/	@triggermesh/engineering
```
antoineco, line 5: This subdirectory is sufficient I think.
We probably don't want to have to validate if someone sends a pull request that touches e.g. the Dependabot config.

```suggestion
/.github/workflows/	@triggermesh/engineering
/test/e2e/		@triggermesh/engineering
```

.github/workflows/e2e-tests.yaml
```diff
@@ -1,23 +1,17 @@
...
-    # branch, where all code changes are expected to have been reviewed by
-    # maintainers.
-    #
-    # We may want to relax this in the future, on condition that safeguards are
-    # added to prevent the workflow from running automatically when an external
-    # contributor submits code changes inside the .github/workflows/ or
-    # test/e2e/ directories.
-    #
-    # Ref. https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
-    branches: [ main ]
```
antoineco, line 16: The comment may not apply anymore, but we still want to run on pushes to `main`.

.github/workflows/e2e-tests.yaml
```diff
@@ -1,23 +1,17 @@
...
+  pull_request_review:
+    types: [submitted]
+    # Run the e2e test once the PR has been approved to allow for a final check before
+    # merging the PR. One of the contributers metioned in CODEOWNERS must approve the PR
+    # if it touches .github/ or test/e2e/ to prevent potential credential leakage.
```
antoineco, line 8: Note: this might make it difficult to get quick feedback when a PR is sent.

Let's take the example of #608, and let's assume that it was sent by someone who isn't a code owner.
The PR is broken to start with (in the context of the repo), but until someone leaves an approval (which might take days) you won't know that by looking at the CI status.

Generally speaking, if I send a PR adding a new test, and expect the CI to tell me whether it passes, I'll have to wait for an approval. But maybe reviewers want to base their decision on the outcome of the CI run, which leads to a chicken and egg situation.

It might not be a bad thing! But worth thinking about it.
Let's try and see if that workflow works for us.

antoineco:
So far working as intended 😄 

image

antoineco:
I don't know why GitHub is complaining about the syntax of the file.
In doubt, I explicitly assigned `write` permissions to the referenced team in this repo (but all its individual members already had `write` permission).

CODEOWNERS
cab105: Wasn't too sure since encompassing `/.github` would include the configs stored inside as well.

.github/workflows/e2e-tests.yaml
cab105: I'd agree, but this way we can ensure there are eyes on the PR in case they slipped in something malicious. I'm thinking of the "here's a clean PR, oh wait let me commit something grab the credentials, and then a follow-up commit to undo it" scenario. Which can become more troublesome if they squash their commits after the experiment and hide the attempt (except to show an extra push was made)

cab105:
> I don't know why GitHub is complaining about the syntax of the file. In doubt, I explicitly assigned `write` permissions to the referenced team in this repo (but all its individual members already had `write` permission).
> 
> I think you need a commit to trigger a re-check.

Yeah the only thing I could think of is that it is comparing the membership against my repo of cab105/triggermesh instead of triggermesh/triggermesh.

CODEOWNERS
antoineco: What I mean is that `.github/` doesn't contain only Actions related files.
The Dependabot config, and potentially other files such issue templates, are also in this directory. For those, we don't need to approve workflow runs (because they don't access repository secrets).

.github/workflows/e2e-tests.yaml
antoineco: GitHub thought about that. Any commit to the PR schedules a workflow run that needs to be approved (it happened to me recently while contributing a change to `ko`).
Approving one e2e run doesn't automatically approve subsequent runs.

antoineco (approved):

antoineco:
Looks like all the tests that rely on the repository's secrets failed.
We need to get to the bottom of this.

antoineco:
Maybe relevant:

> Due to the dangers inherent to automatic processing of PRs, GitHub’s standard `pull_request` workflow trigger by default prevents write permissions and **secrets access** to the target repository. However, in some scenarios such access is needed to properly process the PR. To this end the `pull_request_target` workflow trigger was introduced.
>
> -- https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

cab105:
> Maybe relevant:
> 
> > Due to the dangers inherent to automatic processing of PRs, GitHub’s standard `pull_request` workflow trigger by default prevents write permissions and **secrets access** to the target repository. However, in some scenarios such access is needed to properly process the PR. To this end the `pull_request_target` workflow trigger was introduced.
> > -- https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

A few things to keep in mind:
1. My understanding of github actions is that they don't go into effect until they are merged into the main branch, so the changes in the PR won't go into effect until after it is merged.
2. Even now, since the PR is based on my fork of the tm repo, I don't have any of the required secrets associated with our e2e tests so it would make sense that the tests would fail.
3. I did see the thing about `pull_request_target`, and was under the impression that, while this would do the right thing, the secrets could get leaked due to a malicious action.

antoineco:
That's the thing: it's risky to expose secrets to code coming from external repos, therefore we require approvals for any changes to sensitive parts of the code before executing workflows that do consume those secrets.

The code in PRs will almost always come from forks (I submit PRs from a fork too!), so this has to work, otherwise the situation will be the same as it was before this PR.

(Worth noting: PRs originating from `tm/tm` are not subject to restrictions on secrets access, because GitHub assumes that, if somebody has _direct_ write access to the repo (the @triggermesh/engineering team) they are also allowed to run those workflows.
In other words, we wouldn't be having this discussion if we agreed not to accept contributions from forks, but that'd be a hard sell for an open source project.)

My understanding after reading the linked blog is that `pull_request_target` is the only trigger that _can_ expose repository secrets to PRs from external repos. If that's the case, we will need to rethink the approval flow suggested in this PR, which would be a pity.

antoineco:
Funny, looks like `pull_request_target` doesn't trigger a run.
The workflow doesn't appear in the list of Checks anymore.

cab105:
To be expected. The default types (as enabled) will run when the PR is opened, synchronized (rebased with HEAD), or re-opened: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

PR #611, opened 2022-03-05
Handle lifecycle of Event Grid system topics in Azure Event Grid source

### LINKED ISSUE #446: Wait for availability of Event Grid's System Topic before reporting readiness

antoineco:
Microsoft recommends managing the lifecycle of Azure Event Grid's system topics manually, and creating event subscriptions directly to these topics, instead of relying on the automatic creation of system topics.

Closes #446


pkg/sources/reconciler/azureeventgridsource/errors.go
```diff
@@ -0,0 +1,94 @@
+/*
+Copyright 2021 TriggerMesh Inc.
```
FranBarrera, line 2: The year should be 2022, it also happens in the others files

antoineco: I'll do it, but it's absolutely not mandatory to apply the Apache v2 license. Other Apache projects don't bump the year in their headers. Maintainers even explicitly ask contributors not to do it (I don't know the reason).

FranBarrera: Uh didn't know that, good to know!

antoineco: Done in 65ef49410449935a8f2cf261dc5472d9af79e737

FranBarrera (approved):

PR #618, opened 2022-03-07
Fix incorrect parsing of Azure subresource IDs

### LINKED ISSUE #617: Azure Resource IDs of non-namespaced subresources aren't parsed properly
The resource ID string `/subscriptions/s/resourceGroups/rg/providers/rp/rt/rn/srt/srn` should be parsed with an empty `Namespace`, since there is no `/namespaces/` segment in the resource ID:

```go
v1alpha1.AzureResourceID{SubscriptionID:"s", ResourceGroup:"rg", ResourceProvider:"rp", Namespace:"", ResourceType:"rt", ResourceName:"rn", SubResourceType:"srt", SubResourceName:"srn"}
```

Instead, the `ResourceName` "rn" becomes the `Namespace` and the subresource "srt/srn" becomes the `ResourceType` and `ResourceName`:

```go
v1alpha1.AzureResourceID{SubscriptionID:"s", ResourceGroup:"rg", ResourceProvider:"rp", Namespace:"rn", ResourceType:"srt", ResourceName:"srn", SubResourceType:"", SubResourceName:""}
```

antoineco:
Fixes #617


pkg/apis/sources/v1alpha1/azure_common_types.go
```diff
@@ -156,17 +159,25 @@ func (rID *AzureResourceID) UnmarshalJSON(data []byte) error {
...
 	var namespace string
-	if len(sections) >= azureNamespacedResourceResourceIDSplitElements {
-		namespace = sections[namespaceIdx]
-		resourceType = sections[resourceTypeNsIdx]
-		resourceName = sections[resourceNameNsIdx]
-		if namespace == "" || resourceType == "" || resourceName == "" {
-			return errAzureResourceIDEmptyAttrs
+	var subresourceType string
+	var subresourceName string
+	if len(sections) >= azureSubResourceResourceIDSplitElements {
```
odacremolbap, line 164: I am not sure of the relationship between `azureSubResourceResourceIDSplitElements` and `azureNamespacedSubResourceResourceIDSplitElements` but line 181 sounds like could be overwriting `subresourceType` and `subresourceName `.

I am not sure if a `switch ... case` that replaces those 2 if could improve the readability, maybe not.

odacremolbap (approved):
Made one comment (out of my ignorance, dismiss if it does not apply)

pkg/apis/sources/v1alpha1/azure_common_types.go
antoineco: Maybe yes, I can't answer because the style is quite arbitrary here.

I remember having a large `switch` case in my first version, but the logic ended up being convoluted due to repeated code in each `case`. Each step was at least setting the subscription and resource group, while adding more or less stuff on top based on the type of resource.

Because Azure resource IDs (mostly) read from left to right to determine the "depth" of the resource, I found it easier to write a _step-by-step parsing logic_ here (even if that means overwriting a previous field).

I could mix a switch cases in there if you feel strongly about it, but my preference would be to keep this PR focused on fixing the bug, and only that.

PR #626, opened 2022-03-08
Fix Codegen Api Groups generation

FranBarrera:
Codegen was failing with make 4.3.

The API Groups list generated was wrong: 
`sources/v1alpha1 targets/v1alpha1 flow/v1alpha1 extensions/v1alpha1 routing/v1alpha1,`

This changes fixes that problem (Thanks @antoineco):
`sources/v1alpha1,targets/v1alpha1,flow/v1alpha1,extensions/v1alpha1,routing/v1alpha1`

### DIFF
--- a/hack/inc.Codegen.mk
+++ b/hack/inc.Codegen.mk
@@ -19,8 +19,8 @@ codegen: $(generators)
 
 # http://blog.jgc.org/2007/06/escaping-comma-and-space-in-gnu-make.html
 comma := ,
-null      :=
-space     := $(null) $(null)
+null  :=
+space := $(null) $(null)
 
 deepcopy:
 	@echo "+ Generating deepcopy funcs for $(API_GROUPS)"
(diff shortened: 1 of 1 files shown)


hack/inc.Codegen.mk
```diff
@@ -19,8 +19,8 @@ codegen: $(generators)
 
 # http://blog.jgc.org/2007/06/escaping-comma-and-space-in-gnu-make.html
 comma := ,
-space :=
-space +=
+null      :=
+space     := $(null) $(null)
```
antoineco, line 23:
```suggestion
null  :=
space := $(null) $(null)
```

antoineco (approved):
We don't need to add all these spaces. Otherwise looks sensible to me 👍 
@tzununbekov I remember you also had that issue.

tzununbekov:
>  remember you also had that issue.

yeah, I remember something. I guess I've fixed it by switching from Linux...

hack/inc.Codegen.mk
FranBarrera: Uh totally agree!

PR #628, opened 2022-03-08
Revert addition of  hook

cab105:
The addition of the `pull_request_target` hook was intended to allow for the e2e test to run using the credentials on the upstream repo, however it is not only _doesn't_ run, but may expose more of the environment then is intended. This PR will revert the hook addition ensure that it runs upon approval only.

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -1,9 +1,6 @@
 name: End-to-End Testing
 
 on:
-  pull_request_target:
-    # Rely on default actions of opened, reopened, and syncrhonize to trigger a run
-    # against the target repo to ensure the e2e tests function as intended.
   pull_request_review:
     types: [ submitted ]
     # Run the e2e test once the PR has been approved to allow for a final check before


.github/workflows/e2e-tests.yaml
```diff
@@ -1,9 +1,6 @@
 name: End-to-End Testing
 
 on:
-  pull_request_target:
-    # Rely on default actions of opened, reopened, and syncrhonize to trigger a run
-    # against the target repo to ensure the e2e tests function as intended.
   pull_request_review:
     types: [ submitted ]
     # Run the e2e test once the PR has been approved to allow for a final check before
```
antoineco, line 6: The whole PR should be reverted, not just this hook, shouldn't it?

In the original PR, we saw that secrets were not made available to workflows triggered by `pull_request_review`, which makes it unusable (it will yield failures on _every_ PR).

According to the docs, `pull_request_target` is the only hook that provides access to secrets, but it needs to be handled with extra care.

cab105:
Closing this pull request, and will reopen using an alternate branch.

PR #629, opened 2022-03-08
Fix e2e test workflow

cab105:
Remove the hooks that were introduced and intended to resolve the e2e testing issue of the credentials not being populated until merge.

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -1,22 +1,23 @@
 name: End-to-End Testing
 
 on:
-  pull_request_target:
-    # Rely on default actions of opened, reopened, and syncrhonize to trigger a run
-    # against the target repo to ensure the e2e tests function as intended.
-  pull_request_review:
-    types: [ submitted ]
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-tests.yaml
```diff
@@ -1,14 +1,6 @@
...
-  pull_request_target:
-    # Rely on default actions of opened, reopened, and syncrhonize to trigger a run
-    # against the target repo to ensure the e2e tests function as intended.
-  pull_request_review:
-    types: [ submitted ]
-    # Run the e2e test once the PR has been approved to allow for a final check before
-    # merging the PR. One of the contributers metioned in CODEOWNERS must approve the PR
-    # if it touches .github/ or test/e2e/ to prevent potential credential leakage.
   push:
     branches: [ main ]
```
antoineco, line 5: Could you please revert the entire file to its original revision?
I originally wrote a long comment explaining why we only run E2E tests on `main`, and I think it's important to preserve it.

antoineco (approved):

PR #643, opened 2022-03-12
ci: wait before attaching artifacts to releases

sameersbn:
the `release` job has been failing while attempting to attach release artifacts immediately after creating the release

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -214,6 +214,14 @@ jobs:
             PRE_RELEASE=${CIRCLE_TAG/${CIRCLE_TAG%-rc[0-9]*}/}
             github-release delete -u ${CIRCLE_PROJECT_USERNAME} -r ${CIRCLE_PROJECT_REPONAME} -t ${CIRCLE_TAG} 2>/dev/null ||:
             ./hack/release-notes.sh ${CIRCLE_TAG} | github-release release ${PRE_RELEASE:+-p} -u ${CIRCLE_PROJECT_USERNAME} -r ${CIRCLE_PROJECT_REPONAME} -t ${CIRCLE_TAG} -d -
+
+            # wait for github to report the release before attaching the artifacts
+            max_tries=5
+            for _ in $(seq 1 ${max_tries}); do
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -214,6 +214,7 @@ jobs:
             PRE_RELEASE=${CIRCLE_TAG/${CIRCLE_TAG%-rc[0-9]*}/}
             github-release delete -u ${CIRCLE_PROJECT_USERNAME} -r ${CIRCLE_PROJECT_REPONAME} -t ${CIRCLE_TAG} 2>/dev/null ||:
             ./hack/release-notes.sh ${CIRCLE_TAG} | github-release release ${PRE_RELEASE:+-p} -u ${CIRCLE_PROJECT_USERNAME} -r ${CIRCLE_PROJECT_REPONAME} -t ${CIRCLE_TAG} -d -
+            sleep 3
```
antoineco, line 217: A more robust option would be to run `github-release info` in a loop until it succeeds. Eventual consistency can be quite unpredictable!

sameersbn: thanks for the review. taking you suggestion i've added the command to check for the release with 5 retries in intervals of 1 second between retries.

.circleci/config.yml
```diff
@@ -214,6 +214,16 @@ jobs:
...
             ./hack/release-notes.sh ${CIRCLE_TAG} | github-release release ${PRE_RELEASE:+-p} -u ${CIRCLE_PROJECT_USERNAME} -r ${CIRCLE_PROJECT_REPONAME} -t ${CIRCLE_TAG} -d -
+
+            # wait for github to report the release before attaching the artifacts
+            try=0
+            max_tries=5
+            while [ ${try} -lt ${max_tries} ]; do
+              github-release info -u ${CIRCLE_PROJECT_USERNAME} -r ${CIRCLE_PROJECT_REPONAME} -t ${CIRCLE_TAG} >/dev/null && break
+              try=$(expr ${try} + 1)
+              sleep 1
+            done
```
antoineco, line 225: Tip: No need to increment variables to do something n-times.

```suggestion
            for _ in $(seq 1 5); do
              github-release info -u ${CIRCLE_PROJECT_USERNAME} -r ${CIRCLE_PROJECT_REPONAME} -t ${CIRCLE_TAG} >/dev/null && break
              sleep 1
            done
```

antoineco (approved):

Issue #616, opened 2022-03-07
XSLTTransformation leaks memory


odacremolbap:
XSLTTransformation leaks memory.
After 50 minutes of 200 req/s the pod is evicted because the adapter's memory grows over available.

![pod1-adapter-memory](https://user-images.githubusercontent.com/9891289/157002457-a6ba74ef-50ec-4c9d-b45d-fea2cb8acd84.png)

PR #651, opened 2022-03-13
XSLTTransformation: fix/replace xslt lib

odacremolbap:
Fixes https://github.com/triggermesh/triggermesh/issues/616 by replacing XSLT library.


antoineco (approved):

Makefile
```diff
@@ -53,7 +53,9 @@ GOPKGS_SKIP_TESTS  = $(GOMODULE)/pkg/sources/reconciler/ibmmqsource \
                      $(GOMODULE)/pkg/targets/adapter/ibmmqtarget \
                      $(GOMODULE)/pkg/sources/adapter/ibmmqsource \
                      $(GOMODULE)/cmd/ibmmqsource-adapter \
-                     $(GOMODULE)/cmd/ibmmqtarget-adapter
+                     $(GOMODULE)/cmd/ibmmqtarget-adapter \
+                     $(GOMODULE)/cmd/xslttransformation-adapter \
+                     $(GOMODULE)/pkg/flow/adapter/xslttransformation
```
antoineco, line 58: If we're skipping tests here, we might be able to remove the `apt-get install libxml2` from the CircleCI config, in the `test` workflow.

antoineco:
@odacremolbap the linter is currently failing because it cannot build the package due to missing C libs.

The simplest solution, found by @tzununbekov, is to prevent building those packages based on Go build tags.

So far we've been doing it that way:

https://github.com/triggermesh/triggermesh/blob/0ffd100f029a91d7367a8e5da5192909e3c87402/pkg/sources/adapter/ibmmqsource/adapter.go#L1

https://github.com/triggermesh/triggermesh/blob/0ffd100f029a91d7367a8e5da5192909e3c87402/.golangci.yaml#L14-L15

cmd/xslttransformation-adapter/Dockerfile
```diff
@@ -14,24 +15,33 @@ RUN go build -o /xslttransformation-adapter ./cmd/xslttransformation-adapter
...
 FROM gcr.io/distroless/base-debian11:nonroot
 
 # (!) COPY follows symlinks
 COPY --from=builder \
     /usr/lib/x86_64-linux-gnu/libxml2.so.2 \
+    /usr/lib/x86_64-linux-gnu/libxslt.so.1 \
+    /usr/lib/x86_64-linux-gnu/libexslt.so.0 \
     /usr/lib/x86_64-linux-gnu/libicuuc.so.67 \
+    /usr/lib/x86_64-linux-gnu/libgcrypt.so.20 \
+    /lib/x86_64-linux-gnu/libgpg-error.so.0 \
```
cab105, line 44: I really hope you are copying all dependent libraries as well otherwise I can see some very interesting runtime behavior due to unresolved symbols.

antoineco: For my own knowledge, does the output of `ldd` include all libraries the compiled app depends on?

cab105: It does as well as whether the library is resolvable.

odacremolbap:
> So far we've been doing it that way:

thanks, I think I will still have problems with `make` targets `build` and `test` on `ibm mq` and `xslt`.
Will do as you suggest to solve CI and make this PR look green.

Makefile
antoineco: Ping.

Those lines are no longer necessary:

https://github.com/triggermesh/triggermesh/blob/5911ee64d89d8d5b7d4e7736f1ff183b08f8d9ff/.circleci/config.yml#L38-L41

antoineco:
@odacremolbap by simply looking at the `Makefile`, I can tell for sure that `make build` _is_ broken.
This goal tries to build `cmd/*`, which won't work considered that some of these require extra libraries.

I think the reason why this wasn't caught is because nobody on the team ever uses `make build`. People tend to build only the component they are working on (I guess), and probably from the IDE directly (again, guessing, speaking about my own workflow).

cab105:
@antoineco I wouldn't say that nobody uses `make build`. It is still my primary way of building the repo, but I've coded around it:
```diff
-build: $(COMMANDS)  ## Build all artifacts
+build: $(filter-out $(CUSTOM_BUILD_BINARIES), $(COMMANDS))  ## Build all artifacts
```

antoineco (approved):

odacremolbap:
still not the right libraries ... I'll test locally and fix.

PR #655, opened 2022-03-15
Initial port of JQ Transformation

### LINKED ISSUE #637: Migrate jq transformation
We need to move the jq transformation from https://github.com/triggermesh/poc-adapters/tree/main/jqtransformation to this repo.

It most likely needs to:

- [ ] pass linter
- [ ] add unit test

cab105:
Introduce a new transformation based on the [JSON Query language](https://stedolan.github.io/jq/).

Closes #637


cmd/triggermesh-controller/main.go
```diff
@@ -19,6 +19,8 @@ package main
 import (
 	"knative.dev/pkg/injection/sharedmain"
 
+	"github.com/triggermesh/triggermesh/pkg/flow/reconciler/jqtransformation"
```
antoineco, line 22: This import isn't _that_ special, it can be grouped with other tm/tm imports.

cmd/triggermesh-controller/main.go
```diff
@@ -167,6 +169,7 @@ func main() {
 		// transformation
 		synchronizer.NewController,
 		transformation.NewController,
+		jqtransformation.NewController,
```
antoineco, line 172: Let's preserve the alphabetical order like we do with other components categories.

config/500-controller.yaml
```diff
@@ -171,6 +171,8 @@ spec:
           value: ko://github.com/triggermesh/triggermesh/cmd/transformation-adapter
         - name: XMLTOJSONTRANSFORMATION_IMAGE
           value: ko://github.com/triggermesh/triggermesh/cmd/xmltojsontransformation-adapter
+        - name: JQTRANSFORMATION_IMAGE
```
antoineco, line 174: Same comment about the alphabetical order.

config/samples/flows/jqtransformation/200-jqtransformation.yaml
```diff
@@ -0,0 +1,26 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+apiVersion: flow.triggermesh.io/v1alpha1
+kind: JQTransformation
+metadata:
+  name: testjqtransform
+spec:
+  query: .foo | ..
```
antoineco, line 21: For my own curiosity, what does `| ..` do?

pkg/flow/reconciler/testing/factory.go
```diff
@@ -59,6 +59,7 @@ func MakeFactory(ctor Ctor) rt.Factory {
 		// the controller.Reconciler uses an internal client to handle
 		// component objects
 		ctx, xsltclient := faketminjectionclient.With(ctx, ls.GetXSLTTransformationObjects()...)
+		ctx, jqclient := faketminjectionclient.With(ctx, ls.GetJQTransformationObjects()...)
```
antoineco, line 62: This doesn't look right.
We only need one client interface for _all_ flow types, not one _per_ type.

So, instead of "xsltclient" this should be "flowclient".

See how it's done in sources:
https://github.com/triggermesh/triggermesh/blob/0ffd100f029a91d7367a8e5da5192909e3c87402/pkg/sources/reconciler/testing/factory.go#L64

pkg/flow/reconciler/testing/listers.go
```diff
@@ -87,11 +92,16 @@ func (l *Listers) GetServingObjects() []runtime.Object {
...
 }
 
-// GetXSLTTransformationLister returns a Lister for GoogleSheetTarget objects.
+// GetXSLTTransformationLister returns a Lister for XSLTTransformation objects.
 func (l *Listers) GetXSLTTransformationLister() flowlisters.XSLTTransformationLister {
 	return flowlisters.NewXSLTTransformationLister(l.IndexerFor(&flowv1alpha1.XSLTTransformation{}))
 }
 
+// GetJQTransformationLister returns a Lister for JQTransformation objects.
+func (l *Listers) GetJQTransformationLister() flowlisters.JQTransformationLister {
```
antoineco, line 101: (related to previous comment)

Not needed. We could just collect all flow objects here:
https://github.com/triggermesh/triggermesh/blob/0ffd100f029a91d7367a8e5da5192909e3c87402/pkg/sources/reconciler/testing/listers.go#L87-L89

config/samples/flows/jqtransformation/200-jqtransformation.yaml
cab105: `|` allows for chaining queries together and `..` will recursively descend into the object.  Given the example in the sample, `.foo` will retrieve the value keyed off `foo` while `..` will retrieve all objects. Granted it is a contrived example for this scenario as `.foo` would do the same thing.

antoineco: Exactly what I thought. The example contains unnecessary bits.

On the other hand, we could turn this example into an actual _transformation_. `jq` can do very powerful things.

I can come up with a query that transforms the JSON input into another JSON object if you like the idea.

pkg/flow/reconciler/testing/listers.go
cab105: This one I have a little more of an objection to as my understanding is that the `Lister` is specific to the type as per: https://github.com/triggermesh/triggermesh/blob/0fa76cf2bd91f11c9d211f3098fe584dcafaf3b6/pkg/sources/reconciler/testing/listers.go#L107 (referencing one of the many listeres defined for the reconciler lister).

config/samples/flows/jqtransformation/200-jqtransformation.yaml
cab105: You're right it can, but at this point we just want to highlight that we can pass in a JQ style query. Did convert this into an iteration of arrays and remapping two fields if that helps.

pkg/flow/reconciler/testing/listers.go
antoineco: The two are unrelated. The lister you're referring to is used in a different context.

config/samples/flows/jqtransformation/200-jqtransformation.yaml
antoineco: Nice 👍

pkg/flow/reconciler/testing/listers.go
cab105: Is it?  I must be missing something because I'm referencing it in a similar context. Unless you're referring to mine using the `NewKServiceReconciler` function to initialize the reconciler.

antoineco: Sorry I got completely mixed up with my previous comment on the Factory and the usage of `GetFlowObjects()` 🤦
Yes you're right, we also need this one of course.

cab105: Whew, glad it wasn't me for a change. :)

antoineco (approved):
Besides that "floating" import in `cmd/triggermesh-controller/main.go` it looks good to me.

cab105:
> Besides that "floating" import in `cmd/triggermesh-controller/main.go` it looks good to me.

You mean this?
```go
import (
	pkgadapter "knative.dev/eventing/pkg/adapter/v2"

	"github.com/triggermesh/triggermesh/pkg/flow/adapter/jqtransformation"
)
```

antoineco:
No, the file I mentioned has an import that's on its own.

cab105:
That's the way Github is rendering the comment block. When looking at the raw file from (https://github.com/triggermesh/triggermesh/blob/534f9db4c563c1a977e87ebdd48d3f553f537cee/cmd/triggermesh-controller/main.go), I get:

```go
	"knative.dev/pkg/injection/sharedmain"

	"github.com/triggermesh/triggermesh/pkg/flow/reconciler/jqtransformation"
	"github.com/triggermesh/triggermesh/pkg/flow/reconciler/synchronizer"
	"github.com/triggermesh/triggermesh/pkg/flow/reconciler/transformation"
	"github.com/triggermesh/triggermesh/pkg/flow/reconciler/xmltojsontransformation"
	"github.com/triggermesh/triggermesh/pkg/flow/reconciler/xslttransformation"
	"github.com/triggermesh/triggermesh/pkg/function"
```

PR #657, opened 2022-03-17
Make Bumblebee's Handler implement Adapter interface

antoineco:
In order to implement custom metrics, we should be able to access the standard environment variables propagated by Knative, such as `NAME` and `NAMESPACE`.

This PR is a refactoring that prepares the component for the introduction of those custom metrics.

It simply moves the initialization logic that is currently inside the `main()` to `NewHandler()`, so that we can consume the aforementioned environment variables during the Handler's initialization.

### DIFF
--- a/pkg/flow/adapter/transformation/handler.go
+++ b/pkg/flow/adapter/transformation/handler.go
@@ -21,9 +21,15 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"net/url"
 	"strings"
 
+	"go.uber.org/zap"
+
 	cloudevents "github.com/cloudevents/sdk-go/v2"
+
+	pkgadapter "knative.dev/eventing/pkg/adapter/v2"
+	"knative.dev/pkg/logging"
 
 	"github.com/triggermesh/triggermesh/pkg/apis/flow/v1alpha1"
 	"github.com/triggermesh/triggermesh/pkg/flow/adapter/transformation/common/storage"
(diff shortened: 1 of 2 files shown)


pkg/flow/adapter/transformation/handler.go
```diff
@@ -43,16 +68,29 @@ type ceContext struct {
 	Extensions                  map[string]interface{} `json:"Extensions,omitempty"`
 }
 
-// NewHandler creates Handler instance.
-func NewHandler(context, data []v1alpha1.Transform) (Handler, error) {
-	contextPipeline, err := newPipeline(context)
+// NewHandler satisfies pkgadapter.AdapterConstructor by creating a Handler instance.
+func NewHandler(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, _ cloudevents.Client) pkgadapter.Adapter {
```
antoineco, line 72: @tzununbekov I _think_ we should be able to use the client passed to this function as-is when `K_SINK` is set?
It's not the purpose of this PR so I decided not to touch the existing logic, but it would be interesting to know your opinion about it.

tzununbekov:
@antoineco sorry, I though I pushed my local branch already. Does it overlap with this?
https://github.com/triggermesh/triggermesh/compare/bumblebee-stats?expand=1

pkg/flow/adapter/transformation/handler.go
tzununbekov: thanks for the note, I'll take a closer look at this tomorrow

antoineco:
Oops yes I think it overlaps. But it would probably still be valuable to merge this refactoring separately, because then the implementation of the stats reporter becomes quite a small PR.

Would that work if we still review this and I make you the author by amending my commit?

antoineco:
Forget what I said. It doesn't matter. I see our two versions differ too much.

PR #678, opened 2022-03-22
Refactor internal API interfaces to use generic names

antoineco:
This PR touches the `sources` and `routing` API groups exclusively.

In these API groups, we have interfaces that define what a type should implement in order to be reconcilable by the corresponding _generic reconciler_ (the reconciler that manages adapters for all types in the API group):

- `type EventSource interface { ... }` for sources
- `type Router interface { ... }` for routers

If you look closely, these two interfaces are exactly the same (with the exception of the `GetStatusManager()`, which returns a typed struct).

I would like to take that concept of _component interface_ and matching _generic reconciler_ to other API groups soon, mostly to reduce code duplication across reconcilers and API groups.

:one: The **first step**, implemented in this PR, is to choose a generic interface name that could apply to any TriggerMesh component. I picked `Reconcilable` because it's exactly what this interface exists for (thoughts, anyone?).

Additionally, I renamed `src` and `rtr` variables under `.../reconciler/common/` to `rcl` in order to make the merge easier in the future.

---

:two: The second step will be to create such interface in other API groups as well: targets, flow, extensions.

:three: The third step will be to merge interfaces by keeping only the methods common to all component types, and split the methods that only apply to _some_ types into smaller single-method interfaces.


pkg/apis/routing/v1alpha1/interfaces.go
```diff
@@ -26,50 +26,50 @@ import (
 	"knative.dev/pkg/kmeta"
 )
 
-// Router is implemented by all Routing types.
-type Router interface {
+// TMComponent is implemented by all TriggerMesh components.
+type TMComponent interface {
```
antoineco, line 30: After longer reflection, I think I like `Reconcilable` a lot more than `TMComponent`, because it's exactly what this interface exists for.

Feel free to use this thread to voice your opinions.

tzununbekov: +1 on `Reconcilable`

antoineco: Renamed to `Reconcilable` 👍

tzununbekov (approved):

PR #680, opened 2022-03-23
AWS Targets - Add check for address

FranBarrera:
If Address is nil it will fail with a nil pointer:
https://github.com/triggermesh/triggermesh/blob/main/pkg/apis/targets/v1alpha1/aws_lifecycle.go#L82
https://github.com/triggermesh/triggermesh/blob/main/pkg/apis/targets/v1alpha1/aws_lifecycle.go#L90


antoineco (approved):

JeffNeff:
nice catch! and thank you for this!

PR #681, opened 2022-03-23
IBM MQ target events processing metrics

tzununbekov:
Common events processing metrics exported in IBM MQ target.
Missing `schema.GroupResource` variables "generated" for targets.

### DIFF
--- a/pkg/targets/adapter/ibmmqtarget/adapter.go
+++ b/pkg/targets/adapter/ibmmqtarget/adapter.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"time"
 
 	"go.uber.org/zap"
 
@@ -29,7 +30,9 @@ import (
 	pkgadapter "knative.dev/eventing/pkg/adapter/v2"
 	"knative.dev/pkg/logging"
 
+	"github.com/triggermesh/triggermesh/pkg/apis/targets"
 	"github.com/triggermesh/triggermesh/pkg/apis/targets/v1alpha1"
+	"github.com/triggermesh/triggermesh/pkg/metrics"
 	targetce "github.com/triggermesh/triggermesh/pkg/targets/adapter/cloudevents"
 	"github.com/triggermesh/triggermesh/pkg/targets/adapter/ibmmqtarget/mq"
 )
@@ -42,12 +45,16 @@ type ibmmqtargetAdapter struct {
 	logger   *zap.SugaredLogger
 	mqEnvs   *TargetEnvAccessor
 	queue    *mq.Object
+
+	sr *metrics.EventProcessingStatsReporter
 }
 
 // NewAdapter adapter implementation
 func NewAdapter(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClient cloudevents.Client) pkgadapter.Adapter {
 	env := envAcc.(*TargetEnvAccessor)
 	logger := logging.FromContext(ctx)
+
+	metrics.MustRegisterEventProcessingStatsView()
 
 	replier, err := targetce.New(env.Component, logger.Named("replier"),
 		targetce.ReplierWithStatefulHeaders(env.BridgeIdentifier),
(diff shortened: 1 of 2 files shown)


pkg/targets/adapter/ibmmqtarget/adapter.go
```diff
@@ -86,13 +102,22 @@ func (a *ibmmqtargetAdapter) Start(ctx context.Context) error {
...
+	}()
+
 	var msg []byte
 
 	if a.mqEnvs.DiscardCEContext {
 		msg = event.Data()
 	} else {
 		jsonEvent, err := json.Marshal(event)
 		if err != nil {
+			a.sr.ReportProcessingError(false, ceTypeTag, ceSrcTag)
```
antoineco, line 120: Shouldn't `userManaged` (first function parameter) be `true` if we're using the replier?

cc. @odacremolbap

pkg/targets/adapter/ibmmqtarget/adapter.go
```diff
@@ -107,8 +132,10 @@ func (a *ibmmqtargetAdapter) dispatch(ctx context.Context, event cloudevents.Eve
 	}
 
 	if err := a.queue.Put(msg, correlationID); err != nil {
+		a.sr.ReportProcessingError(false, ceTypeTag, ceSrcTag)
```
antoineco, line 135: Same here.

pkg/targets/adapter/ibmmqtarget/adapter.go
tzununbekov: I'm not completely sure what `userManaged` is stands for, so if you say so, I can make it `true`

antoineco: It means that the replier will respond with "200" and an error payload. This bypasses Knative's delivery mechanism, and therefore the error is considered to be "user-managed".

But in this case, I would like a confirmation from @odacremolbap that the replier is actually doing what I described.

odacremolbap: @antoineco is right.

If you want the event to be re-delivered by Kn you can use a different replier method (that no one has used so far), but if you don't want the sender to re-send the event, and in this case I don't think you want that because the same payload will fail again when marshaling, you use the `a.replier.Error` method that you are using.

That method returns 200, meaning ACK, message received so that it is not re-delivered again.
But it also returns a payload to sender containing an error structure and an attribute that metrics can use to mark as `user-managed` error.

antoineco (approved):

antoineco (changes requested):
I approved too fast.
The view isn't registered with `MustRegister...` so metrics aren't being reported.

https://github.com/triggermesh/triggermesh/blob/9b14b87d657ba212aa6e4756d555d4c89044ea23/pkg/flow/adapter/transformation/handler.go#L79

pkg/targets/adapter/ibmmqtarget/adapter.go
tzununbekov: @odacremolbap wasn't this behavior configurable through the env variable? I see the point in preventing retries on error but in some cases, I'd like to be able to utilize Knative's dead letter sink option (if I cannot change sender logic, for example).

pkg/targets/adapter/ibmmqtarget/adapter.go
```diff
@@ -42,6 +45,9 @@ type ibmmqtargetAdapter struct {
 	logger   *zap.SugaredLogger
 	mqEnvs   *TargetEnvAccessor
 	queue    *mq.Object
+
+	mt *pkgadapter.MetricTag
```
antoineco, line 49: It is not necessary to embed this struct because we don't reuse it inside `Start()`.

antoineco (approved):

pkg/targets/adapter/ibmmqtarget/adapter.go
odacremolbap: If you use Knative's features you get the full pack retries + DLS.
If you want to rely on Knative's you can use replier's [ErrorKnativeManaged](https://github.com/triggermesh/triggermesh/blob/d80e46844aff67ab9c927b1cddfcf59848127092/pkg/targets/adapter/cloudevents/replier.go#L114) method.

Replier has no configuration via environment.

PR #684, opened 2022-03-25
Function's Kn service name prefix added

tzununbekov:


pkg/function/function.go
```diff
@@ -164,7 +164,7 @@ func (r *Reconciler) resolveSink(ctx context.Context, f *v1alpha1.Function) (*ap
 func (r *Reconciler) reconcileConfigmap(ctx context.Context, f *v1alpha1.Function) (*corev1.ConfigMap, error) {
 	logger := logging.FromContext(ctx)
 
-	expectedCm := resources.NewConfigmap(f.Name+"-"+rand.String(6), f.Namespace,
+	expectedCm := resources.NewConfigmap(fmt.Sprintf("function-%s-%s", f.Name, rand.String(6)), f.Namespace,
```
antoineco, line 167: The ConfigMap isn't the issue, the Ksvc is.
Kubernetes will happily accept a name starting with a digit for ConfigMaps, but Knative will not because that name translates to a DNS name eventually.

antoineco (changes requested):

pkg/function/function.go
tzununbekov: Missed the line 😒

antoineco (approved):

PR #704, opened 2022-03-31
Merge generic adapter reconcilers

antoineco:
Third and final part of the grand plan mentioned in #678.

Merges all generic adapter reconcilers into one package under `pkg/reconciler`.

The main components involved are:

- 1 `Reconcilable` interface implemented by all components unconditionally
- 3 sub-interfaces implemented by _some_ components based on their nature
    - `EventReceiver` (components that explicitly declare the event types they can receive)
    - `EventSender` (components that can _send_ to a sink using a CloudEvents client)
    - `EventSource` (components that declare the event types they can emit/respond, but do not necessarily use a sink)
- 2 flavors of generic reconcilers: Deployment and Kn Service
  - Can reconcile different aspects of adapters and components' statuses based on the interfaces above (CE attributes, sink URL, status conditions, ...)

Everything under `pkg/reconciler` and `pkg/apis/common` is "new" (merged from what used to be `pkg/{sources,targets,flow,routing}/reconciler/common` and `pkg/apis/{sources,targets,flow,routing}`).
The rest only contains updates to Go imports.


antoineco:
No matter the outcome of this PR, Git's stats make me happy:

```diff
+3,516
-15,307
```

tzununbekov (approved):

PR #706, opened 2022-04-03
Install required C libs in CodeQL workflow

antoineco:
The CodeQL GitHub Actions workflow is currently failing due to missing C libraries on the CI runner: https://github.com/triggermesh/triggermesh/actions/runs/2083675403

### DIFF
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -37,17 +37,21 @@ jobs:
         restore-keys: |
           ${{ github.job }}-${{ runner.os }}-go-
 
-    - name: Initialize CodeQL
-      uses: github/codeql-action/init@v1
-      with:
-        languages: go
-
-    # Install IBM MQ redis client required by IBM MQ source and target adapters
+    # Install IBM MQ client required by IBM MQ source and target adapters
     - name: Install IBM MQ client
       run: |
         curl https://public.dhe.ibm.com/ibmdl/export/pub/software/websphere/messaging/mqdev/redist/9.2.3.0-IBM-MQC-Redist-LinuxX64.tar.gz -o mq.tar.gz
         mkdir -p /opt/mqm
         tar -C /opt/mqm -xzf mq.tar.gz
+
+    # Install C libraries required by the XSLT transformation adapter
+    - name: Install C libraries for XSLT transformation
+      run: apt-get install -y --no-install-recommends libxml2-dev libxslt1-dev liblzma-dev zlib1g-dev
+
+    - name: Initialize CodeQL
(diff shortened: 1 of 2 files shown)


cmd/xslttransformation-adapter/Dockerfile
```diff
@@ -40,13 +40,13 @@ COPY --from=builder \
...
     /usr/lib/x86_64-linux-gnu/libicuuc.so.67 \
-    /usr/lib/x86_64-linux-gnu/libgcrypt.so.20 \
-    /lib/x86_64-linux-gnu/libgpg-error.so.0 \
     /lib/x86_64-linux-gnu/libz.so.1 \
     /lib/x86_64-linux-gnu/liblzma.so.5 \
+    /usr/lib/x86_64-linux-gnu/libgcrypt.so.20 \
     /usr/lib/x86_64-linux-gnu/libicudata.so.67 \
     /usr/lib/x86_64-linux-gnu/libstdc++.so.6 \
     /lib/x86_64-linux-gnu/libgcc_s.so.1 \
+    /lib/x86_64-linux-gnu/libgpg-error.so.0 \
```
antoineco, line 49: Note: I used the occasion to re-order some of these lines so that they match the output of `ldd /xslttransformation-adapter` (see few lines above in the same file). Easier see at a glance whether we forgot to copy something.

.github/workflows/codeql.yaml
```diff
@@ -37,18 +37,22 @@ jobs:
...
       run: |
         curl https://public.dhe.ibm.com/ibmdl/export/pub/software/websphere/messaging/mqdev/redist/9.2.3.0-IBM-MQC-Redist-LinuxX64.tar.gz -o mq.tar.gz
         mkdir -p /opt/mqm
         tar -C /opt/mqm -xzf mq.tar.gz
 
+    # Install C libraries required by the XSLT transformation adapter
+    - name: Install C libraries for XSLT transformation
+      run: apt-get install -y --no-install-recommends libxml2-dev libxslt1-dev liblzma-dev zlib1g-dev
+
+    - name: Initialize CodeQL
```
cab105, line 51: Surprised by the ordering. Would've thought the missing MQ components would have triggered a failure as well.

cab105 (approved):
Nice quick fix.

.github/workflows/codeql.yaml
antoineco: Apparently the "initialize" step only prepares the environment somehow, it doesn't yet build the code so that's why it works. Still, I think it's more logical to keep CodeQL related steps together.

PR #713, opened 2022-04-06
Make source adapters backed by a Ksvc public

### LINKED ISSUE #712: E2E test for WebhookSource is flaky, fails to POST HTTP request
The test occasionally fails with the following error:

```
Error Posting to http://webhooksource-test-b5jts.e2e-webhooksource-5498.svc.cluster.local: timed out waiting for the condition
```

Relevant code:

https://github.com/triggermesh/triggermesh/blob/46464eca0ae139fb424fe4b4605e5969603bc281/test/e2e/sources/webhook/main.go#L104-L107

This _could_ be caused by the lack of readiness check, although in this case 1 minute should be more than enough for the adapter to serve the endpoint.

As an example, the source for Amazon SNS exposes a `/health` endpoint that is configured by the reconciler as a readiness probe:

https://github.com/triggermesh/triggermesh/blob/46464eca0ae139fb424fe4b4605e5969603bc281/pkg/sources/adapter/awssnssource/adapter.go#L87

https://github.com/triggermesh/triggermesh/blob/46464eca0ae139fb424fe4b4605e5969603bc281/pkg/sources/reconciler/awssnssource/adapter.go#L52-L53

antoineco:
Fixes #712

#704 introduced a change that makes Knative Services private by default:

https://github.com/triggermesh/triggermesh/blob/46464eca0ae139fb424fe4b4605e5969603bc281/pkg/reconciler/adapter.go#L200

In 95% of the cases this is what we want, but there are a few exceptions that require setting the visibility of the Ksvc to _public_, namely in components that expose a public webhook:

- Webhook source
- Slack source
- Amazon SNS source
- Zendesk source
- Twilio source

### DIFF
--- a/pkg/sources/reconciler/awssnssource/adapter.go
+++ b/pkg/sources/reconciler/awssnssource/adapter.go
@@ -22,6 +22,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 
 	"knative.dev/eventing/pkg/reconciler/source"
+	network "knative.dev/networking/pkg"
 	"knative.dev/pkg/apis"
 	"knative.dev/pkg/kmeta"
 	servingv1 "knative.dev/serving/pkg/apis/serving/v1"
@@ -47,6 +48,7 @@ var _ common.AdapterServiceBuilder = (*Reconciler)(nil)
 func (r *Reconciler) BuildAdapter(src commonv1alpha1.Reconcilable, _ *apis.URL) *servingv1.Service {
 	return common.NewMTAdapterKnService(src,
 		resource.Image(r.adapterCfg.Image),
+		resource.Label(network.VisibilityLabelKey, "public"),
 		resource.EnvVars(r.adapterCfg.configs.ToEnvVars()...),
 
 		// NOTE(antoineco): startupProbe isn't yet supported as of Knative 1.2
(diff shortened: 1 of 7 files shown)


odacremolbap (approved):

pkg/sources/reconciler/awssnssource/adapter.go
```diff
@@ -47,6 +48,7 @@ var _ common.AdapterServiceBuilder = (*Reconciler)(nil)
 func (r *Reconciler) BuildAdapter(src commonv1alpha1.Reconcilable, _ *apis.URL) *servingv1.Service {
 	return common.NewMTAdapterKnService(src,
 		resource.Image(r.adapterCfg.Image),
+		resource.Label(network.VisibilityLabelKey, "public"),
```
antoineco, line 51: I'm making the `public` value up here, it doesn't actually exist.
Knative only checks whether that value differs from `cluster-local`, in which case it always falls back to public.

The reason why we need to set _some_ value is because `common.NewMTAdapterKnService()` sets this label to `cluster-local`, which we need to override.

I considered two alternative approaches to this problem:

- changing the helper to _remove_ the label when we set its value to `""` (or even adding a `RemoveLabel()` helper)
    ... but I find this hacky and unintuitive
- removing the line that sets the label to `cluster-local` by default
    ... but then we have to copy that same line to 40 targets and transformers (at the time of writing)

With this approach, only 5 components need to be changed, and setting the value to `public` makes it very explicit that the service is going to be served over a public URL, which IMO is a good thing in terms of readability.

I'm open to comments and suggestions.

antoineco (commented):
Notes to reviewers.

sameersbn (approved):

antoineco:
Closing and reopening because CircleCI missed the PR due to an outage.

PR #717, opened 2022-04-06
Exclude non-API packages from docs generation

### LINKED ISSUE #151: Warning during gen-docs execution
Since the project is using Go 1.17, we should be installing cmds using `go install <cmd>@version` instead of `go get <cmd>`:

```
go get: installing executables with 'go get' in module mode is deprecated.
	To adjust and download dependencies of the current module, use 'go get -d'.
	To install using requirements of the current module, use 'go install'.
	To install ignoring the current module, use 'go install' with a version,
	like 'go install example.com/cmd@latest'.
	For more information, see https://golang.org/doc/go-get-install-deprecation
	or run 'go help get' or 'go help install'.
```

See also #135

### LINKED ISSUE #716: `gen-apidocs` tries to find API objects inside `pkg/apis/common` and fails
```
F0406 15:54:28.810514     536 main.go:135] no API packages found in ./pkg/apis/common
```

We should probably exclude this directory because, similarly to `pkg/apis/duck` in Knative, it doesn't contain any real Kubernetes resources.
We need to ensure that shared structs such as [`ValueFromField`](https://github.com/triggermesh/triggermesh/blob/8d1d841a7567a91fc451398c7a180003de7a19a9/pkg/apis/common/v1alpha1/types.go#L21-L34) and [`Status`](https://github.com/triggermesh/triggermesh/blob/8d1d841a7567a91fc451398c7a180003de7a19a9/pkg/apis/common/v1alpha1/status_conditions.go#L84-L93) are correctly referenced though.

antoineco:
Fixes #151
Fixes #716

`pkg/apis/common` is similar to `pkg/apis/duck` in Knative: it doesn't contain real API resources.
Knative generates docs for those types anyway ([example](https://knative.dev/docs/reference/api/eventing-api/#duck.knative.dev%2fv1)), which has the advantage of embedding their internal types seamlessly inside the Knative docs ([example](https://knative.dev/docs/reference/api/eventing-api/#duck.knative.dev/v1.ChannelableSpec)).
However, I have a feeling that this would require extra work on our side since we split our docs into one page per _API group_, as opposed to one page per project like Knative does (and `common` would be treated as its own API group).

Instead, I added rules that generate valid external links to https://pkg.go.dev for all types inside `pkg/apis/common` and its sub-packages ([example](https://pkg.go.dev/github.com/triggermesh/triggermesh/pkg/apis#ARN)). This fixes the few un-linked types which we currently have in our docs:

image

Last but not least, I updated the script to install the `ahmetb/gen-crd-api-reference-docs` command via `go install`
instead of `go get`. The latter was deprecated in Go 1.17 for installing executables (see #135).

### DIFF
--- a/hack/reference-docs-gen-config.json
+++ b/hack/reference-docs-gen-config.json
@@ -9,11 +9,11 @@
     "externalPackages": [
         {
             "typeMatchPrefix": "^k8s\\.io/apimachinery/pkg/apis/meta/v1\\.Duration$",
-            "docsURLTemplate": "https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#Duration"
+            "docsURLTemplate": "https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Duration"
         },
         {
             "typeMatchPrefix": "^k8s\\.io/apimachinery/pkg/types.UID$",
-            "docsURLTemplate": "https://godoc.org/k8s.io/apimachinery/pkg/types#UID"
+            "docsURLTemplate": "https://pkg.go.dev/k8s.io/apimachinery/pkg/types#UID"
         },
         {
             "typeMatchPrefix": "^k8s\\.io/(api|apimachinery/pkg/apis)/",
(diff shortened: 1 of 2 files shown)


hack/reference-docs-gen-config.json
```diff
@@ -35,6 +35,14 @@
             "typeMatchPrefix": "^knative\\.dev/networking/pkg/apis/networking",
             "docsURLTemplate": "https://pkg.go.dev/knative.dev/networking/pkg/apis/networking#{{.TypeIdentifier}}"
         },
+        {
+            "typeMatchPrefix": "^github\\.com/triggermesh/triggermesh/pkg/apis/common",
+            "docsURLTemplate": "https://pkg.go.dev/github.com/triggermesh/triggermesh/pkg/apis/common/{{arrIndex .PackageSegments -1}}#{{.TypeIdentifier}}"
+        },
```
antoineco, line 41: This rule generates links such as:

```html
<code>certKey</code></br>
<em>
<a href="https://pkg.go.dev/github.com/triggermesh/triggermesh/pkg/apis/common/v1alpha1#ValueFromField">
github.com/triggermesh/triggermesh/pkg/apis/common/v1alpha1.ValueFromField
</a>
</em>
```

(The link is currently invalid because the latest TriggerMesh release doesn't contain `pkg/apis/common`. We'll need to wait until the next release before the link becomes valid.)

hack/reference-docs-gen-config.json
```diff
@@ -35,6 +35,14 @@
...
+        {
+            "typeMatchPrefix": "^github\\.com/triggermesh/triggermesh/pkg/apis",
+            "docsURLTemplate": "https://pkg.go.dev/github.com/triggermesh/triggermesh/pkg/apis#{{.TypeIdentifier}}"
+        },
```
antoineco, line 45: This rule generates links such as:

```html
<code>arn</code></br>
<em>
<a href="https://pkg.go.dev/github.com/triggermesh/triggermesh/pkg/apis#ARN">
github.com/triggermesh/triggermesh/pkg/apis.ARN
</a>
</em>
```

antoineco (commented):
Notes

PR #723, opened 2022-04-08
Add kodata to dockerfiles

### LINKED ISSUE #701: Container images built from Dockerfile do no include KODATA
Knative's logging implementation looks up for information about the Git revision of the compiled code inside the KODATA directory. If it finds it, that information is used to augment the logger with the commit hash, which can be useful while analyzing logs, e.g. to determine whether an error started appearing more frequently after a certain code revision.

Some of our images aren't built using `ko`, therefore that information is missing from those components.

In the past, before migrating our release pipeline from Docker to `ko`, we used to support that feature as follows (simplified version):

```dockerfile
FROM golang:1.17-bullseye as builder

COPY . .

RUN mkdir /kodata && mv .git/* /kodata/

FROM gcr.io/distroless/base-debian11:nonroot

# Emulate ko builds
# https://github.com/google/ko/blob/v0.5.0/README.md#including-static-assets
ENV KO_DATA_PATH /kodata

COPY --from=builder /kodata/ ${KO_DATA_PATH}/
```

cab105:
Update the remaining dockerfiles to not only include the missing copyright, but also ensure the `/kodata` directory is available and populated to ensure knative logging captures the commit ID.

Additionally, true up the workdir for builds to have a common location to facilitate the bringing over of the git data.

Closes #701

### DIFF
--- a/cmd/dataweavetransformation-adapter/Dockerfile
+++ b/cmd/dataweavetransformation-adapter/Dockerfile
@@ -1,10 +1,24 @@
+# Copyright 2022 TriggerMesh Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 FROM golang:1.17-bullseye as builder
 
 RUN set -eux; \
     apt-get update; \
     apt-get install -y --no-install-recommends unzip
 
-WORKDIR ${GOPATH}/project
+WORKDIR /go/triggermesh
 ENV DW_VERSION="1.0.19"
 
 RUN  curl -sSLO https://github.com/mulesoft-labs/data-weave-cli/releases/download/v$DW_VERSION/dw-$DW_VERSION-Linux && \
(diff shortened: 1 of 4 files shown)


antoineco:
Noice!

Before this PR:

```json
{
  "severity": "INFO",
  "timestamp": "2022-04-08T22:06:59.1998709Z",
  "logger": "xslttransformation-adapter",
  "caller": "xslttransformation/adapter.go:92",
  "message": "Starting XSLT transformer"
}
```

After this PR:

```json
{
  "severity": "INFO",
  "timestamp": "2022-04-08T22:08:54.1403952Z",
  "logger": "xslttransformation-adapter",
  "caller": "xslttransformation/adapter.go:92",
  "message": "Starting XSLT transformer",
  "commit": "8f8b756"
}
```

The commit hash is attached to the logger as expected 👍

antoineco (approved):

sameersbn (approved):

cmd/dataweavetransformation-adapter/Dockerfile
```diff
@@ -14,10 +28,14 @@ RUN go build -o /dataweavetransformation-adapter ./cmd/dataweavetransformation-a
 
 FROM debian:stable-slim
 
+# Ensure the /kodata entries are present
+COPY --from=builder /go/triggermesh/.git/ /kodata/
```
sameersbn, line 32: Do we really need the contents of the entire `.git` directory? In the example at https://github.com/google/ko#static-assets I see they only the symlink `.git/HEAD` file, so will just copying over that one file suffice?

antoineco: Only a few selected files are copied in practice, due to this exclusion list:

https://github.com/triggermesh/triggermesh/blob/ba3b7ab1642070cc6a6e804306faf7463e7738f7/.dockerignore#L19

sameersbn: ah ha! thanks for the clarification

PR #724, opened 2022-04-08
Introduce TriggerMesh aggregation roles

### LINKED ISSUE #715: Add our ClusterRoles to our release
Currently we do not ship our internal ClusteRoles as part of the release.

We should bundle them so that a user can aggregate our roles with their own ClusterRoles.

We do have:
- triggermesh-admin
- triggermesh-editor
- triggermesh-viewer

cab105:
Migrated the aggregated cluster roles used internally by TriggerMesh to provide a starting point for others wanting to add finer grained access controls to their sources or targets.

Closes triggermesh/triggermesh#715

### DIFF
--- a/config/200-aggregation-admin-clusterroles.yaml
+++ b/config/200-aggregation-admin-clusterroles.yaml
@@ -1,0 +1,171 @@
+# Copyright 2022 TriggerMesh Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# All sources used at TriggerMesh
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: triggermesh-sources-admin
+  labels:
+    rbac.triggermesh.io/sources-admin: "true"
+rules:
+- apiGroups:
+  - sources.knative.dev
+  - sources.eventing.knative.dev
+  - sources.eventing.triggermesh.dev
+  - sources.triggermesh.io
+  resources: ["*"]
+  verbs:
+  - get
+  - list
+  - create
+  - update
+  - patch
+  - delete
+  - deletecollection
+  - watch
+
+---
+
+# All targets used at TriggerMesh
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: triggermesh-targets-admin
+  labels:
+    rbac.triggermesh.io/targets-admin: "true"
+rules:
+- apiGroups:
+  - targets.triggermesh.io
+  resources: ["*"]
+  verbs:
+  - get
+  - list
+  - create
+  - update
+  - patch
+  - delete
+  - watch
+
+---
+
+# All Kantive eventing objects used at TriggerMesh
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: triggermesh-knative-admin
+  labels:
+    rbac.triggermesh.io/knative-admin: "true"
+rules:
+- apiGroups:
+  - serving.knative.dev
+  - eventing.knative.dev
+  - messaging.knative.dev
+  - messaging.triggermesh.dev
+  resources:
+  - brokers
+  - channels
+  - configurations
+  - revisions
+  - routes
+  - triggers
+  - subscriptions
+  - services
+  - kinesischannels
+  - inmemorychannels
+  - clusterchannelprovisioners
+  verbs:
+  - get
+  - list
+  - create
+  - update
+  - patch
+  - delete
+  - deletecollection
+  - watch
+
+---
+
+# Kubernetes objects for TriggerMesh users
+# NOTE: dupe at config repo, label changed to
+#       match other labels pattern at ClusterRoles.
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: triggermesh-kubernetes-admin
+  labels:
+    rbac.triggermesh.io/kubernetes-admin: "true"
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  - pods/log
+  - pods/exec
+  - services
+  - serviceaccounts
+  - secrets
+  - configmaps
+  verbs:
+  - get
+  - list
+  - create
+  - update
+  - patch
+  - delete
+  - deletecollection
+  - watch
+- apiGroups:
+  - apps
+  resources:
+  - deployments
+  verbs:
+  - get
+  - list
+  - create
+  - update
+  - patch
+  - delete
+  - deletecollection
+  - watch
+
+---
+
+# TriggerMesh bridges object
+# NOTE: dupe at config repo, label name
+#       changed to match other labels pattern.
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: triggermesh-flow-admin
+  labels:
+    rbac.triggermesh.io/flow-admin: "true"
+rules:
+- apiGroups:
+  - flow.triggermesh.io
+  resources: ["*"]
+  verbs:
+  - get
+  - list
+  - create
+  - update
+  - patch
+  - delete
+  - watch
+
--- a/config/200-aggregation-admin-clusterroles.yaml
+++ b/config/200-aggregation-admin-clusterroles.yaml
@@ -168,4 +168,3 @@ rules:
   - patch
   - delete
   - watch
-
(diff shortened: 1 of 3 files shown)


config/200-aggregation-admin-clusterroles.yaml
```diff
@@ -0,0 +1,170 @@
...
+# limitations under the License.
+
+# All sources used at TriggerMesh
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: triggermesh-sources-admin
+  labels:
+    rbac.triggermesh.io/sources-admin: "true"
```
antoineco, line 22: What is the purpose of these labels?

config/200-aggregation-admin-clusterroles.yaml
```diff
@@ -0,0 +1,170 @@
...
+rules:
+- apiGroups:
+  - sources.knative.dev
+  - sources.eventing.knative.dev
+  - sources.eventing.triggermesh.dev
```
antoineco, line 27: I don't think we have such group in tm/tm.

config/200-aggregation-admin-clusterroles.yaml
```diff
@@ -0,0 +1,170 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: triggermesh-sources-admin
+  labels:
+    rbac.triggermesh.io/sources-admin: "true"
+rules:
+- apiGroups:
+  - sources.knative.dev
+  - sources.eventing.knative.dev
```
antoineco, line 26: To my knowledge this API group does not exist.

config/200-aggregation-admin-clusterroles.yaml
```diff
@@ -0,0 +1,170 @@
...
+  - list
+  - create
+  - update
+  - patch
+  - delete
+  - watch
+
+---
+
+# All Kantive eventing objects used at TriggerMesh
```
antoineco, line 65: Knative (typo)

config/200-aggregation-admin-clusterroles.yaml
```diff
@@ -0,0 +1,170 @@
...
+  - update
+  - patch
+  - delete
+  - deletecollection
+  - watch
+
+---
+
+# Kubernetes objects for TriggerMesh users
+# NOTE: dupe at config repo, label changed to
```
antoineco, line 104: The referenced repo is private so the comment may introduce confusion.

config/200-aggregation-admin-clusterroles.yaml
```diff
@@ -0,0 +1,170 @@
...
+  resources:
+  - brokers
+  - channels
+  - configurations
+  - revisions
+  - routes
+  - triggers
+  - subscriptions
+  - services
+  - kinesischannels
```
antoineco, line 88: To me it's questionable whether we should include types which aren't either part of Knative or tm/tm. The list could grow very long and might not remain manageable over time.

config/200-aggregation-admin-clusterroles.yaml
```diff
@@ -0,0 +1,170 @@
...
+  - inmemorychannels
+  - clusterchannelprovisioners
```
antoineco, line 90: For my own curiosity, what is this?

config/200-aggregation-viewer-clusterroles.yaml
```diff
@@ -0,0 +1,199 @@
...
+  - extensions.triggermesh.io
+  resources: ["*"]
+  verbs:
+  - get
+  - list
+  - watch
+
+---
+
+# All Knative eventing objects used at TriggerMesh
```
antoineco, line 75: The comment doesn't match, the role contains more than just eventing resources.

config/200-aggregation-viewer-clusterroles.yaml
```diff
@@ -0,0 +1,199 @@
...
+  name: triggermesh-kubernetes-viewer
+  labels:
+    rbac.triggermesh.io/kubernetes-viewer: "true"
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  - pods/log
+  - services
```
antoineco, line 146: If you can view the Pod you can view all its subresources.
Creating a rule for subresources is useful only if you want to e.g. prevent write access to the resource but allow it for selected subresources, like the status.

config/200-aggregation-admin-clusterroles.yaml
```diff
@@ -0,0 +1,170 @@
...
+  name: triggermesh-kubernetes-admin
+  labels:
+    rbac.triggermesh.io/kubernetes-admin: "true"
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  - pods/log
+  - pods/exec
```
antoineco, line 119: If you can view/edit the Pod you can view/edit all its subresources.
Creating a rule for subresources is useful only if you want to e.g. prevent write access to the resource but allow it for selected subresources, like the status.

antoineco (commented):
Very quick initial review.

We could do a better job at splitting resources per group (e.g. there won't be "services" in "eventing.knative.dev", yet everything is grouped together), but this can be refined later.

antoineco:
General question: what is the fundamental difference between "admin" and "editor"?

cab105:
> General question: what is the fundamental difference between "admin" and "editor"?

As mentioned in the comments for 200-aggregation-editor-clusterroles.yaml (https://github.com/triggermesh/triggermesh/blob/c8a5b0d4bc0304e3f0a150de2938847ce7dfd59f/config/200-aggregation-editor-clusterroles.yaml#L15)
```
# Editors have same permissions as Admins on Kubernetes objects but
# cannot manage Secrets
```

antoineco:
Wouldn't it make sense to remove "admin" entirely and create a role just for secrets management then?
It's easier to manage and equally easy to aggregate.

config/200-aggregation-admin-clusterroles.yaml
cab105: It doesn't anymore, but did back in the 0.13.x days: https://github.com/knative/docs/blob/v0.13.x/docs/reference/eventing/eventing.md

config/200-aggregation-admin-clusterroles.yaml
cab105: Hmmm thought we did for the early sources work. Needless to say, there's quite a bit of truing up that needs to happen internally.

config/200-aggregation-admin-clusterroles.yaml
cab105: The labels were used for higher level aggregation with another cluster role within the TriggerMesh ecosystem.

config/200-aggregation-admin-clusterroles.yaml
cab105: You're right. Can scale this back to focus only on what is provided out of the box for knative and tm/tm while leaving it as an exercise to the reader to provide the updates for their own resources. Would that work?

config/200-aggregation-admin-clusterroles.yaml
cab105: Deprecated. :-D https://github.com/knative/eventing/issues/1627

config/200-aggregation-admin-clusterroles.yaml
cab105: Can you? Thought you could only view top level resources, and you had to explicitly set each sub resource to fall into line.

config/200-aggregation-admin-clusterroles.yaml
cab105: Will reach out of band on this.

cab105:
Thank you for the feedback on this. Will switch this to draft mode to true the other components first, and then request the re-review.

config/200-aggregation-admin-clusterroles.yaml
antoineco: I think so.
I didn't scrutinize all roles but this is the only resource which didn't seem directly related to either Knative or tm/tm.

config/200-aggregation-admin-clusterroles.yaml
antoineco: If that was the case, `kubectl get pod` wouldn't return the `status`, for example.
Giving a permission to the top-level resource automatically propagates that permission to its subresources.

config/200-aggregation-admin-clusterroles.yaml
cab105: For the record, it is/was only used by an internal source that has since been archived.

sebgoa:
There are two concerns for me with this PR:

1. Do we need the roles as part of the release artefacts or documented somewhere else.

-     I would tend to go with the way knative does it and add our roles in our manifests so that the `kubectl apply` installation command creates them

2. We need to clean up the roles
- Right now there is a `triggermesh-sources-admin` and a `triggermesh-targets-admin`, I think they should be merged in favor of a single `triggermesh-admin`
- We should remove any knative reference
- We should remove any non-triggermesh/triggermesh reference

antoineco:
100% agree with removing any API group that isn't a TriggerMesh API group from these roles.
We should think about `triggermesh-admin` and `triggermesh-reader` as roles which are combinable with other roles to create a larger set of permissions, and not as _the_ final set of permissions.

Every cluster administrator will have their own idea about what a certain persona can or can not do, and it's not our call whether these users should be allowed to manipulate Kubernetes objects such as Pods, etc. let alone Knative objects.

In practice it is quite simple: only permissions to interact with TriggerMesh objects are required to be able to use TriggerMesh.
All other objects (pods, deployments, ksvcs, secrets, ...) are either managed or consumed by `triggermesh-controller`, not by users directly.

Even troubleshooting _is_ possible to a certain extent without having to access lower level objects. Reconciliation errors are surfaced in the status of TriggerMesh objects.
Of course objects' statuses aren't always enough, and in certain cases it might help to be able to access things like logs, Kubernetes events, etc. but IMO this is something a _cluster administrator_ should decide.

sebgoa:
closing in favor of #937

PR #729, opened 2022-04-11
Migrate Function to `Reconcilable` interface

### LINKED ISSUE #726: Implement Reconcilable in Function kind
`Function` is the last type that is not using the generic adapter reconciler (ref. #678 #682 #686).

antoineco:
Closes #726

Along the way I

- Re-used the `adapter.go` `controller.go` `reconciler.go` structure used in other components
- Added the missing `resource` helpers to `pkg/reconciler/resource` (`Entrypoint` and `ConfigMapMount`)
- Refactored the reconciliation logic of the ConfigMap for readability

---

Tested against the repo's samples. Here is an example:

```sh
curl -D- http://function-inline-node-function.default.svc.cluster.local \
  -H "Ce-Specversion: 1.0" \
  -H "Ce-Id: 0001" \
  -H "Ce-Type: greeting" \
  -H "Ce-Source: curl" \
  -H "Content-Type: application/json" \
  -d '{"msg":"Hello, TriggerMesh!"}'
```

Response headers:

```
HTTP/1.1 200 OK
content-length: 279
content-type: application/cloudevents+json
date: Mon, 11 Apr 2022 18:01:16 GMT
x-envoy-upstream-service-time: 1229
server: envoy
```

Response body (pretty-printed):

```json
{
  "specversion": "1.0",
  "id": "b0fe1ad0-0cda-48f8-929e-7e700ad4a08c",
  "source": "io.triggermesh.function.default.inline-node-function",
  "type": "io.triggermesh.nodejs.sample",
  "datacontenttype": "application/json",
  "time": "2022-04-11T18:01:16.0811608Z",
  "data": {
    "msg": "Hello, TriggerMesh!"
  }
}
```


tzununbekov:
It was not introduced by this PR, but it may be worth adding a fix:
switching function's `responseIsEvent` parameter from `true` to `false` does not trigger adapter service reconciliation. The problem is that when this parameter is `false` adapter's corresponding env variable is set to an empty string that's probably ignored by the deep derivative comparison (https://github.com/triggermesh/triggermesh/pull/729/files#diff-71020045dbc838393e309dfc1e91f32b5e93db449fa8b1ba94acb392d412394bR89-R92).
Feel free to skip this comment if it is not in the refactoring scope.

antoineco:
That could be a result of the deep derivative comparison, I think you're right.

I'll fix it in this PR 👍

PR #743, opened 2022-04-12
Custom DeepDerivative comparison of EnvVar to allow empty Value

### LINKED ISSUE #741: DeepDerivate semantic comparison ignores EnvVar with emtpy values
Consider two Containers with the following env vars:

```yaml
# desired
env:
- name: FOO
  value: ''
- name: BAR
  value: something
```

```yaml
# current
env:
- name: FOO
  value: something
- name: BAR
  value: something
```

Because the DeepDerivative semantic comparison compares fields _one by one_ instead of considering the `EnvVar` as a whole, those two containers are wrongfully considered equal.

Here is a test case that reproduces the problem:

```patch
diff --git a/pkg/reconciler/semantic/semantic_test.go b/pkg/reconciler/semantic/semantic_test.go
index a57deae0..0e66eba2 100644
--- a/pkg/reconciler/semantic/semantic_test.go
+++ b/pkg/reconciler/semantic/semantic_test.go
@@ -96,6 +101,14 @@ func TestDeploymentEqual(t *testing.T) {
                        },
                        false,
                },
+               "not equal when EnvVar desired value is empty": {
+                       func() *appsv1.Deployment {
+                               desired := current.DeepCopy()
+                               desired.Spec.Template.Spec.Containers[0].Env[0].Value = ""
+                               return desired
+                       },
+                       false,
+               },
        }
```

antoineco:
Fixes #741

### DIFF
--- a/pkg/reconciler/semantic/semantic.go
+++ b/pkg/reconciler/semantic/semantic.go
@@ -89,6 +89,15 @@ var eq = conversion.EqualitiesOrDie(
 
 		return (conversion.Equalities{}).DeepDerivative(a.Handler, b.Handler)
 	},
+	// Needed because DeepDerivative compares EnvVar.Value string fields
+	// without considering EnvVar as a whole. If an EnvVar is specified, we
+	// consider its value to be intentional and force the comparison.
+	func(a, b corev1.EnvVar) bool {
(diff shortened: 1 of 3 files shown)


tzununbekov (approved):
:shipit:

PR #749, opened 2022-04-12
Report linter errors in both the GitHub Actions workflow logs and the Files Changed tab

antoineco:
I've heard a few times that it was confusing that linter errors were reported to the "Files Changed" tab of PRs, but not to the logs of the GitHub Actions workflow.

With this change, errors (with code location) are also printed to the logs of the GitHub Actions workflow.
To achieve that, we have to run the linter a second time in case of errors, which luckily is relatively fast due to caching.

Effect of caching:

image

Files Changed view:

image

Workflow logs view:

image

### DIFF
--- a/.github/workflows/static.yaml
+++ b/.github/workflows/static.yaml
@@ -25,6 +25,17 @@ jobs:
     # the Go setup, in case other steps are added to this job in the future.
     - name: Lint Go code
       uses: golangci/golangci-lint-action@v3
+      with:
+        # (hack) By default, errors are reported to the GitHub commit view only
+        # (or the "Files changed" tab on PRs). We also want errors to be logged
+        # with line numbers to the execution logs of the workflow.
+        #
+        # The args below result in the following flags being passed to the
+        # linter command, which works, quite surprisingly:
+        #   --out-format=github-actions --out-format=colored-line-number
+        #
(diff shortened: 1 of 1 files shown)


odacremolbap (approved):

.github/workflows/static.yaml
```diff
@@ -24,8 +24,21 @@ jobs:
...
+    # (hack) By default, errors are reported to the "Files changed" tab on PRs.
+    # In case errors were found, run the linter a second time but don't direct
+    # the output to GitHub, so that they are also visible in the execution logs
+    # of the workflow, with line numbers.
+    # Ref. https://github.com/golangci/golangci-lint-action/issues/119#issuecomment-981090648
+    - name: Lint Go code (debug)
+      if: ${{ failure() && steps.lint.outcome == 'failure' }}
+      uses: golangci/golangci-lint-action@v3
+      with:
+        args: "--out-${NO_FUTURE}format colored-line-number"
```
antoineco, line 39: I just made a [funny discovery](https://github.com/golangci/golangci-lint-action/issues/119#issuecomment-1096917139).
This workflow steps actually executes the following command:

```
golangci-lint run --out-format=github-actions --out-format=colored-line-number
```

Turns out this effectively logs errors to both GitHub _and_ the workflow logs 😅 
So we might not even need this second `if: faillure()` step.

Issue #705, opened 2022-04-01
Adapter parameters overriding


tzununbekov:
The idea of introducing the mechanism for low-level adapter parameters overriding is inspired by the following cases:

**Knative service labels:**
- Synchronizer issue. Currently, we cannot utilize Synchronizer functionality without a custom connection link between external systems and Triggermesh components (e.g. Kong gateway with CE plugin or Webhook source with a custom CE client),
- It is happening quite often that we need to have an access to the component's endpoint for direct interaction with the adapter for debugging purposes or to use it as a standalone service.

Setting Knative service network visibility scope label would help resolve these issues.

**Adapter's environment variables:**
- CE Overrides. Context attributes overrides are defined as a part of the sources spec but other components do not seem to have that ability, while there is no reason to limit it to the sources only,
- Delivery configuration. In bridges with multiple targets and different conditional routes, CE responses handling can become a very complex task and there are no chances we can have predefined, hardcoded delivery logic (errors, payloads, dls, etc.) in every component that would suit all possible scenarios.

These parameters are usually configurable through the adapter environment variables.

**Custom resources:**
- Extend adapter limits. Some adapters may require more compute resources for events processing tasks in the high load environments,
- Adding custom resources, such as secret mount or external volume may be something that we want to have as an opportunity in the future.

---

Would it make sense to work on the common solution for the described cases or should we solve the problems as they come, extending every particular component's spec one by one? If we go with the former, I see us implementing "adapter overridable" duck type that exposes meta and spec keys with the implementation depending on the adapter type.

antoineco:
Do you have a first draft to suggest, just to kickstart some discussion?

Would that take the shape of a nested Go struct that's embeddable in all types' spec?

tzununbekov:
I thought of starting with the simple structure to override the adapter's labels and envs. I'm not sure about the best way to expose overridable parameters: a predefined set of `labels`, `envs`, `resources`, etc. or just dump a raw corev1.Deployment spec there.

AWS SQS Source example with labels and envs overrides:

Reconcilable interface:
```diff
diff --git a/pkg/apis/common/v1alpha1/object_interfaces.go b/pkg/apis/common/v1alpha1/object_interfaces.go
index f8dd18e2..5be21a1f 100644
--- a/pkg/apis/common/v1alpha1/object_interfaces.go
+++ b/pkg/apis/common/v1alpha1/object_interfaces.go
@@ -39,6 +39,9 @@ type Reconcilable interface {

[...]

PR #750, opened 2022-04-13
adapterOverrides in Reconcilable interface

tzununbekov:
An initial set of adapter parameter overrides that includes Kn Service visibility scope switch and Resource Requests and Limits. 
Draft until testing is complete.
Discussed in triggermesh/triggermesh#705.


tzununbekov:
Note: maybe we should rename `adapterOverrides` to just `adapter` in CRDs since it does not bring any additional value or clarity for users, i.e.
```
spec:
  adapterOverrides:
    public: true
```
to
```
spec:
  adapter:
    public: true
```

pkg/apis/common/v1alpha1/object_interfaces.go
```diff
@@ -39,6 +39,9 @@ type Reconcilable interface {
 
 	// GetStatusManager returns a manager for the component's status.
 	GetStatusManager() *StatusManager
+
+	// GetAdapterOverrides returns the adapter overrides.
+	GetAdapterOverrides() *AdapterOverrides
```
antoineco, line 44: I was about to suggest that we turn this into its own interface, so that we take a Go-_ish_ behaviour-driven approach and allow components to selectively implement overrides or not.
But I see you already implemented the method everywhere (crazy you!) so my suggestion suddenly becomes less relevant.

Still, if we consider that a type can be reconcilable _without_ having to override adapter settings, it might make sense to think about this as an additional `AdapterConfigurable` interface-driven behavior of some sort. Even if we want to encourage all integrations to support it at the end of the day, there might be exceptions.
The very important difference with `Reconcilable` is that, if a component doesn't implement an optional interface, things can **work** nonetheless (which is the case here).

An example of this is `EventSource` below, which can _optionally_ be implemented to expose some information about CE source and types.  In practice, all `source` types should implement it, right?
Nope! We already found a case of Source that will _not_ implement it with @odacremolbap's new `CloudEventsSource`.

pkg/apis/sources/v1alpha1/awscloudwatch_types.go
```diff
@@ -46,6 +46,10 @@ var (
 
 // AWSCloudWatchSourceSpec defines the desired state of the event source.
 type AWSCloudWatchSourceSpec struct {
+	// Adapter spec overrides parameters.
+	// +optional
+	AdapterOverrides *v1alpha1.AdapterOverrides `json:"adapterOverrides,omitempty"`
```
antoineco, line 51: nit: (applies to all Source types)
All `flow`, `extensions` and `router` types have this at the _end_ of their spec. I think it makes a lot more sense that way in terms of readability (at the end) because the type-specific attributes remain at the beginning.
Also, it's nice to be consistent in the way we do things, as a general good practice.

pkg/apis/targets/v1alpha1/alibabaoss_types.go
```diff
@@ -44,6 +44,10 @@ var (
 
 // AlibabaOSSTargetSpec holds the desired state of the AlibabaOSSTarget.
 type AlibabaOSSTargetSpec struct {
+	// Adapter spec overrides parameters.
+	// +optional
+	AdapterOverrides *v1alpha1.AdapterOverrides `json:"adapterOverrides,omitempty"`
```
antoineco, line 49: nit: (applies to all Target types)
All `flow`, `extensions` and `router` types have this at the _end_ of their spec. I think it makes a lot more sense that way in terms of readability (at the end) because the type-specific attributes remain at the beginning.
Also, it's nice to be consistent in the way we do things, as a general good practice.

pkg/reconciler/adapter.go
```diff
@@ -146,7 +146,11 @@ func commonAdapterDeploymentOptions(rcl v1alpha1.Reconcilable) []resource.Object
 		}
 	}
 
-	return objectOptions
+	overrides := rcl.GetAdapterOverrides()
+	if overrides == nil {
+		return objectOptions
+	}
+	return append(objectOptions, adapterOverrideOptions(overrides))
```
antoineco, line 153: To echo one of my previous comments, if we were to make this behavior optional (and I think it logically makes sense to consider it optional), that logic could look like:

```golang
	if cfbl, canConfigureAdapter := rcl.(v1alpha1.AdapterConfigurable); canConfigureAdapter {
		if overrides := cfbl.GetAdapterOverrides(); overrides != nil {
				return append(objectOptions, adapterOverrideOptions(overrides))
		}
	}

	return objectOptions
```

pkg/reconciler/resource/podspecable.go
```diff
@@ -159,3 +159,10 @@ func ConfigMapMount(name, target, cmName, cmKey string) ObjectOption {
 		})
 	}
 }
+
+func ContainerResources(resources *corev1.ResourceRequirements) ObjectOption {
+	return func(object interface{}) {
+		container := firstContainer(object)
+		container.Resources = *resources
+	}
+}
```
antoineco, line 168: The `resource` package is already quite dense. For example, we already have this:

https://github.com/triggermesh/triggermesh/blob/1e3bc1a29fc27455a55a31a8dcf2b2c9bb43aaf7/pkg/reconciler/resource/container.go#L203-L215

For things which aren't meant to be used by individual reconciler implementations but rather by a specific piece of code, I would encourage creating _local_ `ObjectOptions` instead.
In other words, move it to `pkg/reconciler/adapter.go` if you really need it, or use `resource.Requests` and `resource.Limits` instead to avoid having to maintain code that does one thing in different ways.

pkg/reconciler/adapter.go
```diff
@@ -309,3 +317,22 @@ func MaybeAppendValueFromEnvVar(envs []corev1.EnvVar, key string, valueFrom v1al
...
 	return envs
 }
+
+// adapterOverrideOptions applies adapter override parameters depending on
+// deployment type.
+func adapterOverrideOptions(overrides *v1alpha1.AdapterOverrides) resource.ObjectOption {
+	knServiceVisibility := resource.VisibilityClusterLocal
+	if overrides.Public {
+		knServiceVisibility = resource.VisibilityPublic
+	}
```
antoineco, line 327: Echoing a previous comment: we probably want `overrides.Public` to be optional, otherwise `kubectl get` will always show it, even if users never set it.

And I would also move those bits to the `case *servingv1.Service` because that's the only case where this parameter is relevant.

That logic could become something like:

```golang
func adapterOverrideOptions(overrides *v1alpha1.AdapterOverrides) []resource.ObjectOption {
	var opts []resource.ObjectOption // assumes we are now returning individual options

	if _, ok := object.(*servingv1.Service); ok {
		if pub := overrides.Public; pub != nil && pub {
			append(opts, resource.VisibilityPublic)
		}
	}

	if overrides.Resources != nil {
		// do the Resources magic ...
	}
}
```

pkg/reconciler/adapter.go
```diff
@@ -309,3 +317,22 @@ func MaybeAppendValueFromEnvVar(envs []corev1.EnvVar, key string, valueFrom v1al
...
+	return func(object interface{}) {
+		switch o := object.(type) {
+		case *appsv1.Deployment:
```
antoineco, line 330: Dead switch `case`?

antoineco (commented):
Looking really good 👍
I have a few suggestions around making the interface optional and cleaning up the code that assembles the override options, otherwise I think this is mostly good to go.

antoineco:
Personally I'm fine with `spec.adapterOverrides`. It does what it says: overrides the default type's adapter settings.
If we prefer `spec.adapter` that's also fine, but I'm not shocked by the current name.

pkg/apis/sources/v1alpha1/awscloudwatch_types.go
tzununbekov: I put this field at the beginning of the structure because there is another common source field `duckv1.SourceSpec` set as a first-line. Also, some of them seemed to be alphabetically ordered. Anyways, I have moved it to the bottom.

tzununbekov:
@antoineco thanks for the early review, most of the comments are taken care of.
There is an issue with `public` switch - it's based on adding and removing a particular label to Kn Service. The "removing" part of this logic does not play well with deep derivative which ignores the missing label on the expected object. I'm also clarifying the resource requests and limits behavior. Will post the update here as soon as I get something.

antoineco:
@tzununbekov ah right, I see what you mean. Labels are serialized to a Go `map`, so the deep derivative comparison cannot take the order into account (unlike `slices`).

There is a simple way to address this: write a custom [comparison function](https://github.com/triggermesh/triggermesh/blob/05df48fbb4432ecd43641cbb9393ce18e40ebf04/pkg/reconciler/semantic/semantic.go#L43-L45) for `map[string]string` that starts by comparing "special" labels (similarly to how "special" labels are propagated from the parent object to its children).

Example:

```golang
	func(a, b map[string]string) bool { // e.g. metadata.labels
		// special handling of labels that can be either added or removed by
		// TriggerMesh reconcilers.
		removableLabels := []string{network.VisibilityLabelKey}
		for _, l := range removableLabels {
			if a[l] != b[l] {
				return false
			}
		}

		return (conversion.Equalities{}).DeepDerivative(a, b)
	},
```

pkg/apis/common/v1alpha1/object_interfaces.go
```diff
@@ -41,6 +41,11 @@ type Reconcilable interface {
 	GetStatusManager() *StatusManager
 }
 
+type AdapterConfigurable interface {
```
antoineco, line 44: It would be nice to have a comment on that interface, like in the suggestion below, or anything else that conveys what this interface is useful for.

```suggestion
// AdapterConfigurable is implemented by types that can override the default
// configuration of their receive adapter.
type AdapterConfigurable interface {
```

pkg/reconciler/adapter.go
```diff
@@ -309,3 +322,40 @@ func MaybeAppendValueFromEnvVar(envs []corev1.EnvVar, key string, valueFrom v1al
...
+
+	return opts
+}
+
+func toQuantity(resources corev1.ResourceList) (kres.Quantity, kres.Quantity) {
+	var cpu, memory kres.Quantity
+
+	for k, v := range resources {
+		switch k {
+		case "cpu":
```
antoineco, line 354: nit: Kubernetes has a [constant](https://github.com/triggermesh/triggermesh/blob/7547c4b582e34706014a3aaa5f99d4e48e8050da/pkg/reconciler/resource/container.go#L228-L235) defined for that: `corev1.ResourceCPU`

pkg/reconciler/adapter.go
```diff
@@ -309,3 +322,40 @@ func MaybeAppendValueFromEnvVar(envs []corev1.EnvVar, key string, valueFrom v1al
...
+			cpu = v
+		case "memory":
```
antoineco, line 356: nit: Kubernetes has a [constant](https://github.com/triggermesh/triggermesh/blob/7547c4b582e34706014a3aaa5f99d4e48e8050da/pkg/reconciler/resource/container.go#L228-L235) defined for that: `corev1.ResourceMemory`

antoineco (approved):
Looking good to me 👍 

Approving with nits.

config/300-awscloudwatchlogssource.yaml
```diff
@@ -52,6 +52,41 @@ spec:
...
+              adapterOverrides:
+                description: Kubernetes object parameters to apply on top of default adapter values.
+                type: object
+                properties:
+                  public:
+                    description: Adapter visibility scope.
+                    type: boolean
+                  resources:
+                    description: 'Compute Resources required by this container.
+                      Cannot be updated. More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/'
```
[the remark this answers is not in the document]
tzununbekov: Not relevant. The description is about Pod's resource requirements that cannot be updated directly on the container. But in the context of Deployment or Kn Service, a rolling update does the thing. Not sure how this comment ended up here, I took the CRD part from the Knative resources. Updating.

config/300-awscloudwatchlogssource.yaml
```diff
@@ -52,6 +52,41 @@ spec:
...
+                      requests:
+                        additionalProperties:
+                          anyOf:
+                          - type: integer
+                          - type: string
+                          pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                          x-kubernetes-int-or-string: true
+                        description: 'Requests describes the minimum amount
+                          of compute resources required. If Requests is
+                          omitted for a container, it defaults to Limits
```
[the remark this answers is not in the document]
tzununbekov: Apparently, this is something Kubernetes does. Checked manually this behavior, and it matches its description. 

https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
```
If a container specifies its own memory limit, but does not specify a memory request, Kubernetes automatically
assigns a memory request that matches the limit. Similarly, if a container specifies its own CPU limit, but does not
specify a CPU request, Kubernetes automatically assigns a CPU request that matches the limit.
```

pkg/reconciler/resource/container.go
```diff
@@ -225,13 +225,17 @@ func resourcesFrom(object interface{}) (resources *corev1.ResourceRequirements)
 	return
 }
 
-func setResources(res *corev1.ResourceList, cpu, mem resource.Quantity) {
+func setResources(res *corev1.ResourceList, cpu, mem *resource.Quantity) {
```
tzununbekov, line 228: @antoineco I had to set these input types to pointers to distinguish when resource quantities are 0 and when they are not set, otherwise setting resource override for say CPU limit would automatically set the Memory limit to 0, while it should remain undefined.
And one last thing - deep derivative does not seem to react to resource overrides deletion on the desired state 🙁 Is there a way to force semantic comparison on override attributes match states "both ways"?

antoineco: Sure, changing the `resource` package to accept pointers is also OK if it doesn't support what we need.
I didn't know that a zero quantity would translate to literally `0`.

antoineco: Regarding the deep derivative comparison of resources, I'm a bit torn.
Of course we could define a custom comparison function for all the overridable fields (`corev1.ResourceList` and labels for the time being), but I'm wondering how often we expect people to 

1. Set requests/limits
1. Remove those requests/limits

I have a feeling people who do override requests/limits probably will **modify** those values (maybe), but I can't imagine a lot of cases where they will remove them.
If we think it's worth the effort, I'm all game for having an exception for `corev1.ResourceList`, but the case feels a bit edgy to me.

pkg/reconciler/resource/knservice_test.go
```diff
@@ -43,8 +45,8 @@ func TestNewServiceWithDefaultContainer(t *testing.T) {
 		EnvVar("TEST_ENV2", "val2"),
 		Label("test.label/2", "val2"),
 		ServiceAccount("god-mode"),
-		Requests(resource.MustParse("250m"), resource.MustParse("100Mi")),
-		Limits(resource.MustParse("250m"), resource.MustParse("100Mi")),
+		Requests(&cpuRes, &memRes),
+		Limits(&cpuRes, nil),
```
antoineco, line 49: In `deployment_test.go` this is set to

```golang
Requests(&cpuRes, &memRes),
Limits(&cpuRes, &memRes),
```

I don't care which one we go for, be I think we should keep both test cases identical, for consistency and to ensure they both yield the same result given the same input.

config/300-awscloudwatchlogssource.yaml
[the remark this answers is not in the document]
antoineco: Ahhh thank you, learned something!

pkg/reconciler/resource/container.go
tzununbekov: I'm fine with leaving it as it is right now until someone reports it as an issue.

PR #751, opened 2022-04-13
Clean release manifests prior to publishing

### LINKED ISSUE #280: Clean up release YAMLs before publishing them
`ko` tends to replace manifests which didn't pass a filter (label selector) with _null_ documents:

```yaml
apiVersion: myapi/v0
Kind: Example

---
       # <-- this is a null document, e.g. where a CRD used to be
---

---

---

apiVersion: myapi/v0
Kind: Example

---

# ...
```

It also keeps all license headers while concatenating manifests:

```yaml
# Copyright 2021 TriggerMesh Inc.
#
# Licensed under the Apache License, Version 2.0 (the "License");

apiVersion: myapi/v0
Kind: Example

---

# Copyright 2021 TriggerMesh Inc.
#
# Licensed under the Apache License, Version 2.0 (the "License");

apiVersion: myapi/v0
Kind: Example

---

# ...
```

Both of them are harmless, but since we treat this output as release artifacts, we could as well keep them readable by cleaning up these unnecessary parts:

```yaml
# Copyright 2021 TriggerMesh Inc.
#
# Licensed under the Apache License, Version 2.0 (the "License");

apiVersion: myapi/v0
Kind: Example

---

apiVersion: myapi/v0
Kind: Example

---

# ...

antoineco:
Closes #280

Introduces the `hack/manifest-cleaner` tool which performs the clean up tasks described in the referenced issue to the manifests generated by `ko resolve` (filters out null YAML documents, removes redundant copyright headers).

It is written using the same packages as `hack/rbac-check`.

I modified the CircleCI job to make use of it already.


antoineco:
Please be aware that the `gopkg.in/yaml.v3` encoder has its own preference about serializing comments scoped at an element of an array. Instead of leaving it right above the element, it inlines it with the `-` marker.
This is because, after parsing, each comment becomes related to a specific node (a node can be the document itself, a string, an object, an array, ...).

I just learned this syntactic sugar myself, and yes, it is perfectly legal in YAML.
There will be some extra work necessary in case we don't want this behavior for any reason.

image

sebgoa:
that's pretty ugly if you ask me, but oh well.

antoineco:
Let's stick with the status quo then.

sebgoa:
You are closing this too fast. 

It is not because I think the way this linter handles comments in yaml is ugly that the PR should not get merged. 

Go ahead and merge it, it solves the original issue due to `ko` but I can have the opinion that the comments look ugly - something totally subjective and inconsequential-

sebgoa:
I disagree. I would have merged it since it closes the original issue.

antoineco:
Ok sorry that was ambiguous since you said it looked ugly and didn't approve.

sebgoa (approved):

sameersbn (approved):

PR #754, opened 2022-04-14
Normalize Godoc of Spec, Status and List types

antoineco:
We have very heterogenous Godocs for Spec, Status and List types, although they all represent the _same thing_ regardless of the component.

So I ran a string replacement over all `pkg/apis/*/*_types.go` files to apply the following normalized Godoc:

| Type | Normalized doc |
|--------|--------------------------------------------------------|
| Spec   | FooSpec defines the desired state of the component.     |
| Status | FooStatus defines the observed state of the component. |
| List   | FooList is a list of component instances.              |


pkg/apis/flow/v1alpha1/jqtransformation_types.go
```diff
@@ -53,15 +53,9 @@ type JQTransformationSpec struct {
 	duckv1.SourceSpec `json:",inline"`
 }
 
-// JQTransformationStatus communicates the observed state of the JQTransformation (from the controller).
-type JQTransformationStatus struct {
-	duckv1.SourceStatus  `json:",inline"`
-	duckv1.AddressStatus `json:",inline"`
-}
```
antoineco, line 60: This type wasn't used anywhere, so I used the occasion to delete it.

sebgoa: ok

sebgoa (approved):
thanks for the cleanup

Issue #673, opened 2022-03-21
Base Dockerfile images were updated to Go 1.18, builds fail with `error obtaining VCS status`


antoineco:
The Docker tag `bullseye` was recently [updated to Go 1.18](https://github.com/docker-library/golang/pull/411), and builds started failing with the following error:

```
# cd /go/***********; git status --porcelain
fatal: not a git repository (or any of the parent directories): .git
error obtaining VCS status: exit status 128
	Use -buildvcs=false to disable VCS stamping.
```

Relevant extract from the [Go 1.18 release notes](https://go.dev/doc/go1.18):

> The `go` command now embeds version control information in binaries. It includes the currently checked-out revision, commit time, and a flag indicating whether edited or untracked files are present. Version control information is embedded if the `go` command is invoked in a directory within a Git, Mercurial, Fossil, or Bazaar repository, and the `main` package and its containing main module are in the same repository. This information may be omitted using the flag `-buildvcs=false`.

The reason why this fails is because we are building from a directory that contains a `.git` subdirectory, but that subdirectory is only partially copied to the Docker build context due to the following `.dockerignore` exclusion:

https://github.com/triggermesh/triggermesh/blob/ebe47d2f09c1ab700d3b1df44ea97ee29db7b0ff/.dockerignore#L4

We may want to relax this exclusion when we update to Go 1.18, especially since the `.git` subdirectory is only accessed during the [`build` stage](https://github.com/triggermesh/triggermesh/blob/ebe47d2f09c1ab700d3b1df44ea97ee29db7b0ff/cmd/ibmmqsource-adapter/Dockerfile#L1), then dropped when artifacts are copied to the [`runtime` stage](https://github.com/triggermesh/triggermesh/blob/ebe47d2f09c1ab700d3b1df44ea97ee29db7b0ff/cmd/ibmmqsource-adapter/Dockerfile#L25).

Here are the two code locations containing references to the `golang:bullseye` base image:

https://github.com/triggermesh/triggermesh/blob/ebe47d2f09c1ab700d3b1df44ea97ee29db7b0ff/cmd/ibmmqsource-adapter/Dockerfile#L1

https://github.com/triggermesh/triggermesh/blob/ebe47d2f09c1ab700d3b1df44ea97ee29db7b0ff/cmd/ibmmqtarget-adapter/Dockerfile#L1

antoineco:
For posterity, here is the kind of VCS information Go 1.18 embeds in binaries:

```
[...]

PR #755, opened 2022-04-15
Copy Git objects in Docker build steps

antoineco:
This is a preparation step for updating to Go 1.18.

Go 1.18 [embeds VCS info](https://shibumi.dev/posts/go-18-feature/) into built binaries by default, based on the contents of the `.git` directory. (Ref. #673)
Without the `objects/` directory and `index` file, `go build` can't obtain the Git status and fails with:

```
error obtaining VCS status: exit status 128
	Use -buildvcs=false to disable VCS stamping.
```

Alternatively, we could disable VCS stamping, as the error message suggests.
I believe including VCS information might be useful in case some package wants to depend on that information for any reason, but I don't have strong opinions about it.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,7 +1,10 @@
 # Top level files and directories
+#
+# (!) The '.git' directory is purposely not excluded because the Go toolchain
+#     relies on its contents to embed VCS info into binaries.
+#
 /.assets/
 /.circleci/
-/.git/
 /.github/
 /docs/
 /hack/
@@ -15,10 +18,6 @@
 /CODEOWNERS
 /README.md
 /kustomization.yaml
-
-# Files copied to KO_DATA_PATH directory
-!/.git/HEAD
-!/.git/refs/
 
 # Patterns
 **/config/
--- a/cmd/dataweavetransformation-adapter/Dockerfile
+++ b/cmd/dataweavetransformation-adapter/Dockerfile
@@ -28,8 +28,9 @@ RUN go build -o /dataweavetransformation-adapter ./cmd/dataweavetransformation-a
 
 FROM debian:stable-slim
 
-# Ensure the /kodata entries are present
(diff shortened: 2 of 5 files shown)


antoineco:
For the curious, here is a sample of the VCS information included in the compiled binary's buildinfo:

```
(*debug.BuildInfo)(0xc000354360)(go     go1.18.1
...
build   vcs=git
build   vcs.revision=8a454dc07dc762ed5083515b56a9ab44aafe6f67
build   vcs.time=2022-04-15T16:32:54Z
build   vcs.modified=true
)
```

Note: `vcs.modified` will always be `true` in a Docker build, because some irrelevant files are excluded via `.dockerignore` and therefore show as `D` (deleted) in `git status`. I don't think that's important.

.dockerignore
```diff
@@ -16,6 +16,10 @@
 /README.md
 /kustomization.yaml
 
+# Files used by the Go toolchain to embed VCS info into binaries
+!/.git/index
+!/.git/objects/
```
cab105, line 21: At this point, wouldn't we just use `!/.git/`?

cab105 (approved):
Left a comment more for optimizing `.dockerignore`, but isn't a blocker.

.dockerignore
antoineco: That's probably a better approach at that point, you're right.

I'm not familiar with all the files Git uses, but for context I wanted to ensure we include only what's required by `git status`, and not the configuration of remotes, etc.

I'll try to work that out with an exclusion instead of these multiple inclusions.

antoineco: Done in [b04cad3c..bd85d927](https://github.com/triggermesh/triggermesh/compare/b04cad3ccc6dc547a24e8d23117bcd39e177bbbf..bd85d9274795c00c626e6b74825ed06dafed54d5)

cab105 (approved):
Noice!

PR #763, opened 2022-04-19
Bump to Go 1.18 in CI workflows

antoineco:
Our dependencies are starting to adopt Go generics (`Azure/azure-sdk-for-go` leading the way), so I suggest we start updating our own automation to use Go 1.18.

I am purposely not updating `go.mod` in this PR, because TriggerMesh still builds with Go 1.17 on the current `main`.
I think we should bump the version inside `go.mod` only on the next update of Azure SDK for Go (or any other dependency that uses generics).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ jobs:
   gomod-cache:
     executor:
       name: go/default
-      tag: '1.17'
+      tag: '1.18'
     steps:
       - checkout
       - go/mod-download-cached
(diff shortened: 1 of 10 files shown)


.circleci/config.yml
```diff
@@ -133,7 +133,7 @@ jobs:
             - go-build-{{ arch }}-{{ checksum "go.sum" }}
       - run:
           name: Installing ko
-          command: go install github.com/google/ko@v0.11.1
+          command: go install github.com/google/ko@v0.11.2
```
antoineco, line 136: I also bumped the `ko` version because `v0.11.2` fixes a small bug that's specific to Go 1.18.

odacremolbap (approved):

PR #766, opened 2022-04-19
Use generics to simplify RBACOwnersLister

antoineco:
This PR leverages generics to simplify the `RBACOwners()` method all reconcilers are currently implementing, which was a lot of boilerplate code.

```diff
+102
-1027
```


cab105 (approved):
Aren't generics awesome!

antoineco:
They are the perfect tool to solve this specific use-case!

I already have ideas to convert those 75 typed Listers to one generic Lister type, so that adapters don't need to implement yet another interface at all...

PR #772, opened 2022-04-21
Move redundant `RBACOwnersLister` implementations to generic reconcilers

### LINKED ISSUE #771: Clean up redundant implementations of `RBACOwnersLister`
In #766, I simplified the way components implement `RBACOwnersLister` using generics.

I believe we can go one step further and _remove_ all reconciler-specific implementations of this interface by converting all typed `Lister`, currently embedded inside `Reconciler` structs, to a generic `Lister` that we would embed inside `GenericRBACReconciler`.

Example:

```diff
type Reconciler struct {
	base       common.GenericDeploymentReconciler
	adapterCfg *adapterConfig
-
-	srcLister func(namespace string) listersv1alpha1.FooNamespaceLister
}
```

```diff
type GenericRBACReconciler struct {
	// API clients
	SAClient func(namespace string) coreclientv1.ServiceAccountInterface
	RBClient func(namespace string) rbacclientv1.RoleBindingInterface
	// objects listers
	SALister func(namespace string) corelistersv1.ServiceAccountNamespaceLister
	RBLister func(namespace string) rbaclistersv1.RoleBindingNamespaceLister
+	OwnersLister func(namespace string) lister[T]
}
```

Where `lister` is this interface:

https://github.com/triggermesh/triggermesh/blob/0787d1aa658d7e4ae7c2c425dde47b0b8f1c85d2/pkg/reconciler/adapter.go#L382-L384

antoineco:
Closes #771

Instead of expecting every component to return a list of owners for their own ServiceAccount — which was clunky — move that logic to the `GenericRBACReconciler`, which is made possible by Go generics.

This refactoring doesn't reduce the amount of code overall (boilerplate was already removed in #766), but moves a core piece of the reconciliation logic out of the individual implementations.


pkg/reconciler/base.go
```diff
@@ -56,46 +56,48 @@ type GenericDeploymentReconciler struct {
...
 	// API clients
 	SAClient func(namespace string) coreclientv1.ServiceAccountInterface
 	RBClient func(namespace string) rbacclientv1.RoleBindingInterface
 	// objects listers
-	SALister func(namespace string) corelistersv1.ServiceAccountNamespaceLister
-	RBLister func(namespace string) rbaclistersv1.RoleBindingNamespaceLister
+	SALister     func(namespace string) corelistersv1.ServiceAccountNamespaceLister
+	RBLister     func(namespace string) rbaclistersv1.RoleBindingNamespaceLister
+	OwnersLister ListerGetter[T]
 }
```
antoineco, line 83: The most important part of the PR is here.

The typed listers that used to be embedded in each component's `Reconciler` are now moved to the `GenericRBACReconciler` as a generic lister type.

pkg/reconciler/reconcile.go
```diff
@@ -96,9 +88,12 @@ func (r *GenericDeploymentReconciler) ReconcileAdapter(ctx context.Context, ab A
...
-	saOwners, err := ab.RBACOwners(rcl)
-	if err != nil {
-		return fmt.Errorf("listing ServiceAccount owners: %w", err)
+	saOwners := []kmeta.OwnerRefable{rcl}
+	if !v1alpha1.WantsOwnServiceAccount(rcl) {
+		var err error
+		if saOwners, err = rbacOwners(r.OwnersLister(rcl.GetNamespace())); err != nil {
+			return fmt.Errorf("listing ServiceAccount owners: %w", err)
+		}
 	}
```
antoineco, line 97: ... and we use that generic lister to fetch the list of OwnerRefables, instead of requiring reconcilers to implement a `RBACOwners)()` method.

antoineco (commented):
Notes

pkg/reconciler/reconcile.go
```diff
@@ -383,7 +381,7 @@ func (r *GenericServiceReconciler) syncAdapterService(ctx context.Context,
 }
 
 // findAdapter returns the adapter object for a given component instance if it exists.
-func findAdapter(genericReconciler interface{},
+func findAdapter[T kmeta.OwnerRefable](genericReconciler interface{},
```
tzununbekov, line 384: Please bear with me, I'm still trying to understand generics' use-cases. What's the practical value of having `findAdapter` implemented as a generic function instead of making it a method for different types of reconciler receivers?

tzununbekov (approved):
I understand the idea that stands behind this PR but the lack of the Go generics' expertise is not allowing me to do a complete conclusion about this change. Code vise the change is pretty straightforward, and, if it's enough for the review, I approve it.

pkg/reconciler/reconcile.go
antoineco: That's a great question! The single `findAdapter` function was my attempt (already before generics) to try reducing code duplication. But it resulted in something a bit cluttered.

To be clear, the `[T OwnerRefable]` on this line is only here because the function handles generic types, so the compiler forces you to have it here. I agree that it doesn't look right. I want to refactor that file a bit to use generics properly, but meanwhile I can split this function in this PR. I agree, it would make things less cluttered.

antoineco: Addressed in 7ea4ba93

cab105 (approved):

Issue #768, opened 2022-04-20
Allow setting nodeSelector and affinity


isaac88:
Right now we're trying to use that event source: https://docs.triggermesh.io/apis/sources/#sources.triggermesh.io/v1alpha1.AWSSQSSource but the POD awssqssource-adapter is not able to be schedule in our kubernetes cluster because we have different tainted node groups and it's require to set the proper tolerations in the POD, so the POD will run on the correct group of nodes.

It'll be really useful to allow setting nodeSelector, tolerations, and affinity on all Pods to accomplish this.
It will be great to have that option available to any CRDs that create pods.

https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/ 

Regards

antoineco:
@isaac88 `tolerations` can now be set on all TriggerMesh components.

The feature is currently only available on the `main` branch. Deployment manifests can be found here: https://app.circleci.com/pipelines/github/triggermesh/triggermesh/2083/workflows/e48573bb-bcce-4610-88e0-aaa99541b7b7/jobs/5638/artifacts.
Those are development artifacts, the feature will be "officially" available in the next release of TriggerMesh.

I'm keeping this issue open until we implement support for `nodeSelector` and `affinity`.
Meanwhile, could you please share how you expect to use those two features, just so that I get a better understanding of the use case?

sebgoa:
@mukundjalan can you elaborate on what you would need exactly ? nodeSelector ? affinity ? And for which components ?

Any additional details would help schedule this

cc/ @FranBarrera

FranBarrera:
@mukundjalan @isaac88, we just merge the PR to add support for Affinity, so we now support tolerations, nodeSelector and affinity but nodeSelector and affinity are currently only available on the main branch. Deployment manifests can be found here: https://github.com/triggermesh/triggermesh/suites/12206066321/artifacts/645145411
Those are development artifacts. You can safely deploy them on top of your existing TriggerMesh installation in your test environment, but the feature will only become officially available in the next release of TriggerMesh.

isaac88:
Thanks a lot @FranBarrera @sebgoa

PR #774, opened 2022-04-21
Add support for setting adapter tolerations

antoineco:
Addresses one of the features requested in #768.

Example:

```yaml
spec:
  # ...
  adapterOverrides:
    tolerations:
    - key: example-key
      operator: Exists
      effect: NoSchedule
```

Result:

```console
$ kubectl get deployment awssqssource-sample -o jsonpath='{.spec.template.spec.tolerations}'
[{"effect":"NoSchedule","key":"example-key","operator":"Exists"}]
```

### DIFF
--- a/config/300-awscloudwatchlogssource.yaml
+++ b/config/300-awscloudwatchlogssource.yaml
@@ -83,6 +83,31 @@ spec:
                           specified, otherwise to an implementation-defined value. More info:
                           https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/'
                         type: object
+                  tolerations:
+                    description: Pod tolerations, as documented at
+                      https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/
+                    type: array
+                    items:
+                      type: object
+                      properties:
+                        key:
+                          description: Taint key that the toleration applies to.
+                          type: string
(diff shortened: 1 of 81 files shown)


tzununbekov (approved):
Indeed, great timing for adapter overrides!
One thought that I had after adding the initial set of attributes: what do you think about making `adapterOverrides` object in the CRD accept arbitrary data and move the verification logic to the webhook? That would significantly reduce the number of changes required by this kind of simple update.

antoineco:
That is true, I'm just not very familiar with the webhook implementation. I guess we would just need to put the initial effort into adding `_validation.go` files for each type?

tzununbekov:
Yes, adding `Validate` [methods](https://github.com/triggermesh/triggermesh/blob/acae0fbd247e8898ef0bfba7e04ee739b5383ac0/pkg/apis/flow/v1alpha1/transformation_validation.go) for our resources in `_validate.go` and listing them in webhook's [types](https://github.com/triggermesh/triggermesh/blob/f88cfe4e78ea44bb2ca8290cad8f2cf10d9bb059/cmd/triggermesh-webhook/main.go#L37-L40) should do the thing.

antoineco:
OK let's follow up on that.

PR #775, opened 2022-04-22
Merge generic Deployment and Service reconcilers into a truly generic reconciler

antoineco:
Merges `GenericDeploymentReconciler` and `GenericServiceReconciler` into `GenericAdapterReconciler[T]` (uses Go generics).
This removes all the code duplicated in the previous two implementations.

Marking as draft, depends on #772.

Also, I'm not sure I'm happy with the result. Although this change does remove code duplication, some parts of it feel convoluted and this fully generic approach may prevent future additions to one implementation or the other. For example, if we ever want to reconcile a Kubernetes Service as part of the reconciliation of a Deployment, we will need to split the logic again.


antoineco:
Based on this attempt of implementation, it seems clear to me that we shouldn't be doing this, even if we technically can.

A little repetition is better than the wrong abstraction.

I will keep exploring how we can leverage generic _functions_ instead of genetic types in other to reduce code duplication inside `pkg/reconciler`.

PR #779, opened 2022-04-25
Bumblebee: Fix Adapter's CloudEvents client 

### LINKED ISSUE #777: `event_count` metric missing because adapter creates its own CloudEvents client instead of using Knative's
The adapter constructor receives a CloudEvents client from Knative:

https://github.com/triggermesh/triggermesh/blob/1804c1441d2331349eafa719d6cd7c2790a818df/pkg/flow/adapter/transformation/handler.go#L73

However, this client is discarded and a new one is created:

https://github.com/triggermesh/triggermesh/blob/1804c1441d2331349eafa719d6cd7c2790a818df/pkg/flow/adapter/transformation/handler.go#L126

As a result, the `transformation_event_count` (dispatched events) metric isn't published via the metrics exporter.

tzununbekov:
Utilize CloudEvents client provided by the Knative adapter package to export Prometheus metrics.
`handler.go` and `handler_test.go` were renamed to `adapter.go` and `adapter_test.go` to match common structure and better reflect the file's content.
Closes #777

### DIFF
--- a/pkg/flow/adapter/transformation/adapter.go
+++ b/pkg/flow/adapter/transformation/adapter.go
@@ -81,59 +81,42 @@ func NewAdapter(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClie
 	trnContext, trnData := []v1alpha1.Transform{}, []v1alpha1.Transform{}
 	err := json.Unmarshal([]byte(env.TransformationContext), &trnContext)
 	if err != nil {
-		log.Fatalf("Cannot unmarshal Context Transformation variable: %v", err)
+		log.Fatalf("Cannot unmarshal context transformation env variable: %v", err)
 	}
 	err = json.Unmarshal([]byte(env.TransformationData), &trnData)
 	if err != nil {
-		log.Fatalf("Cannot unmarshal Data Transformation variable: %v", err)
+		log.Fatalf("Cannot unmarshal data transformation env variable: %v", err)
 	}
 
-	adapter, err := newHandler(trnContext, trnData)
+	contextPl, err := newPipeline(trnContext)
 	if err != nil {
-		log.Fatalf("Cannot create transformation handler: %v", err)
+		log.Fatalf("Cannot create context transformation pipeline: %v", err)
 	}
+
+	dataPl, err := newPipeline(trnData)
+	if err != nil {
+		log.Fatalf("Cannot create data transformation pipeline: %v", err)
+	}
+
+	sharedStorage := storage.New()
+	contextPl.setStorage(sharedStorage)
+	dataPl.setStorage(sharedStorage)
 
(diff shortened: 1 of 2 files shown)


pkg/flow/adapter/transformation/adapter.go
```diff
@@ -81,59 +81,42 @@ func NewAdapter(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClie
 	trnContext, trnData := []v1alpha1.Transform{}, []v1alpha1.Transform{}
 	err := json.Unmarshal([]byte(env.TransformationContext), &trnContext)
 	if err != nil {
-		log.Fatalf("Cannot unmarshal Context Transformation variable: %v", err)
+		log.Fatalf("Cannot unmarshal context transformation env variable: %v", err)
```
antoineco, line 84: While we're at fatal'ing things, any chance we can use Knative's logger instead of Go's stdlib `log` package, like everywhere else in TriggerMesh?

https://github.com/triggermesh/triggermesh/blob/f88cfe4e78ea44bb2ca8290cad8f2cf10d9bb059/pkg/sources/adapter/googlecloudpubsubsource/adapter.go#L73

...

https://github.com/triggermesh/triggermesh/blob/f88cfe4e78ea44bb2ca8290cad8f2cf10d9bb059/pkg/sources/adapter/googlecloudpubsubsource/adapter.go#L81

I'm fine if you prefer to do it in a separate PR.

pkg/flow/adapter/transformation/adapter_test.go
```diff
@@ -35,22 +36,29 @@ var availableTransformations = []v1alpha1.Transform{
 	{Operation: "delete"},
 }
 
-func TestNewHandler(t *testing.T) {
-	_, err := newHandler(availableTransformations, availableTransformations)
-	assert.NoError(t, err)
-}
-
 func TestStart(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
```
antoineco, line 40: With this refactoring, the instantiation of `ctx` and `cancel` ends up being far away from where they are actually used.
For the sake of code clarity, I would recommend moving it right before the place where `a.Start(ctx)` is called.

```diff
+	ctx, cancel := context.WithCancel(context.Background())

	go func() {
		defer close(errChan)
		errChan <- a.Start(ctx)
	}()

	cancel()
```

antoineco (approved):

pkg/flow/adapter/transformation/adapter.go
tzununbekov: There are a couple of other places where it needs to be changed, so I'd make it in a separate PR
https://github.com/triggermesh/triggermesh/issues/780

PR #782, opened 2022-04-25
Update 200-dataweavetransformation.yaml

jmcx:
Looks like the kind is case sensitive, expects DataWeave instead of Dataweave.


FranBarrera (approved):

antoineco (approved):
Nice catch 👍 

https://github.com/triggermesh/triggermesh/blob/d0136eb039eca945a36bf28e20628e4afe4aee32/config/304-dataweavetransformation.yaml#L34-L36

PR #783, opened 2022-04-25
Support IAM role auth in S3 source

antoineco:
Requested by a user in https://github.com/triggermesh/triggermesh/issues/244#issuecomment-1108796311

Same as #248, but for the Amazon S3 source.

How to use:

```yaml
kind: AWSS3Source
metadata:
  name: sample
spec:
  arn: arn:aws:s3:::triggermeshtest

  auth:
    iamRole: arn:aws:iam::123456789012:role/foo

  sink:
    ref:
      apiVersion: v1
      kind: Service
      name: event-display
```

### DIFF
--- a/config/300-awss3source.yaml
+++ b/config/300-awss3source.yaml
@@ -220,6 +220,16 @@ spec:
                         oneOf:
                         - required: [value]
                         - required: [valueFromSecret]
+                  iamRole:
+                    description: (Amazon EKS only) The ARN of an IAM role which can be impersonated to obtain AWS
+                      permissions. For more information about IAM roles for service accounts, please refer to the Amazon
+                      EKS User Guide at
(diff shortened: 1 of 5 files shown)


cab105 (approved):
I'm surprised this worked given `iamRole` was missing in the config yaml.

antoineco:
@cab105 that's the thing, it wasn't working because we didn't implement it in this source :)

We could replicate the changes from this PR to all AWS sources, since we know it works as expected.

antoineco:
@isaac88 please be aware that, in the case of the S3 source, the TriggerMesh controller still needs to be able to interact with the configuration of your bucket (s).
As such, it should also be assigned sufficient permissions via the `triggermesh-controller` service account located inside the `triggermesh` namespace.

PR #809, opened 2022-04-28
Change signature of `AdapterBuilder` methods to return an error

### LINKED ISSUE #808: Change signature of `AdapterBuilder` methods to return an error
In some cases, the adapter building logic may fail, in which case it would be useful to be able to return an error.

Note: all errors that occur while building an adapter should be considered permanent, and handled/wrapped as such by the generic reconciler.

https://github.com/triggermesh/triggermesh/blob/c48dc55efa3a32ce0d5e9c72bdbc9295482be9e0/pkg/reconciler/reconcile.go#L54-L64

antoineco:
🎁 for @tzununbekov 

Closes #808


pkg/reconciler/reconcile.go
```diff
@@ -54,13 +54,13 @@ var knativeServingAnnotations = []string{
 // AdapterDeploymentBuilder provides all the necessary information for building
 // a component's adapter backed by a Deployment.
 type AdapterDeploymentBuilder interface {
-	BuildAdapter(rcl v1alpha1.Reconcilable, sinkURI *apis.URL) *appsv1.Deployment
+	BuildAdapter(rcl v1alpha1.Reconcilable, sinkURI *apis.URL) (*appsv1.Deployment, error)
 }
```
antoineco, line 58: Method signature updated here.

pkg/reconciler/reconcile.go
```diff
@@ -86,7 +86,11 @@ func (r *GenericDeploymentReconciler[T, L]) ReconcileAdapter(ctx context.Context
 	}
 	rcl.GetStatusManager().MarkSink(sinkURI)
 
-	desiredAdapter := ab.BuildAdapter(rcl, sinkURI)
+	desiredAdapter, err := ab.BuildAdapter(rcl, sinkURI)
+	if err != nil {
+		return controller.NewPermanentError(reconciler.NewEvent(corev1.EventTypeWarning,
+			ReasonInvalidSpec, "Could not generate desired state of adapter Deployment: %s", err))
+	}
```
antoineco, line 93: Error handled here as a fatal.
Rationale: if we can't generate the desired state of the adapter from the provided spec, the spec is most likely invalid.

pkg/sources/reconciler/awscloudwatchsource/adapter.go
```diff
@@ -55,12 +56,15 @@ type adapterConfig struct {
...
+func (r *Reconciler) BuildAdapter(src commonv1alpha1.Reconcilable, sinkURI *apis.URL) (*appsv1.Deployment, error) {
 	typedSrc := src.(*v1alpha1.AWSCloudWatchSource)
 
 	var queries string
 	if qs := typedSrc.Spec.MetricQueries; len(qs) > 0 {
-		q, _ := json.Marshal(qs)
+		q, err := json.Marshal(qs)
+		if err != nil {
+			return nil, fmt.Errorf("serializing spec.metricQueries to JSON: %w")
+		}
```
antoineco, line 67: I used the occasion to actually return an error in the few places where we used to ignore them (3 occurrences across all TriggerMesh components).

antoineco (commented):
Notes

tzununbekov (approved):
Nice!

PR #810, opened 2022-04-28
Add Volumes and VolumeMounts resource helpers

antoineco:
I like how the CloudEvents source has its own helper for applying volumes and mounts.
I think we can decouple that approach even further by separating the creation of the structs from the code that actually applies them.

Next, I would like to apply the same change to the IBM MQ components and remove the overly complex `ConfigMapMount` and `SecretMount` helpers.

### DIFF
--- a/pkg/sources/reconciler/cloudeventssource/adapter.go
+++ b/pkg/sources/reconciler/cloudeventssource/adapter.go
@@ -59,38 +59,35 @@ var _ common.AdapterServiceBuilder = (*Reconciler)(nil)
 func (r *Reconciler) BuildAdapter(src commonv1alpha1.Reconcilable, sinkURI *apis.URL) (*servingv1.Service, error) {
 	typedSrc := src.(*v1alpha1.CloudEventsSource)
 
-	ceOverridesStr := cloudevents.OverridesJSON(typedSrc.Spec.CloudEventOverrides)
-
-	options := []resource.ObjectOption{
-		resource.Image(r.adapterCfg.Image),
-
-		resource.VisibilityPublic,
-
-		resource.EnvVar(adapter.EnvConfigCEOverrides, ceOverridesStr),
-		resource.EnvVars(r.adapterCfg.configs.ToEnvVars()...),
-		resource.EnvVars(makeAppEnv(typedSrc)...),
-	}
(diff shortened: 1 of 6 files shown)


pkg/sources/reconciler/cloudeventssource/adapter.go
```diff
@@ -99,16 +96,34 @@ func (r *Reconciler) BuildAdapter(src commonv1alpha1.Reconcilable, sinkURI *apis
...
+	ceOverridesStr := cloudevents.OverridesJSON(typedSrc.Spec.CloudEventOverrides)
+
+	return common.NewAdapterKnService(src, sinkURI,
+		resource.Image(r.adapterCfg.Image),
+
+		resource.VisibilityPublic,
+
+		resource.Volumes(authVolumes...),
+		resource.VolumeMounts(authVolumeMounts...),
+		resource.EnvVars(authEnvs...),
```
antoineco, line 121: One of the nice properties of `append()` is that the operation is a no-op if the argument is empty.
So if auth isn't set, these 3 actions won't fail but the constructor remains easy to read (I think).

pkg/sources/reconciler/cloudeventssource/adapter.go
```diff
@@ -149,41 +164,27 @@ func makeAppEnv(o *v1alpha1.CloudEventsSource) []corev1.EnvVar {
...
-
-		ksvc.Spec.Template.Spec.Containers[0].VolumeMounts = append(
-			ksvc.Spec.Template.Spec.Containers[0].VolumeMounts,
-			corev1.VolumeMount{
-				Name:      name,
-				ReadOnly:  true,
-				MountPath: mountPath,
+// secretVolumeAndMountAtPath returns a Secret-based volume and corresponding
+// mount at the given path.
+func secretVolumeAndMountAtPath(name, mountPath, mountFile, secretName, secretKey string) (corev1.Volume, corev1.VolumeMount) {
```
antoineco, line 169: This function was quite complex because it mixed structs instantiations with constructor logic that was conditionally creating containers, etc.
Now, it just returns the structs, and the object is constructed elsewhere.

antoineco (commented):
Notes

odacremolbap (approved):

PR #811, opened 2022-04-28
DataWeaveTransformation - Add support to send the spell at runtime

### LINKED ISSUE #804: Add Support to send the spell to a DataWeaveTransformation at runtime

FranBarrera:
Closes #804


pkg/flow/adapter/dataweavetransformation/adapter.go
```diff
@@ -96,49 +115,95 @@ func (a *dataweaveTransformAdapter) Start(ctx context.Context) error {
...
+	if req.InputContentType != "" || req.OutputContentType != "" || req.Spell != "" {
+		if req.InputData == "" {
+			return a.replier.Error(&event, targetce.ErrorCodeRequestValidation,
+				errors.New("inputData not found"), nil)
+		}
+	}
+
+	// Check for spellOverride to be enabled and all the parameters to be present.
+	if a.spellOverride && req.InputData != "" && req.InputContentType != "" && req.OutputContentType != "" && req.Spell != "" {
+		inputData = []byte(req.InputData)
```
antoineco, line 152: I see a huge code smell here.
We rely on a global state but the code is concurrent, so multiple goroutines will be writing this value at the same time.

FranBarrera: It should be fixed now

pkg/flow/adapter/dataweavetransformation/adapter.go
```diff
@@ -96,49 +115,97 @@ func (a *dataweaveTransformAdapter) Start(ctx context.Context) error {
...
-		if !isValidJSON(event.Data()) {
-			return a.replier.Error(&event, targetce.ErrorCodeRequestParsing,
-				errors.New("invalid Json"), nil)
+	err = validateContentType(event.DataContentType(), event.Data())
+	if err != nil {
+		return a.replier.Error(&event, targetce.ErrorCodeRequestValidation,
+			errors.New(err.Error()), nil)
+	}
+
+	req := &DataWeaveTransformationStructuredRequest{}
```
cab105, line 130: Will the `DataWeaveTransformationStructuredRequest` be the required cloudevent payload format going forward, or will it still be possible to pass just the data to be translated?  If the latter, would suggest migrating the `spellOverride` check up.

pkg/flow/adapter/dataweavetransformation/adapter.go
```diff
@@ -96,49 +115,97 @@ func (a *dataweaveTransformAdapter) Start(ctx context.Context) error {
...
+
+	err = validateContentType(inputContentType, inputData)
+	if err != nil {
+		return a.replier.Error(&event, targetce.ErrorCodeRequestValidation,
+			errors.New(err.Error()), nil)
+	}
+
+	switch inputContentType {
+	case "application/json":
+		tmpFile, err = ioutil.TempFile(path, "*.json")
```
cab105, line 182: The file extension doesn't really matter when invoking `dw` given you are explicitly specifying the input/output content-types.

pkg/flow/adapter/dataweavetransformation/adapter.go
```diff
@@ -96,49 +115,97 @@ func (a *dataweaveTransformAdapter) Start(ctx context.Context) error {
...
+			errors.New("no default Spell or in request found"), nil)
+	}
+
+	if req.InputContentType != "" || req.OutputContentType != "" || req.Spell != "" {
+		if req.InputData == "" {
+			return a.replier.Error(&event, targetce.ErrorCodeRequestValidation,
+				errors.New("inputData not found"), nil)
 		}
-		tmpfile, err = ioutil.TempFile(path, "*.json")
+	}
```
cab105, line 150: These checks seem a bit clunky to me as you could rely on the defaults specified by the adapter when checking if the associated attributes are not included with the payload.

pkg/flow/adapter/dataweavetransformation/adapter.go
FranBarrera: It will allow the both ways, but I've added some checks with the DataWeaveTransformationStructuredRequest that it needs to be before the spellOverride check

pkg/flow/adapter/dataweavetransformation/adapter.go
```diff
@@ -73,14 +84,25 @@ func NewTarget(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClien
...
+	if env.DwSpell != "" {
+		adapter.defaultSpell = &env.DwSpell
+	}
+	if env.InputContentType != "" {
+		adapter.defaultInputContentType = &env.InputContentType
+	}
+	if env.OutputContentType != "" {
+		adapter.defaultOutputContentType = &env.OutputContentType
+	} else {
+		adapter.defaultOutputContentType = ptr.String("application/json")
```
cab105, line 105: Interesting. Didn't realize there was a knative pkg/method call for this now. Figured: `str := "application/json"; adapter.defaultOutputContentType = &str` would work.

You don't need to address this.

pkg/flow/adapter/dataweavetransformation/adapter.go
```diff
@@ -96,73 +118,104 @@ func (a *dataweaveTransformAdapter) Start(ctx context.Context) error {
...
+		if req.OutputContentType != "" {
+			outputContentType = &req.OutputContentType
+		}
 
-	event.SetType(event.Type() + ".response")
-	a.logger.Infof("responding with transformed event: %v", event.Type())
-	if a.sink != "" {
-		if result := a.ceClient.Send(ctx, event); !cloudevents.IsACK(result) {
-			a.logger.Errorf("Error sending event to sink: %v", result)
+		if req.InputData != "" {
```
cab105, line 161: Is there a time when the input data is optional?

pkg/flow/adapter/dataweavetransformation/adapter.go
```diff
@@ -96,73 +118,104 @@ func (a *dataweaveTransformAdapter) Start(ctx context.Context) error {
...
+		}
+
+		if req.Spell != "" {
+			spell = &req.Spell
 		}
-		return nil, cloudevents.ResultACK
 	}
 
-	return &event, cloudevents.ResultACK
+	if inputContentType != nil && outputContentType != nil && inputData != nil && spell != nil {
```
cab105, line 170: Wouldn't it be better to check the inverse case?

```
if inputContentType != nil || outputContentType != nil || inputData != nil || spell != nil {
  return a.replier.Error(&event, targetce.ErrorCodeRequestValidation,
			errors.New("parameters not found"), nil)
}
```

cab105 (commented):
Much closer. See a couple more tweaks that can be made.

pkg/flow/adapter/dataweavetransformation/adapter.go
FranBarrera: Well it's not optional if the others parameters are sent, but I'm just checking every parameter before to set it.

cab105 (approved):
Awesome, thanks!

PR #812, opened 2022-04-29
Bump Golang CI lint and rename functions to extensions

FranBarrera:


antoineco (approved):
While we're at it, maybe update `gotestsum` from 1.7.0 to 1.8.0 too?

PR #814, opened 2022-05-01
Remove `spec.public` attribute from Function

antoineco:
It is now redundant with [`spec.adapterOverrides.public`](https://github.com/triggermesh/triggermesh/blob/8eb93f5ee32c559bce379c1e4d74b6e0d57c8044/config/303-function.yaml#L60-L66).

### DIFF
--- a/config/samples/functions/python.yaml
+++ b/config/samples/functions/python.yaml
@@ -19,6 +19,8 @@ metadata:
 spec:
   runtime: python
   responseIsEvent: true
+  adapterOverrides:
+    public: true
   ceOverrides:
     extensions:
       type: io.triggermesh.python.sample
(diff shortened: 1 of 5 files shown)


sebgoa (approved):

tzununbekov (changes requested):
Please update [Python](https://github.com/triggermesh/triggermesh/blob/main/config/samples/functions/python.yaml#L22) and [Ruby](https://github.com/triggermesh/triggermesh/blob/main/config/samples/functions/ruby.yaml#L21) samples accordingly

config/samples/functions/python.yaml
```diff
@@ -19,7 +19,6 @@ metadata:
 spec:
   runtime: python
   responseIsEvent: true
```
tzununbekov, line 21: well, I not exactly what I meant 😅
let's have one sample function public 

```suggestion
  responseIsEvent: true
  adapterOverrides:
    public: true
```

tzununbekov (approved):

PR #816, opened 2022-05-02
Use TCP port 9092 for metrics in all components

### LINKED ISSUE #815: Inconsistent metrics endpoint across TriggerMesh components
The default TCP port of the Prometheus metrics exporter is `9090`. Knative Serving reserves this port (and `9091`) for exposing its own metrics via the queue-proxy container.

For those cases where a component is backed by a Knative Service, we expose our own metrics on TCP port `9092`, but in other cases we expose them on port `9090`.

I think this results in a poor experience from an (human) operator's perspective: how does one know what port to observe, and based on what criteria?

I suggest we simplify that observability experience by consistently exposing TriggerMesh's metrics on TCP port `9092`.

antoineco:
Closes #815

### DIFF
--- a/pkg/sources/reconciler/awss3source/adapter.go
+++ b/pkg/sources/reconciler/awss3source/adapter.go
@@ -64,7 +64,6 @@ func (r *Reconciler) BuildAdapter(src commonv1alpha1.Reconcilable, sinkURI *apis
 		resource.EnvVars(r.adapterCfg.configs.ToEnvVars()...),
 
 		resource.Port(healthPortName, 8080),
-		resource.Port("metrics", 9090),
 
 		resource.StartupProbe("/health", healthPortName),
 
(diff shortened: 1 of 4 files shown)


pkg/sources/reconciler/awss3source/adapter.go
```diff
@@ -64,7 +64,6 @@ func (r *Reconciler) BuildAdapter(src commonv1alpha1.Reconcilable, sinkURI *apis
 		resource.EnvVars(r.adapterCfg.configs.ToEnvVars()...),
 
 		resource.Port(healthPortName, 8080),
-		resource.Port("metrics", 9090),
```
antoineco, line 67: Looks like this source has been broken for quite some time... (ref. #579)

tzununbekov (approved):

Issue #592, opened 2022-02-23
OpenTelemetry's `cortex` exporter package is deprecated 


antoineco:
As per the linter warning in #566:

`pkg/targets/adapter/opentelemetrytarget/adapter.go#L31`

`SA1019: package go.opentelemetry.io/contrib/exporters/metric/cortex is deprecated: This package is no longer supported. Use the go.opentelemetry.io/otel/exporters/otlp/otlpmetric exporter as a replacement to send data to a collector which can then export with its PRW exporter. (staticcheck)`

antoineco:
For the time being, I annotated the import to prevent the linter from failing:

https://github.com/triggermesh/triggermesh/blob/3254475543785a95426405ce22ef6ca56b301fe9/pkg/targets/adapter/opentelemetrytarget/adapter.go#L31-L33

antoineco:
I gave this a shot just to understand the components involved a bit better.
The migration seems more complex than I anticipated.

My understanding is that none of the built-in exporters support the _Prometheus Remote Write API_ anymore (what the deprecation note refers to as _PRW_).

That same deprecation message suggests moving to the `go.opentelemetry.io/otel/exporters/otlp/otlpmetric` package, but this only allows sending to a _collector_, not directly to Prometheus / Cortex. That collector is expected to **export** those metrics in a second step. I don't quite get how this helps us, because the main purpose of TriggerMesh's OpenTelemetry Target is to send metrics to a Cortex endpoint, not to an intermediate collector.

Tagging @odacremolbap in case you haven't seen this issue yet.

antoineco:
Ahh, that makes sense.
I was under the impression that this exporter wasn't maintained anymore, but I wasn't looking at the [right docs](https://pkg.go.dev/github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter).

And `github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter` uses `go.opentelemetry.io/otel/exporters/otlp/otlpmetric` under the hood, I suppose?

cab105:
That's kinda what I thought as well given the docs I did find referenced a tag that was the last release to contain it (think it was https://github.com/open-telemetry/opentelemetry-collector/tree/release/v0.33.x/exporter/prometheuswriteexporter). 

It wasn't until I stumbled upon https://github.com/open-telemetry/opentelemetry-collector/issues/3474 that I found out what happened to it.

[...]

PR #828, opened 2022-05-03
Disable `otel/metric` updates in Dependabot

antoineco:
Until #592 is addressed, these updates will always fail, so we should save Dependabot the effort.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,7 @@
 # Dependabot configuration
 #
 # For more information, please refer to:
-#   https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically
+#   https://docs.github.com/en/code-security/dependabot/dependabot-version-updates
 
 version: 2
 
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,7 +1,7 @@
 # Dependabot configuration
 #
 # For more information, please refer to:
-#   https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically
+#   https://docs.github.com/en/code-security/dependabot/dependabot-version-updates
```
antoineco, line 4: Seemingly unrelated change, but this page was permanently redirected:

```
HTTP/2 301
cache-control: public, max-age=86400
content-type: text/plain; charset=utf-8
location: /en/code-security/dependabot/dependabot-version-updates
...

Moved Permanently. Redirecting to /en/code-security/dependabot/dependabot-version-updates
```

cab105 (approved):
Approving especially since #592 is considered low priority.

PR #835, opened 2022-05-04
Smoke test Eventing before running E2E tests

### LINKED ISSUE #801: Error creating Channel in E2E tests
Looks like we start the test suite although the Eventing webhook isn't up:

```
Error creating Channel: Internal error occurred: failed calling webhook "webhook.eventing.knative.dev": Post "[https://eventing-webhook.knative-eventing.svc:443/defaulting?timeout=10s](https://eventing-webhook.knative-eventing.svc/defaulting?timeout=10s)": dial tcp 10.96.10.65:443: connect: connection refused
```

Full logs: https://github.com/triggermesh/triggermesh/runs/6188607972?check_suite_focus=true

I was expecting the [`container-tools/kind-action`](https://github.com/container-tools/kind-action) GitHub Action to wait for all components to be ready, but their approach is [way too naive](https://github.com/container-tools/kind-action/blob/fdfd7e7ffb39e532f9de844dba8f1a29ae7e0f75/knative.sh#L166) to be considered reliable.

What I find puzzling is that, according to the logs of the workflow run shared above, the Eventing webhook _is_ up and running:

```
NAME                                    READY   STATUS    RESTARTS   AGE
eventing-controller-5b99f46b48-k6cgg    1/1     Running   0          34s
eventing-webhook-5489f465f7-dvgn5       1/1     Running   0          34s
imc-controller-7976d976f8-hjrdw         1/1     Running   0          32s
imc-dispatcher-74fdc75c4d-84fc7         1/1     Running   0          32s
mt-broker-controller-5dcbcd5d74-lgml7   1/1     Running   0          30s
mt-broker-filter-747856db96-qq8bk       1/1     Running   0          31s
mt-broker-ingress-6f7bb6b564-gscgc      1/1     Running   0          31s
```

antoineco:
Fixes #801 (hopefully)

The Knative Eventing webhook occasionally responds with "connection refused" although its Pod(s) reports "Ready", making Knative Eventing unusable in the E2E test suite.

This workflow step performs a smoke test of the Eventing installation by attempting a create/delete cycle of a basic `Channel` object. In case of failure, it deletes the Eventing webhook Pod(s) and retries.

Here is a sample run where I deliberately caused the smoke test to fail consistently:

```
......
pod "eventing-webhook-76b66cd56c-zv997" deleted
......
pod "eventing-webhook-76b66cd56c-ht8z2" deleted
......
Smoke test Channel creation failed: Error from server (AlreadyExists): error when creating "STDIN": channels.messaging.knative.dev "smoke-test" already exists
```

It is also possible that the first attempts fail, then the situation resolves itself, in which case the script returns without error.

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -50,6 +50,9 @@ jobs:
         # use for a container registry.
         registry: 'false'
 
+    - name: Smoke test Eventing webhook
+      run: .github/workflows/scripts/smoke-test-eventing.sh
+
     - name: Read KinD network subnet
       id: kind-subnet
       run: |
@@ -68,7 +71,7 @@ jobs:
         kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/v0.12.1/manifests/metallb.yaml
         kubectl -n metallb-system wait --timeout=1m --for=condition=Available deployments.apps/controller
 
-        cat <<'EOM' | kubectl create -f-
+        kubectl create -f - <<'EOM'
         apiVersion: v1
         kind: ConfigMap
         metadata:
--- a/.github/workflows/scripts/smoke-test-eventing.sh
+++ b/.github/workflows/scripts/smoke-test-eventing.sh
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+
+# The Knative Eventing webhook occasionally responds with "connection refused"
+# although its Pod(s) reports "Ready", making Knative Eventing unusable.
+# This script performs a smoke test of the Eventing installation by attempting
+# a create/delete cycle of a basic Channel object. In case of failure, it
+# deletes the Eventing webhook Pod(s) and retries.
+
+set -eu
+set -o pipefail
+
+function create_channel {
+	local err
+
+	err="$(kubectl create -f - 2>&1 >/dev/null <<-EOM
+	apiVersion: messaging.knative.dev/v1
+	kind: Channel
+	metadata:
+	  name: smoke-test
+	EOM
+	)"
+
+	echo "$err"
+}
+
+function delete_channel {
+	kubectl delete channels.messaging.knative.dev/smoke-test >/dev/null
+}
+
+declare -i succeeded=0
+declare last_err
+
+declare -i max_attempts=3
+declare -i was_retried
+
+# outer loop:
+#  - try to create Channel multiple times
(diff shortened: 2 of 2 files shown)


.github/workflows/e2e-tests.yaml
```diff
@@ -68,7 +71,7 @@ jobs:
         kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/v0.12.1/manifests/metallb.yaml
         kubectl -n metallb-system wait --timeout=1m --for=condition=Available deployments.apps/controller
 
-        cat <<'EOM' | kubectl create -f-
+        kubectl create -f- <<'EOM'
```
cab105, line 74: Shouldn't there be a space: `-f -` vs `-f-`?

.github/workflows/scripts/smoke-test-eventing.sh
```diff
@@ -0,0 +1,72 @@
...
+# a create/delete cycle of a basic Channel object. In case of failure, it
+# deletes the Eventing webhook Pod(s) and retries.
+
+set -eu
+set -o pipefail
+
+function create_channel {
+	local err
+
+	err="$(kubectl create -f- 2>&1 >/dev/null <<-EOM
```
cab105, line 15: Same here: `-f -`

cab105 (commented):
Left my nit on the `-` when used to indicate reading from stdin vs a filename.

.github/workflows/e2e-tests.yaml
antoineco: Both are working for short flags (single dash versions) but I'll add a space if that looks ambiguous.

antoineco: Addressed in [4edd5801..5d40b9c2](https://github.com/triggermesh/triggermesh/compare/4edd580109b4f6f20dcce598819e3b5a23f9be6b..5d40b9c2eba88353589183cf6b8fd1be281db44c)

.github/workflows/scripts/smoke-test-eventing.sh
antoineco: Addressed in [4edd5801..5d40b9c2](https://github.com/triggermesh/triggermesh/compare/4edd580109b4f6f20dcce598819e3b5a23f9be6b..5d40b9c2eba88353589183cf6b8fd1be281db44c)

Issue #847, opened 2022-05-09
Allow arbitrary data in `adapterOverrides` 


tzununbekov:
The idea is to make `adapterOverrides` structure in all our CRs accept arbitrary data by setting the `additionalProperties` attribute. As a result, we'll significantly reduce the CRD sizes and simplify the process of extending the adapter overrides. 
One important precondition is having [a validation webhook](https://github.com/triggermesh/triggermesh/issues/846) in place to check the overrides structure. Since override attributes are the same for all components, validation can be implemented as a common method for all webhooks.

The first mention is in https://github.com/triggermesh/triggermesh/pull/774#pullrequestreview-949427103

antoineco:
Honest opinion: that sounds like a step backwards to me.

OpenAPI is a solid and fairly standard way to describe - and express constraints for - an API. It makes things explicit and easy to document. It also reduces the dependency on webhooks. Also, you can do extra validations using a webhook even if you have an OpenAPI schema, both aren't incompatible.

I was under the impression that we already added everything we deemed relevant to `adapterOverrides`. What did I miss that justifies throwing part of our effort away? (the schema validation)

If the main goal is to _reduce effort_, why not simply invest a day in writing a Python script that can update all CRDs in batches? (like keep a YAML snippet of the `adapterOverrides` in the tool and let it do the structural replacements of `spec.adapterOverrides`) That would be so much more maintainable.

tzununbekov:
Okay, I see the point in the explicit structure of CRDs. The complexity is my main concern - one small adapter knob requires "thousand-lines PR", which makes this structure a little less appealing for wider usage. If you think that a Python script will both solve the complexity problem and still be in the best-practices field, then I'm fine with that. I'm also fine with leaving the decision and its implementation to the next person who will face the need to extend adapter overrides structure 🙂

antoineco:
I agree that _big and repetitive_ isn't appealing and feels broken. We need to address that and make the process easy.
OpenAPI itself supports external schema references using the `$ref` keyword, but [Kubernetes' implementation](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation) doesn't support it in CRD objects, which is the root of the problem here.

I'm not too worried about the _size_ of CRD files. The biggest ones are 16 kB currently, which is roughly a hundred times below the hard limit enforced by Kubernetes. Which is why I think the issue could be addressed with the right developers tooling.
[...]

PR #848, opened 2022-05-09
Add `crd-update` tool

antoineco:
An attempt at easing the process of propagating changes to `adapterOverrides` to all CRDs, as discussed in #847.

The Python tool takes a CRD file on stdin, outputs the updated version to stdout.

A batch script was also added for convenience, to run the Python tool on _all_ CRDs at once and update them on disk.

**Note:** The `ruamel.yaml` library changes the location of line breaks in `description` fields that contain long strings, such as URLs, because it seems pretty relaxed about the value of `yaml.width`. (I opened [this ticket](https://sourceforge.net/p/ruamel-yaml/tickets/427/))
It doesn't bother me personally, but I'm open to comments. I had to relax the YAMLlint rule for lines longer than 120 characters for that exact reason.


tzununbekov:
Nice! Do you think this tool is worth mentioning somewhere in the contribution guide?

antoineco:
Maybe, it's very specific to `adapterOverrides` so it depends whether we plan to add a lot more overrides in the future.

tzununbekov (approved):
Well, I looked at the contents of the `hack` directory and it seems that we're OK with not having documentation there. I hope we'll remember what is there and how to use it 🙂 
Leaving this up to you!

antoineco:
I'll add a README to the hack directory with the list of available utilities and their purpose. You're right, we should do a better job at documenting these otherwise we'll forget what they were created for.

PR #860, opened 2022-05-10
Document utilities under `hack/`

antoineco:
🎁 for @tzununbekov


tzununbekov (approved):
Feels like it is my birthday! Thanks, Antoine

PR #862, opened 2022-05-10
Add `spec.sink` enforcement to `crd-update.py`

antoineco:
I decided to sync `spec.sink` via the `crd-update.py` script, because a lot of components include this attribute.

The rationale is that users will want to see consistent descriptions for common attributes if we start generating our docs from OpenAPI schemas.

Relates to https://github.com/triggermesh/docs/issues/259

---

The script found 5 CRDs with schemas that slightly differ from the template.

I could also add other common attributes such as `spec.ceOverrides` or `spec.eventOptions`. Let me know what you think.

### DIFF
--- a/config/302-filter.yaml
+++ b/config/302-filter.yaml
@@ -122,26 +122,25 @@ spec:
                 - required: [uri]
                 properties:
                   ref:
-                    description: Ref points to an Addressable.
+                    description: Reference to an addressable Kubernetes object to be used as the destination of events.
                     type: object
                     properties:
                       apiVersion:
-                        description: API version of the referent.
                         type: string
                       kind:
-                        description: 'Kind of the referent. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
-                        type: string
-                      name:
-                        description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names'
                         type: string
                       namespace:
-                        description: 'Namespace of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/
-                          This is optional field, it gets defaulted to the object holding it if left out.'
                         type: string
+                      name:
+                        type: string
+                    required:
+                    - apiVersion
+                    - kind
+                    - name
                   uri:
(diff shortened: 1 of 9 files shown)


config/302-filter.yaml
```diff
@@ -122,26 +122,25 @@ spec:
...
                       kind:
-                        description: 'Kind of the referent. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
-                        type: string
-                      name:
-                        description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names'
                         type: string
                       namespace:
-                        description: 'Namespace of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/
-                          This is optional field, it gets defaulted to the object holding it if left out.'
                         type: string
```
antoineco, line 133: It looks like the 5 CRDs I mentioned had `description` attributes for `apiVersion`, `kind`, `namespace` and `name`.
I have a feeling that they are obvious and could be removed, but if we prefer to keep them I'll add them to the template and re-run the script. This will propagate descriptions to all source CRDs (and create a much larger PR).

config/302-filter.yaml
```diff
@@ -122,26 +122,25 @@ spec:
...
+                    required:
+                    - apiVersion
+                    - kind
+                    - name
                   uri:
-                    description: URI can be an absolute URL(non-empty scheme and non-empty host) pointing to the target or
-                      a relative URI. Relative URIs will be resolved using the base URI retrieved from Ref.
+                    description: URI to use as the destination of events.
                     type: string
+                    format: uri
```
antoineco, line 143: Likewise, the 5 aforementioned CRDs have a longer description for the `uri` attribute.
If we think it is useful I'll add it to the template and re-run the script. This will propagate the description to all source CRDs.

antoineco (commented):
Notes to reviewers

odacremolbap (approved):
imo the previous text at those CRDs did not add any value to the descriptions.

tzununbekov:
How about adding this script to the CI pipeline to check if CRDs are in sync? Just like the client's codegen check that we have now

antoineco:
I'm not against it. Since the script was an experiment I didn't want to it be too intrusive but if everybody is OK with having it as a default check let's do this.

PR #878, opened 2022-05-13
Use more recent k8s/kn versions in E2E CI workflow

### LINKED ISSUE #877: E2E workflow fails on `main` due to missing labels
I broke the E2E CI workflow in triggermesh/triggermesh#876, because `app.kubernetes.io/name` labels [weren't yet part of Knative](https://github.com/knative/serving/commit/92b5a121d19ca0d9f4aa0c73da3b20e30a5e8a62) in v1.0.0.

antoineco:
Fixes triggermesh/triggermesh#877

Tried in my fork to confirm that it works now: https://github.com/antoineco/triggermesh/runs/6420757703?check_suite_focus=true#step:10:443
Some tests failed because my fork doesn't have access to TriggerMesh's credentials, but the ones that don't require credentials passed.

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -42,9 +42,10 @@ jobs:
     - name: KinD Cluster
       uses: container-tools/kind-action@v1
       with:
-        knative_eventing: v1.0.0
-        knative_serving: v1.0.0
-        knative_kourier: v1.0.0
+        version: v0.13.0
+        knative_eventing: v1.4.0
+        knative_serving: v1.4.0
+        knative_kourier: v1.4.0
         # ko loads images directly into KinD's container runtime when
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-tests.yaml
```diff
@@ -142,7 +143,7 @@ jobs:
     # To prevent that, we delay the completion of the job until all namespaces
     # labeled "e2e-framework" have been finalized.
     - name: Wait for termination of E2E namespaces
-      if: always()
+      if: always() && steps.e2e.outcome != 'skipped'
```
antoineco, line 146: I added this small tweak to ensure we don't wait for e2e namespaces to terminate if tests didn't even run.

sameersbn (approved):

PR #882, opened 2022-05-13
Deduplicate generic reconciler code using Go generics

antoineco:
Second take on triggermesh/triggermesh#775.

Instead of using generic **types**, which felt very hacky and inflexible, I moved deduplicated code to generic **functions**, as recommended in https://go.dev/blog/when-generics:

-
    ```golang
    func serviceAccountOwners[T kmeta.OwnerRefable]
    ```
-
    ```golang
    func getOrCreateAdapter[T metav1.Object, L k8sLister[T], C k8sClient[T]]
    ```
-
    ```golang
    func findAdapter[T metav1.Object, L k8sLister[T]]
    ```
-
    ```golang
     func syncAdapter[T metav1.Object, C k8sClient[T]]
    ```

I'm _much_ more satisfied with this approach because it doesn't compromise extensibility. Whatever is reusable is inside generic functions, the rest remains inside the types' methods.

Concrete examples:
- The reconciler for Knative Services needs to [copy some annotations](https://github.com/triggermesh/triggermesh/blob/14ee343f8a5a01eff4373a2f2dd7d1d6c8686490/pkg/reconciler/reconcile.go#L403-L408), the one for Deployments doesn't.
- The reconciler for Knative Services propagates the availability of the adapter using [one method](https://github.com/triggermesh/triggermesh/blob/14ee343f8a5a01eff4373a2f2dd7d1d6c8686490/pkg/reconciler/reconcile.go#L355), the one for Deployments does it using [another](https://github.com/triggermesh/triggermesh/blob/14ee343f8a5a01eff4373a2f2dd7d1d6c8686490/pkg/reconciler/reconcile.go#L164).
- The reconciler for Knative Services [appends a HTTP route to the component's URL](https://github.com/triggermesh/triggermesh/blob/14ee343f8a5a01eff4373a2f2dd7d1d6c8686490/pkg/reconciler/reconcile.go#L356-L358) when it is multi-tenant, the one for Deployments doesn't do such thing.

In conclusion, I think we reap the benefits of Go generics by deduplicating code, while still remaining flexible enough to perform changes to one implementation or the other (which was impossible with the solution implemented in my previous PR).


tzununbekov (approved):
Same as before - I can read it, but for the overall conclusion my generics expertise is not sufficient. If there are no other opinions, we can merge now with a promise to review this code when we have more experience in generics 🙂

PR #883, opened 2022-05-13
Merge adapter builders into generic `AdapterBuilder` interface

antoineco:
Merges `AdapterDeploymentBuilder` and `AdapterServiceBuilder`.
These two interfaces had the same signatures, just with a different concrete type, which makes them a perfect candidate for a generic interface.

The reason why I didn't include this change in triggermesh/triggermesh#882 is because it induces a lot of boilerplate changes to all `adapter.go` and `reconciler_test.go` files, which can make things harder to review.

### DIFF
--- a/pkg/reconciler/reconcile.go
+++ b/pkg/reconciler/reconcile.go
@@ -53,20 +53,16 @@ var knativeServingAnnotations = []string{
 	serving.UpdaterAnnotation,
 }
 
-// AdapterDeploymentBuilder provides all the necessary information for building
-// a component's adapter backed by a Deployment.
-type AdapterDeploymentBuilder interface {
-	BuildAdapter(rcl v1alpha1.Reconcilable, sinkURI *apis.URL) (*appsv1.Deployment, error)
-}
-
-// AdapterServiceBuilder provides all the necessary information for building a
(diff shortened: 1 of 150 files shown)


pkg/reconciler/reconcile.go
```diff
@@ -51,53 +53,34 @@ var knativeServingAnnotations = []string{
...
+// component's adapter object.
+type AdapterBuilder[T metav1.Object] interface {
+	BuildAdapter(rcl v1alpha1.Reconcilable, sinkURI *apis.URL) (T, error)
 }
 
 // ReconcileAdapter reconciles a receive adapter for a component instance.
-func (r *GenericDeploymentReconciler[T, L]) ReconcileAdapter(ctx context.Context, ab AdapterDeploymentBuilder) reconciler.Event {
-	rcl := v1alpha1.ReconcilableFromContext(ctx)
+func (r *GenericDeploymentReconciler[T, L]) ReconcileAdapter(ctx context.Context,
+	ab AdapterBuilder[*appsv1.Deployment]) reconciler.Event {
```
antoineco, line 64: Here is the main change in this PR.

The rest is just boilerplate, and adjustments to common testing files.

tzununbekov (approved):

Issue #133, opened 2021-10-09
Remove empty `ClusterRoles` used for event sources


antoineco:
### Context

Currently, each event source _kind_ gets a dedicated `ServiceAccount` managed by their controller (e.g. `foosource-adapter` is used by all instances of `FooSource`).

Controllers try to bind these `ServiceAccounts` to a `ClusterRole` of the same name using a `RoleBinding`:

https://github.com/triggermesh/triggermesh/blob/b4b1d5dda573a2ffcc335272a9e05a64272589b1/pkg/sources/reconciler/common/reconcile.go#L423-L425

### Problem

Not all source adapters require permissions to communicate with Kubernetes. Yet, a missing `ClusterRole` would cause a binding failure:

```
Events:
  Type     Reason            Age    From                  Message
  ----     ------            ----   ----                  -------
  Warning  FailedRBACCreate  2m20s  foosource-controller  Failed to create adapter RoleBinding "foosource-adapter": clusterroles.rbac.authorization.k8s.io "foosource-adapter" not found
```

Because of that design, our release manifests include a bunch of `ClusterRoles`, most of which with [an empty set of rules](https://github.com/triggermesh/triggermesh/blob/76a084f7de64c386d17efada037832540adfb775/config/200-clusterroles.yaml#L469-L539), just to accommodate the minority of sources which do require access to Kubernetes (only 2).

### Proposed solution

Either add a new interface, or simply reuse the existing [`multiTenant` interface](https://github.com/triggermesh/triggermesh/blob/76a084f7de64c386d17efada037832540adfb775/pkg/apis/sources/v1alpha1/interfaces.go#L50-L59), to determine whether or not the controller should bind the `ServiceAccount` to a `ClusterRole`.

PR #885, opened 2022-05-16
Knative v0.31.0

tzununbekov:
In order to use ConfigMap watcher for adapters logging configuration hot-reload we need to update Knative to v0.31.0.

**[!]** This PR sets **Kubernetes** minimal required version to **v1.22** (https://github.com/knative/pkg/pull/2485)


antoineco (approved):

tzununbekov:
/hold

Controller will crash due to the k8s version requirement. I leave it up to @sameersbn to unhold when our clusters are ready.

antoineco:
(I'm writing this here because I don't know where else to write it)

One important note about enabling ConfigMap watchers in all components by default:
This requires explicitly granting components access to the Kubernetes API using a ClusterRole associated with their ServiceAccount (via a RoleBinding).

ServiceAccounts are dynamically created by the TriggerMesh controller in the [RBAC reconciler](https://github.com/triggermesh/triggermesh/blob/f68166fc231499099bd8f9898bbfb75b536bf4a8/pkg/reconciler/reconcile.go#L484), so that part is already covered.
However, the RoleBinding can only be created if the corresponding ClusterRole exists.

---

Prior to #133 (and corresponding PR) we used to maintain a bunch of empty ClusterRoles corresponding to each TriggerMesh component type, just so that we could [bind ServiceAccounts to something](https://github.com/triggermesh/triggermesh/blob/f68166fc231499099bd8f9898bbfb75b536bf4a8/pkg/reconciler/reconcile.go#L495). **This was tedious**.
Luckily, we isolated the minority of components which actually needed a ClusterRole (multi-tenant adapters), and got rid of the dozens of other ClusterRoles.

With this story, we might need to get back to the pre-#133 status of having one ClusterRole per component type, with an added twist: each of them needs permissions to [`list` and `watch` ConfigMaps](https://github.com/triggermesh/triggermesh/blob/f68166fc231499099bd8f9898bbfb75b536bf4a8/config/200-clusterroles.yaml#L483-L489).

Nothing complicated, but definitely maintenance heavy without proper YAML tooling in the long run. So before jumping head first, I would question whether _"I want to change the log level without restarting Pods"_ really is a strong enough incentive for this. 
I have a feeling that most users will be happy just setting `loglevel.foo-adapter: debug` centrally, inside [`triggermesh/config-logging`](https://github.com/knative/serving/blob/57766358ee2ea7d520b4483b4362348508c806ec/config/core/configmaps/logging.yaml#L70-L78). How many users expressed the will to repeatedly switch from one log level to another at runtime? Creating a new ConfigMap for each component just to be able to adjust the logger's config is also not that intuitive.

---

_Alternative to maintaining a bunch of identical ClusterRoles:_ ship a single `triggermesh-config-reader` ClusterRole with TriggerMesh and programmatically [bind all ServiceAccounts to it](https://github.com/triggermesh/triggermesh/blob/f68166fc231499099bd8f9898bbfb75b536bf4a8/pkg/reconciler/reconcile.go#L494-L495) in the RBAC reconciler (regardless of multi-tenancy).

tzununbekov:
It was mentioned in the "TriggerMesh Logging" document that config hot-reload is optional as switching between levels is expected to happen mostly in the integration development phase where pod restart is not a problem. I'm fine with leaving it out of the scope although I believe that the introduction of the `triggermesh-config-reader` role and triggermesh-specific configmap is probably a question of time.

sameersbn:
> Controller will crash due to the k8s version requirement. I leave it up to @sameersbn to unhold when our clusters are ready.

@tzununbekov @antoineco please note that staging has been updated to k8s 1.22 and knative 1.4.

antoineco:
> I believe that the introduction of the `triggermesh-config-reader` role and triggermesh-specific configmap is probably a question of time.

@tzununbekov @odacremolbap FYI once #924 gets merged _all_ receive adapters will automatically be granted read access to ConfigMaps, so components can start [injecting `ConfigMapWatcher` via the shared main](https://github.com/knative/eventing/blob/c3e13967001ff25584f791a28fd4d48b5e75b14c/cmd/mtping/main.go#L43-L54) and things will _just work™_.

antoineco (approved):
Re-approving, because something that good deserves several approvals.

PR #915, opened 2022-05-18
Watch Pod changes in Deployment reconciler to propagate their status quickly

### LINKED ISSUE #884: Subscribe to Kubernetes events from Pods in generic Deployment reconciler
Unlike Knative Services, Deployments don't bubble up the statuses of managed Pods (`status.containerStatuses`) to their own status conditions.

We already have [a workaround](https://github.com/triggermesh/triggermesh/blob/14ee343f8a5a01eff4373a2f2dd7d1d6c8686490/pkg/status/status.go#L47) in place for reading the statuses of individual Pods and propagating them to our objects (e.g. "MissingSecret", "BadContainerImage", "AppRuntimeFailure"), but currently these can take **several minutes** to propagate because we only have an event handler on Deployments, not Pods, and Deployments' statuses don't get updated on every Pod status change (a Deployment is either Ready or Failed, there is no granularity in "Failed" cases).

Potential drawbacks:

- Adding a Pod informer means caching all Pods, which will increase the memory usage of the TriggerMesh controller.
- Increased complexity in informer's event handler because the object to enqueue (TriggerMesh component instance) is the parent of a long chain of dependencies down to the Pod (`custom object` -> `deployment` -> `replicaSet` -> `pod`) which needs to be evaluated in reverse (like [`Impl.EnqueueControllerOf`](https://pkg.go.dev/knative.dev/pkg@v0.0.0-20220512013937-2d8305b2e59a/controller#Impl.EnqueueControllerOf) but recursive).

antoineco:
Closes triggermesh/triggermesh#884

Adds a Pod informer to the generic Deployment reconciler and attaches an event handler which, on status changes, finds the outermost ancestor of the Pod (TriggerMesh custom object) and enqueues it.
This results in a rapid propagation of Pods' state changes to their respective TriggerMesh object status.

This is pretty much what Knative Serving does with Services / Revisions. We do it here with Deployments.

> **Note**
> There may be 34 changed files in this PR, but 27 of them are this change to account for the new informer:
>
> ```patch
> -	TestControllerConstructor(t, NewController)
> +	TestControllerConstructor(t, NewController,
> +		// we expect "Pod" as an additional informer in this reconciler implementation
> +		ExpectExtraInformers(1),
> +	)
> ```

### Demo

I created a source with a reference to a secret that didn't exist, and the status of that source transitioned immediately with the different states of the Pod:

```console
$ kubectl get foosources.sources.triggermesh.io -w
NAME     READY   REASON               SINK
sample
sample   False   AdapterUnavailable   http://event-display.default.svc.cluster.local
sample   False   ContainerCreating    http://event-display.default.svc.cluster.local
sample   False   MissingSecret        http://event-display.default.svc.cluster.local
```

The same mechanism would report errors such as missing images, etc. Basically anything that can prevent a Pod from either starting or being created.

Prior to this change, the status would remain `AdapterUnavailable` for several minutes, until a watch eventually expired and the status got re-checked.
This gives users a much quicker feedback about what's going on with the adapter.

### DIFF
--- a/pkg/reconciler/base.go
+++ b/pkg/reconciler/base.go
@@ -24,6 +24,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	appsclientv1 "k8s.io/client-go/kubernetes/typed/apps/v1"
 	coreclientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
@@ -35,6 +36,8 @@ import (
 
 	k8sclient "knative.dev/pkg/client/injection/kube/client"
 	deploymentinformerv1 "knative.dev/pkg/client/injection/kube/informers/apps/v1/deployment"
+	replicasetinformerv1 "knative.dev/pkg/client/injection/kube/informers/apps/v1/replicaset"
+	podinformerv1 "knative.dev/pkg/client/injection/kube/informers/core/v1/pod"
 	sainformerv1 "knative.dev/pkg/client/injection/kube/informers/core/v1/serviceaccount"
 	rbinformerv1 "knative.dev/pkg/client/injection/kube/informers/rbac/v1/rolebinding"
 	"knative.dev/pkg/controller"
@@ -53,10 +56,10 @@ type GenericDeploymentReconciler[T kmeta.OwnerRefable, L Lister[T]] struct {
 	// URI resolver for sinks
 	SinkResolver *resolver.URIResolver
 	// API clients
-	Client    k8sClientGetter[*appsv1.Deployment, appsclientv1.DeploymentInterface]
-	PodClient func(namespace string) coreclientv1.PodInterface
+	Client k8sClientGetter[*appsv1.Deployment, appsclientv1.DeploymentInterface]
 	// objects listers
-	Lister func(namespace string) appslistersv1.DeploymentNamespaceLister
+	Lister    func(namespace string) appslistersv1.DeploymentNamespaceLister
+	PodLister func(namespace string) corelistersv1.PodNamespaceLister
 
 	*GenericRBACReconciler[T, L]
 }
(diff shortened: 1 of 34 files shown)


pkg/reconciler/base_test.go
```diff
@@ -107,6 +118,191 @@ func TestHasAdapterLabelsForType(t *testing.T) {
...
+		},
+	}
+
+	for n, tc := range testCases {
+		t.Run(n, func(t *testing.T) {
+			withK8sCli := func(ctx context.Context, _ *rest.Config) context.Context {
+				ctx, _ = fakek8sclient.With(ctx, tc.objects...)
+				return ctx
+			}
+			injection.Fake.RegisterClient(withK8sCli)
```
antoineco, line 282: @tzununbekov I remember we were struggling to implement a similar test in Knative Eventing (https://github.com/knative/eventing/pull/5880) and resorted to do something super complicated where we injected our own started informer to the context.

I turns out that registering a ClientSet with pre-populated objects is in fact very simple 🙃

tzununbekov: Nice! It would save me a lot of headache in the upstream issue.

pkg/reconciler/base.go
```diff
@@ -146,21 +149,31 @@ func NewGenericDeploymentReconciler[T kmeta.OwnerRefable, L Lister[T]](ctx conte
...
 	})
 
+	var outermostCtlrType T
+
+	podinformerv1.Get(ctx).Informer().AddEventHandler(cache.FilteringResourceEventHandler{
+		FilterFunc: adapterPodWithAncestorOfKind(ctx, outermostCtlrType),
+		Handler: controller.HandleAll(
+			podOutermostAncestorHandlerFn(ctx, adapterHandlerFn),
+		),
+	})
```
antoineco, line 175: This is where we handle changes to Pod statuses (or any other change to Pod objects):

- Filter out Pods that do no have an ancestor of type `T` (where `T` is the type of the component being reconciled by this specific reconciler instance).
- If it's a match, pass the ancestor to `adapterHandlerFn` (enqueue it for reconciliation).

pkg/reconciler/base.go
```diff
@@ -273,3 +286,120 @@ func EnqueueObjectsInNamespaceOf(inf cache.SharedInformer, resyncFn filteredGlob
...
+			ancestorCtlrRef.APIVersion == gvk.GroupVersion().String() &&
+			ancestorCtlrRef.Kind == gvk.Kind
+	}
+}
+
+// outermostAncestorControllerRef returns the outermost ancestor controller of
+// the given object.
+func outermostAncestorControllerRef(ctx context.Context, obj metav1.Object) *metav1.OwnerReference {
+	return resolveOutermostAncestorControllerRef(ctx, obj, schema.GroupVersionKind{})
+}
```
antoineco, line 320: `resolveOutmostAncestorControllerRef` is **recursive** (see its code right below).

However, for simplicity, because we don't want to deal with dynamic clients and prefer to look up objects in memory (through listers), the recursion is only able to look for an ancestor up that chain of dependencies: `pod` -> `replicaset` -> `deployment` -> `TriggerMesh component`.

I can't think about a case where we would need to recurse up some other object type, but if that ever becomes the case, the function can simply be extended via additional switch cases.

pkg/reconciler/base_test.go
```diff
@@ -107,6 +118,191 @@ func TestHasAdapterLabelsForType(t *testing.T) {
 	})
 }
 
+func TestOuterMostAncestorControllerRef(t *testing.T) {
+	const tNs = "test"
+
+	testCases := map[string]struct {
+		objects []runtime.Object
+		expect  *metav1.OwnerReference
+	}{
```
antoineco, line 127: I highly recommend checking those test cases in order to better understand how the ancestor look up works.

antoineco (commented):
Notes

odacremolbap (approved):

PR #916, opened 2022-05-20
Add Event Processing stat reporter to all components

antoineco:
Changes:

- Enable the Event Processing stat reporter in all components that are able to process events (flow, targets).
- Adjust adapter tests under `pkg/flow/adapter/` to play nice with metrics:
  - Reset global state between tests
  - Register views only once

> **Note**
> This PR doesn't include the logic for _using_ the reporter, which is component-specific and will be addressed in separate PRs to facilitate the review process and deliver value step by step.

### DIFF
--- a/pkg/flow/adapter/dataweavetransformation/adapter_test.go
+++ b/pkg/flow/adapter/dataweavetransformation/adapter_test.go
@@ -22,21 +22,24 @@ import (
 	"testing"
 	"time"
 
-	cloudevents "github.com/cloudevents/sdk-go/v2"
-	cetest "github.com/cloudevents/sdk-go/v2/client/test"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
-	"github.com/triggermesh/triggermesh/pkg/apis/flow/v1alpha1"
-	"github.com/triggermesh/triggermesh/pkg/metrics"
-	targetce "github.com/triggermesh/triggermesh/pkg/targets/adapter/cloudevents"
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	cetest "github.com/cloudevents/sdk-go/v2/client/test"
+
 	"knative.dev/eventing/pkg/adapter/v2"
 	adaptertest "knative.dev/eventing/pkg/adapter/v2/test"
 	logtesting "knative.dev/pkg/logging/testing"
+	"knative.dev/pkg/ptr"
+
+	"github.com/triggermesh/triggermesh/pkg/apis/flow/v1alpha1"
+	"github.com/triggermesh/triggermesh/pkg/metrics"
+	metricstesting "github.com/triggermesh/triggermesh/pkg/metrics/testing"
+	targetce "github.com/triggermesh/triggermesh/pkg/targets/adapter/cloudevents"
 )
 
 const (
-	tBridgeID               = "bride-abdc-0123"
 	tCloudEventID           = "ce-abcd-0123"
 	tCloudEventType         = "ce.test.type"
 	tCloudEventResponseType = "ce.test.type.response"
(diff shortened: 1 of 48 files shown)


pkg/flow/adapter/dataweavetransformation/adapter_test.go
```diff
@@ -301,36 +303,41 @@ func TestDataWeaveTransformationEvents(t *testing.T) {
...
-				r := recover()
-				switch {
-				case r == nil:
-					assert.Empty(t, tc.expectPanic, "Expected panic did not occur")
-				case tc.expectPanic == "":
-					assert.Fail(t, "Unexpected panic", r)
-				default:
-					assert.Contains(t, r, tc.expectPanic)
-				}
-			}()
```
antoineco, line 315: Was unused, none of these tests expect a panic.

pkg/flow/adapter/xmltojsontransformation/adapter_test.go
```diff
@@ -56,35 +58,40 @@ func TestSink(t *testing.T) {
...
-				assert.Equal(t, tXML1, string(body))
-				fmt.Fprintf(w, "OK")
-			}))
-			defer svr.Close()
-
-			env := &envAccessor{
-				EnvConfig: adapter.EnvConfig{
-					Component: tCloudEventSource,
-					Sink:      svr.URL,
-				},
```
antoineco, line 73: I rewrote this test entirely — based on a similar test inside `xslttransformation/adapter_tests.go` — because it didn't test anything at all.

The adapter code wasn't even involved. We just spawned a httptest server (not even a CloudEvent receiver) which had nothing to do with the adapter, and which was asserting that the received body was equal to some constant (???)

pkg/flow/adapter/xslttransformation/adapter_test.go
```diff
@@ -171,21 +220,6 @@ func TestXSLTTransformationEvents(t *testing.T) {
...
-
-			expectPanic: "XSLT validation error: failed to parse xsl",
-		},
-
-		"wrong configuration not providing XSLT": {
-			allowXSLTOverride: false,
-			xslt:              "",
-
-			expectPanic: "if XSLT cannot be overriden by CloudEvent payloads, configured XSLT cannot be empty",
-		},
```
antoineco, line 187: I relocated these two test cases to a new `TestNewTarget` test, because they were only testing for panic in the constructor, instead of performing assertions on events.

I think it's more clear that way because we test the constructor and the dispatcher separately (one doesn't influence the other in the production code).
Comments welcome.

pkg/metrics/metrics_test.go
```diff
@@ -130,25 +129,6 @@ func TestEventProcessingStatsReporter(t *testing.T) {
...
-	)
-
-	MustRegisterEventProcessingStatsView()
-
-	metricstest.AssertNoMetric(t,
-		"event_processing_success_count",
-		"event_processing_error_count",
-		"event_processing_latencies",
-	)
-}
```
antoineco, line 150: This helper is useful for potentially any adapter test, so I created a global test helper for it.

pkg/flow/adapter/jqtransformation/adapter.go
```diff
@@ -63,7 +66,9 @@ func NewAdapter(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClie
 		replier:  replier,
 		ceClient: ceClient,
 		logger:   logger,
-		mt:       mt,
+
+		mt: mt,
```
tzununbekov, line 70: here and in the rest of flow group components: why do we need metric tag `mt` in the adapter object?

tzununbekov (approved):

PR #923, opened 2022-05-24
Logger grooming

### LINKED ISSUE #780: Switch Bumblebee to Knative logger
As originally mentioned in https://github.com/triggermesh/triggermesh/pull/779#discussion_r857511278, we need to update Bumblebee's adapter to use Knative's logger instead of Go's std library

tzununbekov:
- Switched from Go's log package to injected zap.Logger,
- Removed excessive and useless output,
- Renamed `log` to `logger` to match common style.

Closes #780

### DIFF
--- a/pkg/flow/adapter/transformation/adapter.go
+++ b/pkg/flow/adapter/transformation/adapter.go
@@ -20,13 +20,14 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"log"
 	"strings"
 	"time"
 
 	cloudevents "github.com/cloudevents/sdk-go/v2"
+	"go.uber.org/zap"
 
 	pkgadapter "knative.dev/eventing/pkg/adapter/v2"
+	"knative.dev/pkg/logging"
 
 	"github.com/triggermesh/triggermesh/pkg/apis/flow"
 	"github.com/triggermesh/triggermesh/pkg/apis/flow/v1alpha1"
@@ -56,6 +57,7 @@ type adapter struct {
 	sink string
 
 	client cloudevents.Client
+	logger *zap.SugaredLogger
 }
 
 // ceContext represents CloudEvents context structure but with exported Extensions.
@@ -70,6 +72,8 @@ func NewEnvConfig() pkgadapter.EnvConfigAccessor {
 }
 
 func NewAdapter(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClient cloudevents.Client) pkgadapter.Adapter {
+	logger := logging.FromContext(ctx)
+
 	mt := &pkgadapter.MetricTag{
 		ResourceGroup: flow.TransformationResource.String(),
 		Namespace:     envAcc.GetNamespace(),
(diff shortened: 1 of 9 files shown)


pkg/flow/adapter/transformation/adapter.go
```diff
@@ -83,21 +87,21 @@ func NewAdapter(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClie
 	trnContext, trnData := []v1alpha1.Transform{}, []v1alpha1.Transform{}
 	err := json.Unmarshal([]byte(env.TransformationContext), &trnContext)
 	if err != nil {
-		log.Fatalf("Cannot unmarshal context transformation env variable: %v", err)
+		logger.Fatalf("Cannot unmarshal context transformation env variable: %v", err)
```
antoineco, line 90: In other components we often put the actual Go error in a separate field using

```go
logger.Panicw("What happened in clear text", zap.Error(err))
```

so the log ends up looking like

```json
{
  "msg": "What happened in clear text",
  "err": "the detail of the error"
}
```

It's not mandatory, just a suggestion to remain as consistent as possible.

pkg/sources/adapter/azureiothubsource/adapter.go
```diff
@@ -78,7 +77,7 @@ func NewAdapter(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClie
 		env.ConnectionString,
 	)
 	if err != nil {
-		log.Fatal(err)
+		logger.Fatal(err)
 	}
```
antoineco, line 81: Maybe provide some context to the error, like "Failed to obtain IoT client".

pkg/sources/adapter/httppollersource/adapter.go
```diff
@@ -82,7 +82,7 @@ func NewAdapter(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClie
 		httpRequest: httpRequest,
 
 		ceClient: ceClient,
-		logger:   logging.FromContext(ctx),
+		logger:   logger,
```
antoineco, line 85: Nice catch

antoineco (approved):

pkg/flow/adapter/transformation/adapter.go
tzununbekov: Yeah, this is a common issue across all our logs, I'll handle it in a separate PR

pkg/sources/adapter/azureiothubsource/adapter.go
tzununbekov: Done

pkg/sources/adapter/azureiothubsource/adapter.go
```diff
@@ -78,7 +77,7 @@ func NewAdapter(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClie
 		env.ConnectionString,
 	)
 	if err != nil {
-		log.Fatal(err)
+		logger.Fatal("Failed to obtain IoT client", zap.Error(err))
```
antoineco, line 80: This needed to be `Fatalw` to leverage the additional field 😉 
No big deal, maybe we'll notice it in the future and fix the format as part of a larger change.

tzununbekov: Oopsie. I'll fix that as a part of output format standardization PR.

PR #928, opened 2022-05-24
Loglevel overrides

tzununbekov:
In this PR:

- Knative Eventing updated to the latest version to include log level propagation fix: https://github.com/knative/eventing/pull/6391
- `-adapter` prefix removed from adapter component names for consistency. If we think that it was somehow useful, I can bring it back, we just need to have a common style there
- Sample `config-logging` configmap YAML added


cmd/alibabaosstarget-adapter/main.go
```diff
@@ -22,5 +22,5 @@ import (
 )
 
 func main() {
-	pkgadapter.Main("alibabaosstarget-adapter", alibabaosstarget.EnvAccessorCtor, alibabaosstarget.NewTarget)
+	pkgadapter.Main("alibabaosstarget", alibabaosstarget.EnvAccessorCtor, alibabaosstarget.NewTarget)
```
antoineco, line 25: I fully endorse this 👍
We had a lot of inconsistencies regarding whether the "-adapter" suffix was included so this is a nice touch IMO.

config/config-logging.yaml
```diff
@@ -0,0 +1,81 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-logging
+  namespace: triggermesh
+data:
+  _example: |
```
antoineco, line 21: How about having a real productive ConfigMap instead of this large example?
Historically people have been confused in Knative and many tried to edit the example instead of the real config, wondering why nothing was getting applied.
That's why I suggest having a real config.

If we are afraid that people are going to `kubectl apply` over their own change, let's move it to a sub-directory like `config/configmaps` or whatever.

antoineco (approved):

config/config-logging.yaml
```diff
@@ -0,0 +1,81 @@
...
+    loglevel.webhook: "info"
+    loglevel.activator: "info"
+    loglevel.hpaautoscaler: "info"
+    loglevel.net-certmanager-controller: "info"
+    loglevel.net-istio-controller: "info"
+
+    # Log level overrides for TriggerMesh components.
+    # Log level changes require manual reconciliation for adapter components.
+    # Component names are matching their folder names in /cmd directory,
+    # without "-adapter" suffix:
```
antoineco, line 76: Probably easier to say that the logger name is the lowercase component kind? `FooSource` -> `foosource`, easy and intuitive.

There are exceptions, like sources that share the same adapter as other sources, but I think the pattern applies in the majority of cases, and people don't have to come look into the triggermesh repo to figure out the correct value.

config/config-logging.yaml
```diff
@@ -0,0 +1,81 @@
...
+          "stacktraceKey": "stacktrace",
+          "lineEnding": "",
+          "levelEncoder": "",
+          "timeEncoder": "iso8601",
+          "durationEncoder": "",
+          "callerEncoder": ""
+        }
+      }
+
+    # Log level overrides
```
cab105, line 60: Stupid question time: I'm assuming this is part of the `_example` configuration that would be global in scope instead of associated with the `zap-logger-config`?

cab105 (approved):

config/config-logging.yaml
antoineco: There are two types of keys that are read by Knative from this configmap:

- `zap-logger-config`: the global logger config, applied by default to anything that reads from this configmap
- `<foo>.loglevel`: optional, allows overriding the logging level (and just the level) of all components named \<foo>

config/config-logging.yaml
tzununbekov: Ok, that makes sense.

config/config-logging.yaml
tzununbekov: Done

PR #929, opened 2022-05-24
Set a fixed number of concurrent builds in release job

antoineco:
Build threads are still being killed.

I don't know what is suddenly causing this in CircleCI because builds go through smoothly in GitHub Actions, for the time being I suggest reducing the concurrency further.

image


antoineco:
OK, that finally fixed it.
I'd be curious to know what is causing this sudden increase in memory usage during builds.

image

PR #936, opened 2022-05-25
Filter and Splitter path parsing fix

tzununbekov:
Tiny PR with:
- parse Splitter and Filter name from URL path,
- a bit of log format improvements.


tzununbekov:
If this is urgent, feel free to merge it. I'm adding tests, but it can be done in a follow-up PR

odacremolbap:
I would like to see this in the next release (about to happen) so yes, if it is fine with you let's merge now.

PR #938, opened 2022-05-25
Document IAM permissions required by the Amazon EventBridge source

antoineco:
End-to-end [failed due to missing permissions](https://github.com/triggermesh/triggermesh/runs/6591619517?check_suite_focus=true) ❌, so here is an updated version of our IAM policy that covers the permissions required to run the source.

Here is a [successful run after updating the policy](https://github.com/triggermesh/triggermesh/actions/runs/2384254587) ✅.

image

### DIFF
--- a/test/e2e/framework/aws/eventbridge/eventbridge.go
+++ b/test/e2e/framework/aws/eventbridge/eventbridge.go
@@ -56,7 +56,7 @@ func ForceDeleteEventBus(ebClient eventbridgeiface.EventBridgeAPI, eventBusName 
 // any associated rule should also be deleted.
 func deleteEventBus(ebClient eventbridgeiface.EventBridgeAPI, eventBusName string, force bool) {
 	if force {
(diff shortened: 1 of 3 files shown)


test/e2e/framework/aws/eventbridge/eventbridge.go
```diff
@@ -56,7 +56,7 @@ func ForceDeleteEventBus(ebClient eventbridgeiface.EventBridgeAPI, eventBusName
 // any associated rule should also be deleted.
 func deleteEventBus(ebClient eventbridgeiface.EventBridgeAPI, eventBusName string, force bool) {
 	if force {
-		removeAllRules(ebClient, eventBusName)
+		deleteAllRules(ebClient, eventBusName)
```
antoineco, line 59: I renamed this helper so that it better matches the name of the actual API call, which is `DeleteRule` (see screenshot above).

sameersbn (approved):

PR #941, opened 2022-05-25
Enable passing tracing config via env vars

antoineco:
Suppresses the warning seen in triggermesh/triggermesh#939.

```json
{"level":"warn","ts":"2022-05-25T16:26:56.651Z","logger":"awssqssource","caller":"v2/config.go:185","msg":"Tracing configuration is invalid, using the no-op default{error 26 0  empty json tracing config}","commit":"14d5989"}
```

I can't think about any good reason for not passing `K_TRACING_CONFIG` to adapters. In the worse case, they will simply ignore it.
When `config-tracing` doesn't exist in the `triggermesh` namespace, a default config with no backend is passed:

```yaml
- name: K_TRACING_CONFIG
  value: '{"backend":"","debug":"false","sample-rate":"0"}'
```

### DIFF
--- a/pkg/extensions/reconciler/function/controller.go
+++ b/pkg/extensions/reconciler/function/controller.go
@@ -46,7 +46,7 @@ func NewController(
 	// Calling envconfig.Process() with a prefix appends that prefix
 	// (uppercased) to the Go field name, e.g. MYTARGET_IMAGE.
 	adapterCfg := &adapterConfig{
-		obsConfig: source.WatchConfigurations(ctx, app, cmw, source.WithLogging, source.WithMetrics),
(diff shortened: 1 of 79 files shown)


antoineco:
By the way, this doesn't invalidate the configmap watcher story.
Once watchers are implemented, the same config will be read, just from a different location.

tzununbekov (approved):
Ah, I had to look into the `source.WatchConfigurations` to understand what's going on:
```
// If no option is passed, the ConfigWatcher observes ConfigMaps for
// logging, metrics and tracing.
```

antoineco:
@tzununbekov yes sir, sorry for the confusion:)

PR #944, opened 2022-05-27
Address warnings from AWS IAM policy editor

antoineco:
Addresses the following three warnings:

> The `kinesis:ListShards` action in your policy does not support resource-level permissions and requires you to choose **All resources**.

> The `sns:Unsubscribe` action in your policy does not support resource-level permissions and requires you to choose **All resources**.

> Specify **object** resource ARN for the `s3:DeleteObject` and 1 more action.

---

👉 Please note that the IAM console is showing a few other warnings which do not seem to make any sense.
For instance, it's asking us to specify **stream** resource ARNs for actions such as `kinesis:GetRecords`, but when I use the IAM wizard to address the "issue", it generates the exact same pattern as we already have 🤷 

Most likely a bug because our policy is split into multiple Sids, even for the same given service, whereas the wizard is only able to group all actions of the same service into a single Sid.

image

image

image


odacremolbap (approved):

odacremolbap (approved):
It looks like this one already includes https://github.com/triggermesh/triggermesh/pull/945
Mentioning here just in case there is something missing from that PR.

antoineco:
@odacremolbap the other way around, #945 includes this, but yes that was basically so I can conveniently open two PRs at the same time.

PR #948, opened 2022-05-31
release-notes.sh: read release highlights from tag annotation

sameersbn:
Use `git tag --annotate --cleanup=whitespace {TAG}` to create the release tag. Enter the notable changes in the tag annotation. These will be prepended to the autogenerated release notes


sameersbn:
> I didn't fully understand at which release step the `# comments` need to be removed, or if that is something we should take into account when creating PRs.

You may have noticed that while writing a commit message, git will add some helper lines that are commented out using the `#` character. These normally aren't added to the commit message, but the `--cleanup=whitespace` flag disables this behavior which means you need to explicitly remove these line. 

The reason I added this flag is so the we can add the `#` characters for the markdown formatting.

PR #950, opened 2022-06-02
Increase controller limits

odacremolbap:
Increase controller memory limits.

### DIFF
--- a/config/500-controller.yaml
+++ b/config/500-controller.yaml
@@ -42,10 +42,10 @@ spec:
         resources:
           requests:
             cpu: 50m
-            memory: 500Mi
+            memory: 150Mi
           limits:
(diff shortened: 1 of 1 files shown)


config/500-controller.yaml
```diff
@@ -42,7 +42,7 @@ spec:
         resources:
           requests:
             cpu: 50m
-            memory: 100Mi
+            memory: 500Mi
```
sameersbn, line 45: requests are greater than limits :p

sameersbn (changes requested):

config/500-controller.yaml
odacremolbap: I modified the wrong one :-1:

PR #960, opened 2022-06-06
Updates confluent target to allow two more authentication methods

JeffNeff:
This update allows users to utilize an additional two security additional security mechanisms: GSSAPI, and SCRAM-SHA-256.

### DIFF
--- a/pkg/targets/adapter/confluenttarget/adapter.go
+++ b/pkg/targets/adapter/confluenttarget/adapter.go
@@ -49,21 +49,69 @@ func NewTarget(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClien
 
 	env := envAcc.(*envAccessor)
 
-	kafkaClient, err := NewKafkaClient(&kafka.ConfigMap{
-		"bootstrap.servers":       env.BootstrapServers,
-		"sasl.username":           env.SASLUsername,
-		"sasl.password":           env.SASLPassword,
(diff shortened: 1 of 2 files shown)


pkg/targets/adapter/confluenttarget/adapter.go
```diff
@@ -160,7 +206,7 @@ func (a *confluentAdapter) dispatch(event cloudevents.Event) cloudevents.Result
 		return cloudevents.ResultNACK
 	}
 
-	a.logger.Debugf("Delivered message to topic %s [%d] at offset %v",
+	a.logger.Infof("Delivered message to topic %s [%d] at offset %v",
```
JeffNeff, line 209: yes, I meant to do this. Please do not ask me to put it back to debug because I hate when we log to Debug.

JeffNeff:
Please note... the added Kerberos method will... probably not work with our docker image... and i have no way of testing Kerberos. I will have to hand that off to @FranBarrera

pkg/targets/adapter/confluenttarget/adapter.go
tzununbekov: Could you elaborate on why do you "hate when we log to Debug"? Logging discussion was open for weeks and as a team we agreed that printing individual events on the default level is not only pointless from the user perspective, but also can be financially dangerous for the production deployments.

tzununbekov (changes requested):

PR #965, opened 2022-06-09
Support setting env-vars on adapters

### LINKED ISSUE #964: support setting environment variables on Sources
It would be nice if one could add a list of env vars to `Sources`, like the `AzureServiceBusTopicSource`, that way we can set env-flags like http_proxy. I guess this fits under `adapterOverrides` allow for an `env` object, to contain a list of `name/value` pairs.

davidkarlsen:
Fixes #964
Signed-off-by: David J. M. Karlsen <david@davidkarlsen.com>

### DIFF
--- a/config/300-awscloudwatchlogssource.yaml
+++ b/config/300-awscloudwatchlogssource.yaml
@@ -152,6 +152,16 @@ spec:
                 description: Kubernetes object parameters to apply on top of default adapter values.
                 type: object
                 properties:
+                  env:
+                    description: Adapter environment variables.
+                    type: array
+                    items:
+                      type: object
+                      properties:
+                        name:
+                          type: string
+                        value:
+                          type: string
                   resources:
                     description: Compute Resources required by the adapter. More info at https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
                     type: object
--- a/config/300-awscloudwatchsource.yaml
+++ b/config/300-awscloudwatchsource.yaml
@@ -208,6 +208,16 @@ spec:
                 description: Kubernetes object parameters to apply on top of default adapter values.
                 type: object
                 properties:
+                  env:
+                    description: Adapter environment variables.
+                    type: array
+                    items:
+                      type: object
+                      properties:
+                        name:
+                          type: string
+                        value:
+                          type: string
                   resources:
                     description: Compute Resources required by the adapter. More info at https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
                     type: object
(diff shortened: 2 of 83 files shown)


davidkarlsen:
Maybe you want some tests too? Can you point me to the right *_test.go files?

davidkarlsen:
> Tests for the `pkg/adapter` reside at https://github.com/triggermesh/triggermesh/blob/main/pkg/reconciler/base_test.go, but since we lack a decent coverage at that package, and to keep it simple for this one I would create a new test file `pkg/adapter/adapter_test.go` and add a `TestAdapterOverrideOptions` function that covers the env variables case.

Hmmm, then it won't have access to the method, as it is not exported, even though I declare the test in the same package `reconciler`

I added it to `base_test.go` instead, but the test is not good, as I could not find a good way to test the return, as it's functions/callbacks, and not value objects to test on, and I don't have a `Deployment` that the callbacks have been applied on (which is a lot more involved).

tzununbekov (changes requested):
@davidkarlsen thanks for the contribution!
CRDs need to be updated to make the new `adapterOverrides` attribute available in all resources. We have the script to do that, you just need to edit the `hack/crd-update/crd-update.py` file and then run `./hack/crd-update/update-all-crds.sh`.

```diff
diff --git a/hack/crd-update/crd-update.py b/hack/crd-update/crd-update.py
index 492828f8..7d055371 100755
--- a/hack/crd-update/crd-update.py
+++ b/hack/crd-update/crd-update.py
@@ -17,6 +17,16 @@ try:
           description: Kubernetes object parameters to apply on top of default adapter values.
           type: object
           properties:
+            env:
+              description: Adapter environment variables.
+              type: array
+              items:
+                type: object
+                properties:
+                  name:
+                    type: string
+                  value:
+                    type: string
             public:
               description: Adapter visibility scope.
               type: boolean
```

And yes, we should work on proper tests for overrides, but it should not be a blocker for this PR.

tzununbekov (approved):

davidkarlsen:
@tzununbekov do you publish snapshot images anywhere?

tzununbekov:
@davidkarlsen build is still in progress, when it is finished you can get the deployment manifests with the latest image references from our CI:
https://app.circleci.com/pipelines/github/triggermesh/triggermesh/2561/workflows/eef5a710-db90-4908-b707-e112fe61aeca/jobs/7072/artifacts

odacremolbap:
> @tzununbekov do you publish snapshot images anywhere?

Yes. 
The pipeline is currently creating and pushing the images.

https://app.circleci.com/pipelines/github/triggermesh/triggermesh/2561/workflows/eef5a710-db90-4908-b707-e112fe61aeca/jobs/7072

Issue #981, opened 2022-06-21
AWS SQS Source needs validation at spec.receiveOptions.visibilityTimeout


odacremolbap:
AWS SQS Source `spec.receiveOptions.visibilityTimeout` is expecting a (go formatted) duration.
Informing a non duration value in this field leads to issues listing AWS SQS Source objects at the controller, hence making it fail to reconcile any instance of it.

This is also a potential problem for other components mapping a string to a duration at the API.

The preferred solution would be using a regular expression at the CRD that complies with [this](https://pkg.go.dev/time#ParseDuration).

AmanSarraf:
Assign this issue to me

PR #982, opened 2022-06-24
Addressing the awsSQS time duration validation issue

AmanSarraf:
Closes issue #981


pkg/status/status.go
```diff
@@ -186,7 +186,7 @@ func isResourceMissingError(reason, msg string) (bool, string /*resource type*/)
 		if unicode.IsLetter(rune(msg[i-1])) {
 			// if we reached the beginning of the message, it means
 			// the type is at the beginning of the message
-			if i-1 == 0 {
+			if i == 0 {
```
JeffNeff, line 189: ?

JeffNeff, line 189:
```suggestion
			if i-1 == 0 {
```

PR #1003, opened 2022-06-29
Includes the capability to create the service bus client via wss 

JeffNeff:


JeffNeff:
Closing in favour of another implementation method.

PR #1017, opened 2022-07-06
Kafka Source & Kafka Target - Add tlsEnable parameter

FranBarrera:


cab105 (changes requested):
Thank you for adding in the `tlsEnable` component. The wait group implementation is not correct, and will result in buggy behavior.

pkg/sources/adapter/kafkasource/adapter.go
```diff
@@ -127,19 +127,31 @@ func (a *kafkasourceAdapter) Start(ctx context.Context) error {
...
 	}
 
-	wg := &sync.WaitGroup{}
-	wg.Add(1)
-
-	defer wg.Done()
-	for {
-		go func() {
+	e := make(chan error, 1)
+	go func() {
```
cab105, line 131: So if I understand this:
1. Start function is invoked when the sources starts up
2. Spawn a new thread
2a. Inside this new thread, loop continuously and process messages as they are received
3. Block until we receive a response from the new thread of `nil` or `err`

At this point, switching from wait groups to channels doesn't yield anything as this is still "single threaded," but with the caveat that a new thread has been spun up to process the messages. Normally, it'd be best to spawn the processing in a separate thread, and provide callback handle that can be used to stop processing of all other threads.

cab105: It is, but that's not what is implemented.

What @FranBarrera has is closer to:
```
start routine:
  fork:
    // child
    loop:
      err := dosomething()
      return err

  // parent
  block for response
  return result
```

In this case, the for loop is effectively ignored, and we're spawning a thread to do a single task, which I highly doubt is the intent.

config/300-kafkasource.yaml
```diff
@@ -233,6 +236,7 @@ spec:
                         type: boolean
                 required:
                 - saslEnable
+                - tlsEnable
```
cab105, line 239: Should be okay to make this optional. Just so long as `omitempty` is set on the type.

config/301-kafkatarget.yaml
```diff
@@ -233,6 +236,7 @@ spec:
                         type: boolean
                 required:
                 - saslEnable
+                - tlsEnable
```
cab105, line 239: Should be okay to make this optional. Just so long as omitempty is set on the type.

cab105 (approved):
Much better. Left a minor nit. Also, will you be rerunning `make codegen` to regenerate the client/deepcopy files?

Providing a conditional approval pending the above get resolved.

odacremolbap (approved):

PR #1027, opened 2022-07-12
Export MakeAppEnv functions

tzununbekov:
Mostly a batch update of `BuildAdapter` methods to export `makeAppEnv` function that extracts environment variables from TriggerMesh objects - I think this is the least intrusive way to enable re-use of our adapters in different platforms. Required by WIP testing tool.
No logic should have been changed, a pair of eyes may help find possible mechanic errors/typos.


tzununbekov:
For reviewers - see linked PR on how this change is supposed to be used. If there is another option to achieve the same result with less effect on the core, please comment here, and we can discuss it.

odacremolbap (approved):

PR #1028, opened 2022-07-12
[WebhookSource] add options for generating CE extended attributes

odacremolbap:
Enables WebhookSource to produce CloudEvents extended attributes based on path, method, host, querystrings and headers.

```yaml
kind: WebhookSource
metadata:
  name: mywebhook
spec:
  eventType: my.event
  eventSource: sample
  eventExtensionAttributes:
    from:
    - method
    - path
    - host
    - queries
    - headers
  sink:
    ref:
      apiVersion: eventing.knative.dev/v1
      kind: Broker
      name: default

```


pkg/sources/adapter/webhooksource/webhook.go
```diff
@@ -158,3 +197,38 @@ func healthCheckHandler(w http.ResponseWriter, _ *http.Request) {
...
+	}
+
+	if needsStripping {
+		stripped := []byte{}
+		for i := range name {
+			if (name[i] >= 'a' && name[i] <= 'z') || (name[i] >= '0' && name[i] <= '9') {
+				stripped = append(stripped, name[i])
+			}
+		}
+		name = string(stripped)
```
tzununbekov, line 221: Is this double loop logic about saving some memory write operations? If yes, what do you think about replacing it with a single loop?

```
...
	for i := 0; i < len(name); i++ {
		if !((name[i] >= 'a' && name[i] <= 'z') || (name[i] >= '0' && name[i] <= '9')) {
			name = name[:i] + name[i+1:]
			i--
		}
	}
...
```

pkg/sources/adapter/webhooksource/webhook.go
```diff
@@ -158,3 +197,38 @@ func healthCheckHandler(w http.ResponseWriter, _ *http.Request) {
...
+
+	// truncate if longer than 20 characters
+	if len(name) > 20 {
+		name = name[:20]
+	}
+
+	// data is a reserved element at CloudEvents
+	if name == "data" || name == "path" || name == "method" || name == "host" {
+		return "data0"
+	}
```
tzununbekov, line 232: So if one of the headers or query parameters will be named `data`/`path`/`method`/`name`/`host` it will be set as `data0` CE attribute? And if there are several of them (these reserved names are pretty common, IMO), they will override each other? Is there a way to use some prefix to avoid overriding?

pkg/sources/adapter/webhooksource/webhook.go
tzununbekov: So what about the single loop that I suggested? It does not write to the memory if there is no stripping needed, it does not allocate any additional memory nor create new variables.

pkg/sources/adapter/webhooksource/webhook.go
odacremolbap: You mean if a user sends something like?

```
http://myhost/my/path?data=0&data=1
```

In that case (and the same case for headers) we receive a single element named "data" that will be translated to "data0" containing an array of values. That array will be iterated here and it will be appended a new integer to each element

https://github.com/triggermesh/triggermesh/blob/49f55ad70898f626890eb0ca45ee2dcb9e5676ba/pkg/sources/adapter/webhooksource/webhook.go#L168-L173

There is a chance of overlapping that has nothing to do with this function if querystring and headers contain the same variable names (very unlikely if you ask me), and that is something to clarify at user docs, I did not find a way to solve that case.

pkg/sources/adapter/webhooksource/webhook.go
odacremolbap: That is what I (and the internals of go) want to avoid.
At each iteration a new array is allocated, no matter if a transformation is needed or not.

tzununbekov: Alright, benchmarking showed 3 times more allocations in a single loop (3 vs. 9). That's something new to me 🙂 
Scratch that thread, Pablo

tzununbekov: To get some closure, I've run more benchmarks for different input strings and with different loop combinations.

Current double loop:
```
BenchmarkShortStringSanitized-10        133044349                8.866 ns/op           0 B/op          0 allocs/op
BenchmarkShortStringUnsanitized-10      29006846                40.46 ns/op           16 B/op          2 allocs/op
BenchmarkLongStringSanitized-10         11087374               108.6 ns/op             0 B/op          0 allocs/op
BenchmarkLongStringUnsanitized-10        3717823               321.4 ns/op           328 B/op          6 allocs/op
```

Single loop from the first comment:
```
BenchmarkShortStringSanitized-10        226595140                5.210 ns/op           0 B/op          0 allocs/op
BenchmarkShortStringUnsanitized-10      36080911                31.57 ns/op           16 B/op          1 allocs/op
BenchmarkLongStringSanitized-10         17858637                67.07 ns/op            0 B/op          0 allocs/op
BenchmarkLongStringUnsanitized-10        1397174               867.7 ns/op          3312 B/op         27 allocs/op
```

This shows that the single loop approach does not handle well cases with the long input string (130 chars) where half of the symbols must be removed as not suitable, but for strings that are already valid it works better than the two loops.

However, I tested a third variation of the sanitizer loop:

```go
...
stripped := []byte{}
for i := 0; i < len(name); i++ {
	if !((name[i] >= 'a' && name[i] <= 'z') || (name[i] >= '0' && name[i] <= '9')) {
		stripped = append(stripped, name[i])
	}
}
...
```

and results showed that it has strengths of both options:

```
BenchmarkShortStringSanitized-10        163743864                7.169 ns/op           0 B/op          0 allocs/op
BenchmarkShortStringUnsanitized-10      48975343                23.26 ns/op            8 B/op          1 allocs/op
BenchmarkLongStringSanitized-10         13295133                89.41 ns/op            0 B/op          0 allocs/op
BenchmarkLongStringUnsanitized-10        6528620               181.4 ns/op           184 B/op          5 allocs/op
```

These results can be disregarded as the test subject is not expected to be the system's bottleneck, nor the central piece of the events routing.

pkg/sources/adapter/webhooksource/webhook.go
odacremolbap: https://github.com/triggermesh/triggermesh/pull/1028/commits/f789144ef77fcc583002f271e773906f89fbfc51 @tzununbekov ^^

tzununbekov (approved):

PR #1031, opened 2022-07-14
KafkaTarget - Add support to create topic

FranBarrera:
KafkaTarget will create the topic if it doesn't exists.

### DIFF
--- a/pkg/targets/adapter/kafkatarget/adapter.go
+++ b/pkg/targets/adapter/kafkatarget/adapter.go
@@ -124,8 +124,8 @@ func NewTarget(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClien
 		createTopicIfMissing:      env.CreateTopicIfMissing,
 		flushTimeout:              env.FlushOnExitTimeoutMillisecs,
 		topicTimeout:              env.CreateTopicTimeoutMillisecs,
-		newTopicPartitions:        env.NewTopicPartitions,
 		newTopicReplicationFactor: env.NewTopicReplicationFactor,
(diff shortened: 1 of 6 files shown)


pkg/targets/adapter/kafkatarget/adapter.go
```diff
@@ -106,20 +106,26 @@ func NewTarget(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClien
 		logger.Panicw("Error creating Sarama Client", err)
 	}
 
+	sac, err := sarama.NewClusterAdminFromClient(sc)
+	if err != nil {
+		logger.Panicw("Error creating Sarama Admin Client", err)
```
tzununbekov, line 111: Logging levels with `w` at the end (`Infow`, `Errorw`, `Panicw`, etc.) accept arguments in the key-value format, eg:
```
logger.Panicw("Error creating Sarama Admin Client", "error", err)
```
where `error` is the name of the key that will be added to the log entry:
```
{"level":"panic", ... "msg":"Error creating Sarama Admin Client","error":"Sarama error message", ...
```
If there is no key provided with the variable argument, the logger will spit the error:
```
{"level":"error",... "msg":"Ignored key without a value.","ignored":"Sarama error message",...
```

So the options here are either to add a key (eg. `zap.Error(err)`) or switch from `Panicw` to `Panicf` for example.

FranBarrera: Thanks! Fixed.

tzununbekov (approved):

PR #1038, opened 2022-07-20
Fixes a bad pointer ref in the Zendesk Target

JeffNeff:


pkg/targets/adapter/zendesktarget/adapter.go
```diff
@@ -139,14 +150,9 @@ func (a *zendeskAdapter) createTicket(ctx context.Context, event cloudevents.Eve
...
 
-	if it.Body == "" {
-		ticket.Comment.Body = event.String()
-	} else {
-		ticket.Comment.Body = it.Body
-	}
-
 	nT, err := a.zclient.CreateTicket(ctx, zendesk.Ticket(*ticket))
 	if err != nil {
+		fmt.Println(err)
```
odacremolbap, line 155: maybe use logger at the adapter instead of fmt.Println

odacremolbap (approved):
pre-aproving.

PTAL at the comment

Issue #1069, opened 2022-08-16
Reduce e2e tests execution time


tzununbekov:
The first candidate for optimization is `ko apply` command which takes most of the test execution time. We could re-use images that are built by the CircleCI and pushed to our registry.

tzununbekov:
Done. Github CI does not build its own images, instead, it reuses images pushed by Circle CI. It did not reduce overall execution time but it solved the timeout issue caused by the CI resources exhaustion.

PR #1083, opened 2022-08-22
Report e2e status

tzununbekov:
Report e2e test results even if workflow failed.
Part of #1069.


tzununbekov:
Self-approving. Critical CI fix.

PR #1099, opened 2022-08-25
export CommonObjectLables into function that can be optionally overriden

nkreiger:
the Triggermesh `Generic$ResourceReconciler` reconciler is a really nice use of generics to maintain custom k8s resources when developing a controller.

The resources it creates during the underlying adapter-receiver process are automatically stamped with the triggermesh common labels...this PR would expose those as a variable, maintaining current functionality, but allowing developers to leverage the triggermesh reconciler code to create/stamp resources with labels of their choice to decouple the underlying controllers.

Open to a better way to do this, threw it up quick!

### DIFF
--- a/pkg/reconciler/adapter.go
+++ b/pkg/reconciler/adapter.go
@@ -326,9 +326,9 @@ func OwnByServiceAccount(obj metav1.Object, owner *corev1.ServiceAccount) {
 	})
 }
 
-// CommonObjectLabels returns a set of labels which are always applied to
-// objects reconciled for the given component type.
-func CommonObjectLabels(o kmeta.OwnerRefable) labels.Set {
+// TMCommonObjectLabels returns a set of labels which are always applied to
+// triggermesh objects reconciled for the given component type.
+func TMCommonObjectLabels(o kmeta.OwnerRefable) labels.Set {
 	return labels.Set{
 		appNameLabel:      ComponentName(o),
 		appComponentLabel: componentAdapter,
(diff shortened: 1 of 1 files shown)


pkg/reconciler/adapter.go
```diff
@@ -326,9 +326,9 @@ func OwnByServiceAccount(obj metav1.Object, owner *corev1.ServiceAccount) {
 	})
 }
 
-// CommonObjectLabels returns a set of labels which are always applied to
-// objects reconciled for the given component type.
-func CommonObjectLabels(o kmeta.OwnerRefable) labels.Set {
+// TMCommonObjectLabels returns a set of labels which are always applied to
+// triggermesh objects reconciled for the given component type.
```
tzununbekov, line 330:
```suggestion
// TriggerMesh objects reconciled for the given component type.
```

pkg/reconciler/adapter.go
```diff
@@ -337,6 +337,10 @@ func CommonObjectLabels(o kmeta.OwnerRefable) labels.Set {
 	}
 }
 
+// CommonObjectLabels set of labels which are always applied to
+// resource objects reconciled for the given component type
```
tzununbekov, line 341:
```suggestion
// resource objects reconciled for the given component type.
```

tzununbekov (commented):
A couple of styling nits

tzununbekov (approved):

JeffNeff:
Merge?

Issue #1119, opened 2022-09-06
AWS S3 Target Does Not Properly Sanitize Blob Name


JeffNeff:
This line should be checked for bad returns. Including un-escaped characters.

https://github.com/triggermesh/triggermesh/blob/5da18fa55b7f8c85555907774697a46b47075897/pkg/targets/adapter/awss3target/adapter.go#L115

JeffNeff:
something like 
```
 escapedType :=  strings.ReplaceAll( event.Type() ,  `"` ,  `\"` )
 escapedType :=  strings.ReplaceAll( event.Type() ,  `/` ,  `//` )

 escapedSource :=  strings.ReplaceAll( event.Source() ,  `"` ,  `\"` )
 escapedSource:=  strings.ReplaceAll( event.Source() ,  `/` ,  `//` )
 ```
 
 Should probably work

PR #1120, opened 2022-09-08
Removes the Event Source From the File/Naming Structure

JeffNeff:
Previously we included the event source as part of the path that we place the file. We could spend some time figuring out a fun way to sanitize special characters out of the source.. but I personally think this is a poor design choice and we should just go with `event.Type() "/" + event.Time().String()`


tzununbekov:
What's the status of this? Can we close it?

jmcx:
Linked to https://github.com/triggermesh/triggermesh/issues/1119.
Self-assigning to look into it.

PR #1125, opened 2022-09-11
Creates a Containerized Development Environment for Developers

JeffNeff:
Try it out:

On github:
* Select `<>Code` in the top right of the repository
* Select `Codespaces`
* Select `Create Codespace`

Via vscode and Docker:
* Checkout the repo on this branch
* Open the repo in VSCode
* VSCode should see the dev container config and ask if you would like to work in the container. Select yes.

Notes:
* It seems that minikube is not working... that was something M$ set up so ill have to go back and see what is going on with that.. but I'm hesitant to put much work there because I think an environment like this is better off pointing to an external cluster. 
* I was installing AWS & Google cli's but i removed that logic because I thought it might be a bit over opinionated and it added lots of time to the build process. Im interested in thoughts there?!


tzununbekov:
What's the status of this? Can we close it?

tzununbekov:
Closing as stale.

PR #1129, opened 2022-09-16
feat(webhooksrc): Add ability to apply the cloud event ID to the webhook pass through payload

nkreiger:
This PR would allow the ability to passthrough a GUID from the original payload headers, to the inner cloud events payload that is injected into the eventing system.

Why?

The cloud event ID can be an important attribute to mark the start of a transaction (series of events across a system)...by default a unique ID is created automatically. However, if you wanted to use that ID to track the original event -> to the end of the link you would need to know that ID ahead of time.

Since the webhooksource is a resource for taking base JSON payloads and -> to cloud event form...and you can pass attributes, it would make sense I think that you would also be able to pass your own cloud event ID.

Thoughts?


pkg/sources/adapter/webhooksource/webhook.go
```diff
@@ -166,10 +166,15 @@ func (h *webhookHandler) handleAll(ctx context.Context) http.HandlerFunc {
 			if h.extensionAttributesFrom.headers {
 				for k, v := range r.Header {
 					// Prevent Authorization header from being added
-					// as a CloudEvent atribute
+					// as a CloudEvent attribute
 					if k == "Authorization" {
 						continue
 					}
+					if k == "Ce-Id" {
+						event.SetID(v[0])
```
JeffNeff, line 174: Verification that `v[0]` != nil is needed here.

JeffNeff: How do we verify that `v[0]` is unique within the scope of the producer?

JeffNeff (commented):
I understand what your going after here. However, I'm a bit... un-convinced that custom cloudevent id's are the way to go here. Couldnt the same filtering be done on a ce-extension, type, or id?

JeffNeff:
have a look at what Timur did with the MQ source here -> https://github.com/triggermesh/triggermesh/blob/ab7dd98275f7bbe122f3d0a53c80e7fd522cae29/pkg/sources/adapter/ibmmqsource/adapter.go#L109

He created a CE Extension for the `Correlation ID`. For this same use case.

JeffNeff:
I have to take a closer look at the code or ask @odacremolbap if its possible to add/configure CE extensions to the events.

odacremolbap:
Hi @nkreiger,

if you can control the incoming request to add the CE attributes as HTTP headers, would the CloudEventsSource be a better fit for this case?

https://docs.triggermesh.io/guides/cloudeventssource/

pkg/sources/adapter/webhooksource/webhook.go
```diff
@@ -70,7 +70,7 @@ func (h *webhookHandler) Start(ctx context.Context) error {
 	return runHandler(ctx, s)
 }
 
-// runHandler runs the HTTP event handler until ctx get cancelled.
+// runHandler runs the HTTP event handler until ctx gets cancelled.
```
sameersbn, line 73: please excuse me for adding this change. Was using this PR to debug the CI build

nkreiger:
> Hi @nkreiger,
> 
> if you can control the incoming request to add the CE attributes as HTTP headers, would the CloudEventsSource be a better fit for this case?
> 
> https://docs.triggermesh.io/guides/cloudeventssource/



> Hi @nkreiger,
> 
> if you can control the incoming request to add the CE attributes as HTTP headers, would the CloudEventsSource be a better fit for this case?
> 
> https://docs.triggermesh.io/guides/cloudeventssource/

@odacremolbap correct, this could also be achieved by using the direct cloud events source...however, the incoming event will not be in the CE form, those attributes (type, source) are declaratively driven through the webhooksource definition.

nkreiger:
> have a look at what Timur did with the MQ source here ->
> 
> https://github.com/triggermesh/triggermesh/blob/ab7dd98275f7bbe122f3d0a53c80e7fd522cae29/pkg/sources/adapter/ibmmqsource/adapter.go#L109
> 
> He created a CE Extension for the `Correlation ID`. For this same use case.

I had considered that, but this PR is mostly to extend the ability to test...If I use a separate field for the correlation in the CE extensions, then it adds a requirement of that field having to exist (which is nice about using the ID because its uniquely generated unless passed)...so by using the ID field I can keep everything the same, but for functional testing I can pass the ID in to perform comparison at the end of the test

JeffNeff (approved):
lgtm

sameersbn:
not sure whats up with circleci and this PR. 
@nkreiger could you please rebase this PR with the `main` branch?

nkreiger:
> not sure whats up with circleci and this PR. @nkreiger could you please rebase this PR with the `main` branch?

I tried, keeps saying up to date...do you have doc link for what commands you would like me to run?

sameersbn:
Can you try


```

git checkout main

git pull origin main

git checkout -

git rebase main

git push origin <branch> -f

```

sameersbn:
it's interesting that when I add a commit to this PR, circleci picks up the config.yaml and starts the build. But when you make a commit it fails to do so.. So sure what I need to configure at circle ci to fix this behavior.

Anyway the build is now triggered and completed should be cleared to be merged cc @tzununbekov @JeffNeff

sameersbn:
Try

```
git checkout main
git pull origin main
git checkout -
git rebase main
git push origin <branch> -f
```

nkreiger:
sorry about that y'all I got super busy, do you still need me to rebase it? I see its been closed?

PR #1178, opened 2022-10-19
AWS S3 - Allow to use IAM Role

FranBarrera:


odacremolbap (approved):
Been talking with Fran about this.

The code looks good and is backward compatible, so it does not break anything.
This is also a bit of a playground and it would be nice to do some manual tests on this but we don't have the cycles atm.

Since this is being requested by users we are going to merge it and let them test.

PR #1191, opened 2022-10-25
Azure Sources - Add support for AKS Managed Identity

FranBarrera:


odacremolbap (approved):
+1

As an improvement, can we add to the `auth` element at the CRD that when it is not present our components will try to use AKS managed identity?

PR #1211, opened 2022-11-09
config: removed unused permissions from webhook and controller clusterroles

sameersbn:
NOTE: Not yet tested on openshift

### DIFF
--- a/config/200-clusterrole-webhook.yaml
+++ b/config/200-clusterrole-webhook.yaml
@@ -87,7 +87,6 @@ rules:
   - secrets
   verbs:
   - get
-  - create
   - update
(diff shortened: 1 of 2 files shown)


tzununbekov:
I'm not a reviewer, but this does not seem right

odacremolbap (dismissed):

tzununbekov:
or how the webhook would create a certificate secrets. Maybe there are another roles created somewhere that I'm, not aware of

sameersbn:
> how controller would manage adapter deployments

You are right about the deployments. I missed testing that one. Had to restore `create/update` for deployments.

PR #1212, opened 2022-11-11
clusterroles: split clusterroles for fine-grained namespace access

sameersbn:
By default the `triggermesh-controller` has access to all namespaces of
the cluster and therefore can manage resources in any namespace.

This default may not be desirable for all users who which to explicitly
specify the namespaces that the triggermesh-controller can create
resources in.

In this PR, we have segregated the clusterrole in such a way that users
can achieve this fine grained access simply by deleting the
`triggermesh-controller` ClusterRoles and creating the following
RoleBinding in the namespaces that the controller should be able to
create resources in:

```
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: triggermesh-controller
  labels:
    app.kubernetes.io/part-of: triggermesh
subjects:
  - kind: ServiceAccount
    name: triggermesh-controller
    namespace: triggermesh
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: triggermesh-controller
```

### DIFF
--- a/config/200-clusterrole-webhook.yaml
+++ b/config/200-clusterrole-webhook.yaml
@@ -15,7 +15,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: triggermesh-webhook
+  name: triggermesh-webhook-namespaced
   labels:
     app.kubernetes.io/part-of: triggermesh
 
@@ -92,6 +92,16 @@ rules:
   - list
   - watch
 
+---
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: triggermesh-webhook
+  labels:
+    app.kubernetes.io/part-of: triggermesh
+
+rules:
 # Validation webhook gets system namespace to use it as an owner.
 - apiGroups:
   - ''
--- a/config/200-clusterroles.yaml
+++ b/config/200-clusterroles.yaml
@@ -15,6 +15,102 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
+  name: triggermesh-controller-watcher
+  labels:
+    app.kubernetes.io/part-of: triggermesh
+
+rules:
+# List and watch receive-adapters
+- apiGroups:
+  - apps
+  resources:
+  - deployments
+  verbs:
+  - list
+  - watch
+
+# List and watch reconciled TriggerMesh resources
+# +rbac-check
+- apiGroups:
+  - sources.triggermesh.io
+  resources:
+  - awscloudwatchlogssources
+  - awscloudwatchsources
+  - awscodecommitsources
+  - awscognitoidentitysources
+  - awscognitouserpoolsources
+  - awsdynamodbsources
+  - awseventbridgesources
+  - awskinesissources
+  - awsperformanceinsightssources
+  - awss3sources
+  - awssnssources
+  - awssqssources
+  - azureactivitylogssources
+  - azureblobstoragesources
+  - azureeventgridsources
+  - azureeventhubsources
+  - azureiothubsources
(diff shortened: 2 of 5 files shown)


odacremolbap (commented):
+1 after linter is happy.

config/200-clusterroles.yaml
```diff
@@ -12,6 +12,86 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: triggermesh-controller-watcher
```
sameersbn, line 18: the controller MUST HAVE clusterscope for these

sameersbn:
> +1 after linter is happy.

please take a look

config/200-clusterrole-webhook.yaml
```diff
@@ -15,7 +15,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: triggermesh-webhook
+  name: triggermesh-webhook-namespaced
```
tzununbekov, line 18: Does this role need to be a **Cluster**Role?

config/200-clusterroles.yaml
tzununbekov: Are you sure about this role? I'm kind of surprised by the set of resources - deployments, but not kservices, sources, but not transformations, etc.

tzununbekov (commented):
I couple of comments after a quick glance. I hope this change is being tested thoroughly because it has all chances to break a lot of things.

config/200-clusterrole-webhook.yaml
sameersbn: Well no, it does not have to be a `ClusterRole`. I'll change it

sameersbn:
> Are you sure about this role? I'm kind of surprised by the set of resources - deployments, but not kservices, sources, but not transformations, etc.

the `triggermesh-controller-watcher` has all the clusterscope permissions the controller was complaining about, so at the moment these are the once that ARE NEEDED for the controller to function. It's possible some more maybe required, but I haven't encountered any in my testing.  Do you think it would make more sense to give `list` and `watch` permissions for everything that's listed under the `triggermesh-controller` clusterrole?

tzununbekov:
I have a subjective feeling that restricting naturally-cluster-scoped controller in the namespaces is a bit of the special edge-case that no one ever requested before. Wouldn't it be more safe to keep tested-over-the-years version in the upstream and have "restricted" flavor (kustomization overlay?) in a separate... directory? If you feel lucky, I'll gladly leave it up to you.
cc @odacremolbap @sebgoa

config/200-clusterroles.yaml
tzununbekov: > the triggermesh-controller-watcher has all the clusterscope permissions the controller was complaining about, so at the moment these are the once that ARE NEEDED for the controller to function. It's possible some more maybe required, but there are the ones i've encountered in my testing. Do you think it would make more sense to give list and watch permissions for everything that's listed under the triggermesh-controller clusterrole?

It's always nice to be sure about our own product and changes that we make. RBAC that's created based on the experiment scares me a little.
@odacremolbap maybe you have some ideas why this role includes these resources only?

sameersbn:
Agreed. But note that the default will still remain the same, i.e.
clusterscoped.

odacremolbap:
I am with Timur here.

There is a need for kn services watchers, if the tests did not show it is because either that permission is being granted somewhere else, or because no kn components were tested.

Some components (specially multi-tenant) are granted roles via triggermesh-controller, it might happen that such scenarios are also not being tested.

Maybe a doc at this repo with the roles, cluster roles and bindings would shed some light, for us and for users that want to downscope.

PR #1223, opened 2022-11-22
Google Sources - Add support for Workload Identity

### LINKED ISSUE #1200: Support Workload Identity for authenticating GoogleCloud event sources
Similar to "[Support alternative auth methods such as Kube2IAM and IAM Roles for Service Accounts](https://github.com/triggermesh/triggermesh/issues/244)", we'd like to rely on GKE's "[Workload Identity](https://cloud.google.com/kubernetes-engine/docs/concepts/workload-identity)" feature to manage the GoogleCloud event sources' authentication to the corresponding GCP APIs.

FranBarrera:
Closes #1200

Steps to use:

A GKE cluster ready with Workload Identity (https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity)
Following the guide, it is needed to do some steps:

Allow the triggermesh service account to impersonate the IAM:
```
gcloud iam service-accounts add-iam-policy-binding my-sa@my-project.iam.gserviceaccount.com \
    --role roles/iam.workloadIdentityUser \
    --member "serviceAccount:my-project.svc.id.goog[triggermesh/triggermesh-controller]"
```
Also the service account for the specific source, ex: pubsub:
```
gcloud iam service-accounts add-iam-policy-binding my-sa@my-project.iam.gserviceaccount.com \
    --role roles/iam.workloadIdentityUser \
    --member "serviceAccount:my-project.svc.id.goog[my-namespace/googlecloudpubsubsource-adapter]"
```

Annotate the triggermesh-controller service account:
```
kubectl annotate serviceaccount triggermesh-controller \
    --namespace triggermesh \
    iam.gke.io/gcp-service-account=my-sa@my-project.iam.gserviceaccount.com
```

Finally add the following annotation to the Google source: ex: pubsub:
```
kind: GoogleCloudPubSubSource
metadata:
  name: sample
  annotations:
    iam.gke.io/gcp-service-account: my-sa@my-project.iam.gserviceaccount.com
spec:
  topic: projects/my-project/topics/my-topic

  sink:
    ref:
      apiVersion: eventing.knative.dev/v1
      kind: Broker
      name: default
```

### DIFF
--- a/pkg/reconciler/adapter.go
+++ b/pkg/reconciler/adapter.go
@@ -264,14 +264,18 @@ func commonAdapterKnServiceOptions(rcl v1alpha1.Reconcilable) []resource.ObjectO
 func newServiceAccount(ctx context.Context, rcl v1alpha1.Reconcilable, owners []kmeta.OwnerRefable) *corev1.ServiceAccount {
 	extraAnnotations := v1alpha1.ServiceAccountAnnotationsFromContext(ctx)
 
-	resource := resource.NewServiceAccount(rcl.GetNamespace(), ServiceAccountName(rcl),
+	opts := []resource.ObjectOption{
 		resource.Owners(owners...),
 		resource.Labels(CommonObjectLabels(rcl)),
+	}
+
+	for k, v := range extraAnnotations {
+		opts = append(opts, resource.Annotation(k, v))
+	}
+
+	return resource.NewServiceAccount(rcl.GetNamespace(), ServiceAccountName(rcl),
+		opts...,
 	)
-
-	resource.Annotations = extraAnnotations
-
-	return resource
 }
 
 // newConfigWatchRoleBinding returns a RoleBinding object that binds a ServiceAccount
--- a/pkg/reconciler/adapter.go
+++ b/pkg/reconciler/adapter.go
@@ -17,6 +17,7 @@ limitations under the License.
 package reconciler
 
 import (
+	"context"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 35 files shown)


pkg/reconciler/adapter.go
```diff
@@ -260,11 +261,17 @@ func commonAdapterKnServiceOptions(rcl v1alpha1.Reconcilable) []resource.ObjectO
...
+	extraAnnotations := v1alpha1.ServiceAccountAnnotationsFromContext(ctx)
+
+	resource := resource.NewServiceAccount(rcl.GetNamespace(), ServiceAccountName(rcl),
 		resource.Owners(owners...),
 		resource.Labels(CommonObjectLabels(rcl)),
 	)
+
+	resource.Annotations = extraAnnotations
+
+	return resource
```
odacremolbap, line 274: The package and the var is `called` resource, can you change the var to something like `res`.
Also, it would be nice if the ServiceAccount is fully created with options, including the annotations.

I haven't tested this, but you get the idea:

```suggestion
      opts := []resource.ObjectOption{}
       for k, v := range extraAnnotations {
		opts = append(opts, resource.Annotation(k,v)
       }
	res := resource.NewServiceAccount(rcl.GetNamespace(), ServiceAccountName(rcl),
		resource.Owners(owners...),
		resource.Labels(CommonObjectLabels(rcl)),
		opts
	)

	return res
```

odacremolbap (changes requested):

odacremolbap (approved):

PR #1225, opened 2022-11-22
Rename azureeventhubsource to azureeventhubssource

### LINKED ISSUE #1216: AzureEventHubSource => AzureEventHubsSource (hubs plural)

odacremolbap:
Closes #1216

This change will require that current users remove manually the old CRD from their clusters afterthe new version is installed

```console
kubectl delete crd azureeventhubsource
```


sebgoa:
we are moving to a plural form ?

odacremolbap:
> we are moving to a plural form ?

Yes, this is a backwards incompatible change, but:

- that is how the [product is named](https://azure.microsoft.com/en-us/products/event-hubs/#overview).
- the target uses the correct plural name.

FranBarrera (approved):

PR #1244, opened 2022-12-15
add proposed changes

nkreiger:
Signed-off-by: Noah Kreiger <noahkreiger@gmail.com>

This change proposes a switch to the logic in the tranformation, for handling a value that does NOT exist.

In a normal coding / average coding flow, you would not usually append a `nil` value to a path, or the value would not default to the variable name. I find this to be both confusing, and cause issues in practice where you may want to perform routing based on whether a value exists or not, without defining an extra filter, and 2 transformations, etc.

I didn't see any unit tests for this file?

Scenarios:

POST with

```json
{
    "detail": "noah",
    "source": "kreiger"
}
```

store -> $detail as $detail and $source as $source

add -> $detail.$source -> results in noah.kreiger

POST with

```json
{
    "detail": "noah"
}
```

store -> $detail as $detail and $source as $source

add -> $detail.$source -> results in noah

POST with

```json
{
    "detail": "noah",
    "source": ""
}
```

store -> $detail as $detail and $source as $source

add -> $detail.$source -> results in noah.

(still unsure if this should use the empty string, or treat it like a nil)

### DIFF
--- a/pkg/apis/flow/v1alpha1/transformation_types.go
+++ b/pkg/apis/flow/v1alpha1/transformation_types.go
@@ -70,8 +70,9 @@ type Transform struct {
 
 // Path is a key-value pair that represents JSON object path
 type Path struct {
-	Key   string `json:"key,omitempty"`
-	Value string `json:"value,omitempty"`
+	Key       string `json:"key,omitempty"`
+	Separator string `json:"separator,omitempty"`
+	Value     string `json:"value,omitempty"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/flow/adapter/transformation/pipeline.go
+++ b/pkg/flow/adapter/transformation/pipeline.go
@@ -30,6 +30,10 @@ import (
 	"github.com/triggermesh/triggermesh/pkg/flow/adapter/transformation/transformer/store"
 )
 
+const (
+	DefaultEventPathSeparator = "."
+)
+
 // Pipeline is a set of Transformations that are
 // sequentially applied to JSON data.
 type Pipeline struct {
@@ -60,7 +64,12 @@ func newPipeline(transformations []v1alpha1.Transform) (*Pipeline, error) {
 			return nil, fmt.Errorf("transformation %q not found", transformation.Operation)
 		}
 		for _, kv := range transformation.Paths {
-			pipeline = append(pipeline, operation.New(kv.Key, kv.Value))
+			sep := DefaultEventPathSeparator
+			if kv.Separator != "" {
+				sep = kv.Separator
+			}
+
+			pipeline = append(pipeline, operation.New(kv.Key, kv.Value, sep))
 		}
 	}
 
--- a/pkg/flow/adapter/transformation/transformer/add/add.go
+++ b/pkg/flow/adapter/transformation/transformer/add/add.go
@@ -99,23 +99,24 @@ func (a *Add) retrieveVariable(key string) interface{} {
 func (a *Add) composeValue() interface{} {
 	result := a.Value
 	for _, key := range a.variables.ListKeys() {
-		index := strings.Index(result, key)
-		if index == -1 {
-			continue
-		}
-		// return value if it exists, or properly nil for references to empty values
-		storedValue := a.retrieveVariable(key)
+		for strings.Index(result, key) != -1 {
+			index := strings.Index(result, key)
+			if index == -1 {
+				continue
+			}
+			// return value if it exists, or properly nil for references to empty values
+			storedValue := a.retrieveVariable(key)
 
-		if result == key {
-			return storedValue
-		}
+			if result == key {
+				return storedValue
+			}
 
-		// no variable assigned to key in multi-path assignment, remove from path
-		if storedValue == nil {
-			// result = fmt.Sprintf("%s", result[:index-1])
-			result = fmt.Sprintf("%s%s", result[:index], result[index+len(key):])
-		} else {
-			result = fmt.Sprintf("%s%v%s", result[:index], storedValue, result[index+len(key):])
+			// no variable assigned to key in multi-path assignment, remove from path
+			if storedValue == nil {
+				result = fmt.Sprintf("%s", result[:index-1])
+			} else {
+				result = fmt.Sprintf("%s%v%s", result[:index], storedValue, result[index+len(key):])
+			}
 		}
 	}
 	return result
--- a/pkg/flow/adapter/transformation/transformer/add/add.go
+++ b/pkg/flow/adapter/transformation/transformer/add/add.go
@@ -65,7 +65,7 @@ func (a *Add) InitStep() bool {
 }
 
 // New returns a new instance of Add object.
-func (a *Add) New(key, value string) transformer.Transformer {
+func (a *Add) New(key, value string, _ string) transformer.Transformer {
 	return &Add{
 		Path:  key,
 		Value: value,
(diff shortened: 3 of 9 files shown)


nkreiger:
@tzununbekov I think that is an improvement to the code, however, I'm not sure where you are going with the `:` delimiter, because the default split for store, parse, shift is `.`. So this change wouldn't have an effect on that, as far as I know.

With your logic, running something llike this:

add:
  key: $example
  value: $storeOne.$storeTwo.$storeThree
  
  where two and three do not exist in some cases.
  
  could result in something like this:
  `storeOne..$storeThree`
  
  which is not the response I believe we would want. That would not be the ideal source to route off of downstream.
  
  I have pushed up further changes, that moves the DefaultSeparator at the top. And allows the user to optionally override the separator variable. I added your changes to clean up the original code, and determined we don't nee the trim suffix, a simple string cut works fine.
  
take an object like this:

```json
{
   "scan": {
           "result": {
                     "status.success": 3
                    }
           }  
}
```

you cannot split by `.`, you would need to define your own separator that doesn't clash like `+` or `$$`

now you can like this:

```yaml
store:
  - key: $hold
    separator: +
    value: scan+result+status.success
    
add:
  - key: new.value
    value: $hold
```

the separator is scoped only to the object. I think this solves a lot of flexibility problems that would make the transformation useless.

nkreiger:
@tzununbekov I also noticed that it looks for only the first instance of a variable. Another corner case, but I pushed a fix for that.

e.g. `$eventKey.$eventKey` would result in `key.$eventKey` and now it will be `key.key`

tzununbekov:
@nkreiger custom separator and multi-variable fix are legit improvements, but the `add` behavior update still unsettles me. The main concern for me is that with this update `add` starts not only to add values but also does some conditional output changes that depend on hardcoded data (`.`). I understand that in some cases it may be  the solution needed, but if someone is not aware of this specific thing it may confuse a lot. For example, if I want to have a date string composed of variables, current behavior seems more predictable:

```yaml
add:
  - key: date
    value: $dd.$mm.$yy
```
which currently may result in either `12.12.22` if everything is defined or `12.12.$yy` (`$dd.12.22`, etc.) if something is not set.
With the proposed update the result may vary from `12.12.22` to any combination - `12.12`, `12.22`, `12`, `22`, which is an entirely different kind of output. And if I replace the `.` separator in date with `/`, then the `add` behavior changes to its old logic. Having this kind of operation logic switch based on the user input without the explicit declaration from the user brings the behavior ambiguity - one cannot predict the `add` output if there is `.` and $variable in it.


I have an idea of how the expected result may be achieved in a different way.

Input:
```json
{
    "detail": "noah",
    "source": "kreiger"
}
```

Transformation:
```yaml
store:
  - key: $detail
    value: detail
  - key: $source
    value: "."source  # value supports static data that will be added only if the path (source) is set
add:
  - key: detail
    value: $detail$source
```

The syntax for this update is TBD (e.g. `value: .+source`, `value: .(source)`, etc.), but the idea is simple - `store` supports `sprintf`-like behavior. If the path is not available, then the stored value is empty, and static data (`.` in this case) is ignored. Right now it seems that it provides the functionality you're looking for and also fits the behavior constraints I'm concerned about. If it suits you, we can split the changes - this PR is for the custom separator and multi-vars fix, and a different one for `store` update. I could also work on that one. What do you think?

nkreiger:
> @nkreiger custom separator and multi-variable fix are legit improvements, but the `add` behavior update still unsettles me. The main concern for me is that with this update `add` starts not only to add values but also does some conditional output changes that depend on hardcoded data (`.`). I understand that in some cases it may be the solution needed, but if someone is not aware of this specific thing it may confuse a lot. For example, if I want to have a date string composed of variables, current behavior seems more predictable:
> 
> ```yaml
> add:
>   - key: date
>     value: $dd.$mm.$yy
> ```
> 
> which currently may result in either `12.12.22` if everything is defined or `12.12.$yy` (`$dd.12.22`, etc.) if something is not set. With the proposed update the result may vary from `12.12.22` to any combination - `12.12`, `12.22`, `12`, `22`, which is an entirely different kind of output. And if I replace the `.` separator in date with `/`, then the `add` behavior changes to its old logic. Having this kind of operation logic switch based on the user input without the explicit declaration from the user brings the behavior ambiguity - one cannot predict the `add` output if there is `.` and $variable in it.
> 
> I have an idea of how the expected result may be achieved in a different way.
> 
> Input:
> 
> ```json
> {
>     "detail": "noah",
>     "source": "kreiger"
> }
> ```
> 
> Transformation:
> 
> ```yaml
> store:
>   - key: $detail
>     value: detail
>   - key: $source
>     value: "."source  # value supports static data that will be added only if the path (source) is set
> add:
>   - key: detail
>     value: $detail$source
> ```
> 
> The syntax for this update is TBD (e.g. `value: .+source`, `value: .(source)`, etc.), but the idea is simple - `store` supports `sprintf`-like behavior. If the path is not available, then the stored value is empty, and static data (`.` in this case) is ignored. Right now it seems that it provides the functionality you're looking for and also fits the behavior constraints I'm concerned about. If it suits you, we can split the changes - this PR is for the custom separator and multi-vars fix, and a different one for `store` update. I could also work on that one. What do you think?

@tzununbekov I think I'm almost there...

so you are saying, that the upstream `Store` functionality is changed to indicate when a value in the `Add` should be skipped because the stored value is empty? I would be fine with that, I can split that out of my PR.

tzununbekov:
@nkreiger the `store` operation is not changed yet, but the idea is correct. I see a couple of gaps in that model and I want to experiment with that first. I'll keep you posted on that topic. Meanwhile, let's bring in the separator and multi-var updates. Thanks!

tzununbekov:
So, I've played a bit with the `store` design that I suggested and I didn't really like it 😅 
Instead, what I have locally right now is a similar update, but for the `add` operation that seems to cover all cases.

I'm making `add` operation conditional by concluding stored variables in brackets. Here are a couple of examples:

```yaml
add:
  - key: detail
    value: noah(.$source)   # "noah" if $source is not set, "noah.kreiger" otherwise
```

```yaml
add:
  - key: message
    value: (hello $name)  # will either be "hello <Name>" or empty string if $name is not set
```

The idea is to add the value in round brackets only if the variable inside this brackets is set. Otherwise return empty string. 

This way we're not bound to the hardcoded `.` symbol and we can use any strings as a conditional output.
If no brackets is used in the value, then variables are added just like before, with the key as its default value. If we need to use brackets like a regular string, then they should be screened with `\` - I believe it is pretty reasonable constraint to have as a tradeoff for this new feature.

Let me know if this sounds ok to you @nkreiger. If it is, I can push it to the upstream next week.

tzununbekov:
I noticed how many conflicts with this PR I created in my latest transformation update so I took the responsibility and moved over all proposed fixes (multi-key add, custom separator) to the PR #1261 with discussed "conditional Add" feature. Noah is listed as a co-author there.
If I didn't miss anything and #1261 covers everything we discussed here, this PR can be safely closed as obsolete. Thanks @nkreiger once again for pointing out issues and making our components better.

nkreiger:
> I noticed how many conflicts with this PR I created in my latest transformation update so I took the responsibility and moved over all proposed fixes (multi-key add, custom separator) to the PR #1261 with discussed "conditional Add" feature. Noah is listed as a co-author there. If I didn't miss anything and #1261 covers everything we discussed here, this PR can be safely closed as obsolete. Thanks @nkreiger once again for pointing out issues and making our components better.

@tzununbekov thanks for doing that, I got pulled into other things. I will close this.

PR #1251, opened 2022-12-23
Add Kafka Source & Target Auth optional

### LINKED ISSUE #1248: Make AUTH optional for KafkaSource and KafkaTarget
Changes:
* `auth` should be documented as optional (is documented as required)
* `security mechanisms` is required, can we make it optional
* make sure our default settings match those of a clean install of Kafka or RedPanda

FranBarrera:
Closes #1248


odacremolbap (approved):

jmcx:
@odacremolbap @FranBarrera I get the impression that the autogenerated CRD reference docs for KafkaSource have not been updated to reflect the fact that `auth` is now optional: 

Sources_-_TriggerMesh

On the other hand, the KafkaTarget looks OK:

Targets_-_TriggerMesh

Also, when you look into the details of the `auth` object's spec, some things are optional and some aren't:

Sources_-_TriggerMesh

Any chance we can fix this before release?

PR #1252, opened 2022-12-24
fixing the links to the docs

sebgoa:
the installation guide was moved and I linked to the first guide in Guides because there is no index.


odacremolbap (approved):
This is better than 404, so lets merge.
Not sure if getting started should point to creating a DLS though ...

PR #1256, opened 2023-01-02
KafkaSource: parse certs

odacremolbap:
There is a bug parsing certs at the KafkaSource.

This PR:

- Uses the passed certs as an inline string, not a file path.
- Improves a little bit structure


sebgoa (approved):
should this be changed in the Kafka target as well ?

Issue #1275, opened 2023-01-10
Allow setting labels on adapters


snuggie12:
We have OPA policies in place for `Deployments`, `Pods`, and `Services`. I don't entirely follow the conversation in https://github.com/triggermesh/triggermesh/issues/705 but if this is already possible please let me know how.

If not possible it'd be great if that could be added. I had tried to add a label to `GoogleCloudPubSubSource` but I didn't see label transfer in the code like I did with the annotation for workload identity.

tzununbekov:
All correct, although thread in #705 is relevant to your question, the `adapterOverrides` [implementation](https://github.com/triggermesh/triggermesh/blob/v1.23.0/pkg/apis/common/v1alpha1/types.go#L39-L49) does not include adapter labels override at this moment. We could probably extend adapter overrides to include labels or just make adapter deployments inherit the parent object meta attributes.

sebgoa:
@tzununbekov related: https://github.com/triggermesh/triggermesh/issues/768

snuggie12:
I leave the implementation to you all obviously, but I do think the `adapterOverrides` is the more intuitive location.

I've recently started learning to program and I started with golang. I was reading some devs are against overloading `context` with too many values. Purely from wanting to learn, but if you do go with that solution for `labels` I'd like to hear thoughts on that statement.

tzununbekov:
"Label overrides" support added in #1278, and will be available in our next release. I'll post links to dev builds after we merge it with the upstream.

Regarding the `context`, I personally agree with the opinion that it should not be abused as a values storage, however, there are a lot of [examples](https://github.com/knative/pkg/search?q=context.WithValue) where this is practiced without any problems.

tzununbekov:
https://app.circleci.com/pipelines/github/triggermesh/triggermesh/3505/workflows/93ba47c9-99f0-4cc0-afc8-f2cc590112e2/jobs/9720/artifacts

@snuggie12 please let us know if this helped in your case

automaticgiant-ml:
yeah that's working, @tzununbekov 
thanks
do you have an expected release date?

tzununbekov:
Glad that it helped. I don't have the date for our next release, but you can join our slack where somebody from the team will help you out with that:
https://join.slack.com/t/triggermesh-community/shared_invite/zt-1kngevosm-MY7kqn9h6bT08hWh8PeltA

sebgoa:
[...]

PR #1278, opened 2023-01-11
Add "labels" to adapterOverrides

tzununbekov:
`adapterOverrides` extended with labels that should be passed to adapter pod meta:

```yaml
...
spec:
  adapterOverrides:
    labels:
      label1: value1
      label2: value2
```

Related to #1275

### DIFF
--- a/config/300-awscloudwatchlogssource.yaml
+++ b/config/300-awscloudwatchlogssource.yaml
@@ -156,6 +156,11 @@ spec:
                 description: Kubernetes object parameters to apply on top of default adapter values.
                 type: object
                 properties:
+                  labels:
+                    description: Adapter labels.
+                    type: object
+                    additionalProperties:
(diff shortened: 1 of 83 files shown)


odacremolbap:
looking good.

afaik OPA allows almost any pattern (let's focus on labels) on any object (let's focus on the ones we create).
The original issue mentions deployments, services and pods.

This change so far does it for pods, could we also add the labels to deployments and kn services?
idk if kn services propagate labels to their created kubernetes services, but if they do not that would be an issue at Kn.

tzununbekov:
Totally forgot why we needed these labels, thanks. Adding it to a parent object.

tzununbekov:
Just checked, k8s services are propagated with the labels from parent kn service

odacremolbap (approved):

PR #1286, opened 2023-01-19
MongoDB Target

JeffNeff:
This PR intends to add a mongo DB Target.


config/301-mongodbtarget.yaml
```diff
@@ -0,0 +1,214 @@
...
+          "schema": "https://raw.githubusercontent.com/triggermesh/triggermesh/main/schemas/io.triggermesh.mongodb.response.json",
+          "description": "the gneric response to all requests"
+        }
+      ]
+    registry.triggermesh.io/acceptedEventTypes: |
+      [
+        { "type":  "io.triggermesh.mongodb.insert" },
+        { "type":  "io.triggermesh.mongodb.query.kv" },
+        { "type":  "io.triggermesh.mongodb.update" },
+        { "type": "*" }
```
FranBarrera, line 37: I think this (*) should be deleted since the only accepted event types are the others.

JeffNeff: It does accept arbitrary events :)

JeffNeff:
Thanks for the review @FranBarrera

cmd/triggermesh-controller/main.go
```diff
@@ -166,6 +167,7 @@ func main() {
 		jiratarget.NewController,
 		kafkatarget.NewController,
 		logztarget.NewController,
+		mongodbtarget.NewController,
```
odacremolbap, line 170: nit: move a line down to keep alphabetic ordering

config/301-mongodbtarget.yaml
```diff
@@ -0,0 +1,214 @@
...
+  labels:
+    knative.dev/crd-install: 'true'
+    triggermesh.io/crd-install: 'true'
+    duck.knative.dev/addressable: 'true'
+  annotations:
+    registry.knative.dev/eventTypes: |
+      [
+        {
+          "type": "io.triggermesh.mongodb.response",
+          "schema": "https://raw.githubusercontent.com/triggermesh/triggermesh/main/schemas/io.triggermesh.mongodb.response.json",
```
odacremolbap, line 28: maybe @FranBarrera knows more than me about this.
If we have a commitment on keeping the schemas at this repo + location, this is right.

I was not aware of that (does not mean this is wrong).

odacremolbap: I see this is the case for some other schemas.
Not sure how it plays if we upgrade versions + schemas, but maybe it is better to have them than not.

odacremolbap: The schema file might be missing from this PR @JeffNeff ^^

config/samples/targets/mongodbtarget/200-target.yaml
```diff
@@ -0,0 +1,28 @@
...
+apiVersion: targets.triggermesh.io/v1alpha1
+kind: MongoDBTarget
+metadata:
+  name: mongo-db-target
+spec:
+  # The database to use by default if a `database` parameter is missing from the event payload(s)
+  defaultDatabase: dummydb
+  # The collection to use by default if a `database` parameter is missing from the event payload(s)
+  defaultCollection: dummycol
+  # A refrence to a kubernetes secret containing a connection string I.E mongodb+srv://<user>:<password>@<database_url>/?retryWrites=true&w=majority
```
odacremolbap, line 24: Just want to make sure that you are aware that anyone that can reach this target can write to any database/collection that they want and the connection allows. Could be considered a security concern for some.

docs/targets/mongodb.md
```diff
@@ -0,0 +1,146 @@
...
+```
+
+#### This type expects a JSON payload with the following properties:
+
+| Name  |  Type |  Comment |
+|---|---|---|
+| **database** | string | The name of the database.  |
+| **collection** | string | The value of the collection. |
+| **itemName** | string | This value will be used to assing a Key.  |
+| **mapStrVal** | map[string]string | This value will be used to assing a Value. |
```
odacremolbap, line 56: is `itemName` a better name than `key`?
same for `mapStrVal`, is that better than `value`?

docs/targets/mongodb.md
```diff
@@ -0,0 +1,146 @@
...
+
+| Name  |  Type |  Comment |
+|---|---|---|
+| **database** | string | The name of the database.  |
+| **collection** | string | The value of the collection. |
+| **itemName** | string | This value will be used to assing a Key.  |
+| **searchKey** | string | . |
+| **searchValue** | string | .  |
+| **updateKey** | string | .  |
+| **updateValue** | string |. |
```
odacremolbap, line 87: not sure if the dots mean they are self defined by the names. Just mentioning in case they are not.

config/301-mongodbtarget.yaml
JeffNeff: it is missing :) I need to generate it.

config/samples/targets/mongodbtarget/200-target.yaml
JeffNeff: or did I miss something?

docs/targets/mongodb.md
JeffNeff: these should be renamed. you are correct.

pkg/targets/adapter/mongodbtarget/adapter.go
```diff
@@ -0,0 +1,250 @@
...
+
+// NewTarget returns the adapter implementation.
+func NewTarget(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClient cloudevents.Client) pkgadapter.Adapter {
+	env := envAcc.(*envAccessor)
+	logger := logging.FromContext(ctx)
+	metrics.MustRegisterEventProcessingStatsView()
+
+	client, err := mongo.Connect(ctx, options.Client().ApplyURI(env.ServerURL))
+	if err != nil {
+		logger.Fatal("error connecting to mongodb", err)
```
odacremolbap, line 50: maybe this is a bit more friendly for log parsers

```suggestion
		logger.Panicw("error connecting to mongodb", zap.Error(err))
```

pkg/targets/adapter/mongodbtarget/adapter.go
```diff
@@ -0,0 +1,250 @@
...
+	sr       *metrics.EventProcessingStatsReporter
+}
+
+func (a *adapter) Start(ctx context.Context) error {
+	a.logger.Info("Starting MongoDB adapter")
+	return a.ceClient.StartReceiver(ctx, a.dispatch)
+}
+
+func (a *adapter) dispatch(event cloudevents.Event) (*cloudevents.Event, cloudevents.Result) {
+	ctx := context.Background()
```
odacremolbap, line 99: Not sure about why the context is created at 99.
I think you could use the signature that allows passing the request's context instead (unless there is a reason I am not getting)

```suggestion
func (a *adapter) dispatch(ctx context.Context, event cloudevents.Event) (*cloudevents.Event, cloudevents.Result) {
```

pkg/targets/adapter/mongodbtarget/adapter.go
```diff
@@ -0,0 +1,250 @@
...
+	db := a.defaultDatabase
+	if qpd.Collection != "" {
+		col = qpd.Collection
+	}
+	if qpd.Database != "" {
+		db = qpd.Database
+	}
+
+	collection := a.mclient.Database(db).Collection(col)
+	filterCursor, err := collection.Find(ctx, bson.M{qpd.Key: qpd.Value})
```
odacremolbap, line 171: just to make sure: this only allows to retrieve a doc that has a value at a certain key. If the event type needs to allow for other filters we would need to use a new event type.

pkg/targets/adapter/mongodbtarget/adapter.go
```diff
@@ -0,0 +1,250 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var episodesFiltered []bson.M
```
odacremolbap, line 176: Is episodes some sort of new mongo lang?

pkg/targets/adapter/mongodbtarget/adapter.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	responseEvent := cloudevents.NewEvent(cloudevents.VersionV1)
+	err = responseEvent.SetData(cloudevents.ApplicationJSON, episodesFiltered)
+	if err != nil {
+		return nil, err
+	}
+
+	responseEvent.SetType("io.triggermesh.mongodb.query.kv.result")
+	responseEvent.SetSource("io.triggermesh.mongodb")
```
odacremolbap, line 188: is the source supposed to be hard-coded?
whatever was discussed in the design meetings is ok with me, just asking.

config/samples/targets/mongodbtarget/200-target.yaml
odacremolbap: as long as the decision makers are aware that is fine with me.
just wanted to make sure you were aware.

odacremolbap:
@JeffNeff I think this is ok on the code-wise side, but maybe the context at the dispatch.
Nothing should explode if we merge. :heavy_check_mark:

JeffNeff:
@odacremolbap thank you for the review. I will make these adjustments.

pkg/targets/adapter/mongodbtarget/types.go
```diff
@@ -0,0 +1,48 @@
...
+*/
+
+package mongodbtarget
+
+// InsertPayload defines the expected data structure found at the event payload.
+type InsertPayload struct {
+	Database   string            `json:"database"`
+	Collection string            `json:"collection"`
+	StrValue   string            `json:"strValue"`
+	MapStrVal  map[string]string `json:"mapStrVal"`
```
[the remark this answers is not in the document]
JeffNeff: how about `JsonMessage` ?

pkg/targets/adapter/mongodbtarget/adapter.go
JeffNeff: mhmm. probably should not be. Im open to suggestions.

pkg/targets/adapter/mongodbtarget/adapter.go
JeffNeff: no. this should be revised.

pkg/targets/adapter/mongodbtarget/adapter.go
JeffNeff: This will retrieve all collections that contain a matching key:pair.

JeffNeff: and yes. more search(s) type(s) == more event types (at least thats how I see it )

pkg/targets/adapter/mongodbtarget/adapter.go
JeffNeff: I did not know you could do this and that the CE Receiver would automatically pass it. thank you.

pkg/targets/adapter/mongodbtarget/adapter.go
JeffNeff: xd :) good catch

pkg/apis/targets/v1alpha1/mongodb_lifecycle.go
```diff
@@ -0,0 +1,82 @@
...
+// GetEventTypes implements EventSource.
+func (t *MongoDBTarget) GetEventTypes() []string {
+	return []string{
+		EventTypeMongoDBStaticResponse,
+	}
+}
+
+// AsEventSource implements EventSource.
+func (t *MongoDBTarget) AsEventSource() string {
+	return t.Spec.Collection
```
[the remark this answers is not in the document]
JeffNeff: It should be required at the CRD.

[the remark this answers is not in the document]
JeffNeff: What is `database.collection` ?

pkg/targets/adapter/mongodbtarget/types.go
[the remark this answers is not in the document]
odacremolbap: is that a native go type? I don't know that one.
This is the one I am referring to
https://pkg.go.dev/encoding/json#RawMessage

pkg/targets/adapter/mongodbtarget/adapter.go
odacremolbap: I meant that probably the `EventTypeMongoDBQueryKV` will be re-used in the future if we want to provide other mongo filters as parameters to the Find collection.

Or maybe you are planning to create a new type/schema if new filters are required, not sure how it was designed. I don't mean to stop this PR from merging due to that, just want to make sure that those questions were discussed.

pkg/targets/adapter/mongodbtarget/types.go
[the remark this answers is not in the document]
JeffNeff: ahh. okay.

pkg/targets/adapter/mongodbtarget/adapter.go
JeffNeff: I honestly do not know... This was not discussed or has been thought out completely to be honest.

pkg/targets/adapter/mongodbtarget/adapter.go
odacremolbap: The component name + the collection where the data comes from would be nice when available.

schemas/io.triggermesh.mongodb.response.json
```diff
@@ -0,0 +1,64 @@
...
+                    "default": "",
+                    "examples": [
+                        "63c829397c2fdbfebdd93883"
+                    ]
+                }
+            }
+        }
+    },
+    "additionalProperties": true
+}
```
odacremolbap, line 64: nit: add blank at the end of this file

odacremolbap (approved):
@JeffNeff I am pre-approving

there is still an issue that might be important regarding the source attribute of returned events, for your consideration.

merge when you consider.

pkg/targets/adapter/mongodbtarget/types.go
```diff
@@ -0,0 +1,50 @@
...
+package mongodbtarget
+
+import "encoding/json"
+
+// InsertPayload defines the expected data structure found at the "io.triggermesh.mongodb.insert" payload.
+type InsertPayload struct {
+	Database    string          `json:"database"`
+	Collection  string          `json:"collection"`
+	Key         string          `json:"key"`
+	JSONMessage json.RawMessage `json:"JSONMessage"`
```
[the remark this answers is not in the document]
JeffNeff: I think this looks a lot nicer!

pkg/targets/adapter/mongodbtarget/adapter.go
JeffNeff: ok! Lets do this!

JeffNeff: I didnt know how to access the component name.. so i went with 
```
responseEvent.SetSource(fmt.Sprintf("%s/%s", a.defaultDatabase, a.defaultCollection))
```

PR #1308, opened 2023-02-02
fix: EventBridge rule ARN differs for default bus

### LINKED ISSUE #966: EventBridge Source sets non default bus
The eventbridge source creates a non-default event bus, but configures a rule which is supposed to get all AWS account events.

That rule does not work all account events get sent to the default bus.

I am still experimenting to see what would be the best default behavior

### LINKED ISSUE #1306: AWSEventBridgeSource - Error getting current targets for rule
Seeing this error when creating a new AWSEventBridgeSource

> Error getting current targets for rule "rule/io_triggermesh_awseventbridgesources-3021178537": ValidationException: 1 validation error detected: Value 'rule/io_triggermesh_awseventbridgesources-3021178537' at 'rule' failed to satisfy constraint: Member must satisfy regular expression pattern: [\.\-_A-Za-z0-9]+

I was able to replicate the error message by running `aws events list-targets-by-rule --rule "rule/io_triggermesh_awseventbridgesources-3021178537"`.

The error message went away when running `aws events list-targets-by-rule --rule "io_triggermesh_awseventbridgesources-3021178537"`, which suggests that the "rule/" at the start of the string is the cause.

antoineco:
Fixes #966
Fixes #1306 

cc. @sebgoa

### DIFF
--- a/pkg/sources/reconciler/awseventbridgesource/eventbus.go
+++ b/pkg/sources/reconciler/awseventbridgesource/eventbus.go
@@ -56,6 +56,8 @@ import (
 const (
 	eventbusARNResourcePrefix = "event-bus/"
 	ruleARNResourcePrefix     = "rule/"
+
+	eventbusNameDefault = "default"
 )
 
 const awsTagOwner = "owned-by"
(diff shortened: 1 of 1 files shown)


sebgoa (approved):
thanks for the investigation

PR #1309, opened 2023-02-02
feat(src/sqs): customizable serviceAccount name

antoineco:
### Context

Users have been expressing concerns regarding the usage of the IAM role auth feature of AWS sources (https://github.com/triggermesh/triggermesh/pull/248#issuecomment-1413980675).

Per Amazon's design, users must create one _IAM Role_ and one _Trust Relationship_ **per Service Account** (hence, per source instance). This is quite a tedious process, and for some people it is just fine to use a shared serviceAccount for selected sources.

### Proposal

This PR introduces an **alpha** feature  — currently only enabled in the event source for Amazon SQS (`AWSSQSSource` kind) — which allows overriding the name of the service account by annotating the source object with `alpha.triggermesh.io/aws-iam-service-account: <service account name>`.

I have no intention to introduce API changes here. An annotation is easier to test and less intrusive.
If it works, implementing this "properly" inside CRDs will be trivial.

### Demo

```yaml
kind: AWSSQSSource
metadata:
  name: sample1
  annotations:
    alpha.triggermesh.io/aws-iam-service-account: my-shared-serviceaccount  # arbitrary name
spec:
  arn: arn:aws:sqs:eu-central-1:123456789012:queue1
  auth:
    iamRole: arn:aws:iam::123456789012:role/foo  # MUST be identical in sample1 and sample2
  sink:
    ref:
      apiVersion: v1
      kind: Service
      name: event-display

---

kind: AWSSQSSource
metadata:
  name: sample2
  annotations:
    alpha.triggermesh.io/aws-iam-service-account: my-shared-serviceaccount  # arbitrary name
spec:
  arn: arn:aws:sqs:eu-central-1:123456789012:queue2
  auth:
    iamRole: arn:aws:iam::123456789012:role/foo  # MUST be identical in sample1 and sample2
  sink:
    ref:
      apiVersion: v1
      kind: Service
      name: event-display
```

```console
$ kubectl get serviceaccounts
NAME                       SECRETS   AGE
default                    1         13d
my-shared-serviceaccount   1         28s
```

```console
$ kubectl get deployments
NAME                             READY   UP-TO-DATE   AVAILABLE   AGE
awssqssource-sample1             1/1     1            1           3m56s
awssqssource-sample2             1/1     1            1           3m56s
event-display-00001-deployment   1/1     1            1           3m58s
```

```console
$ kubectl get deployments -o jsonpath='{.items[*].spec.template.spec.serviceAccount}'
my-shared-serviceaccount my-shared-serviceaccount
```

### Caveats

- There is no check whatsoever whether the serviceAccount already exists, is used by someone else, etc.
- There is no check whatsoever whether all sources that share a given service account are using the same iamRole value. If not, the iamRole will always be updated to the last reconciled source that uses this serviceAccount.
- The `ownerReferences` of the serviceAccount will always be updated to the last reconciled source that uses this serviceAccount.

Because this is an **alpha**, undocumented feature, I believe that none of the above caveats should prevent its testing.

cc. @sebgoa

### DIFF
--- a/pkg/apis/common/v1alpha1/aws_types.go
+++ b/pkg/apis/common/v1alpha1/aws_types.go
@@ -72,3 +72,29 @@ func AwsIamRoleAnnotation(iamRole apis.ARN) resource.ServiceAccountOption {
 		metav1.SetMetaDataAnnotation(&sa.ObjectMeta, annotationEksIAMRole, iamRole.String())
 	}
 }
+
+const annotationAlphaIamRoleServiceAccount = "alpha.triggermesh.io/aws-iam-service-account"
+
+// AlphaAwsIamRoleCustomServiceAccountName returns a functional option that
+// sets a custom name on a ServiceAccount, if requested by the component.
+// Alpha feature, only relevant for components with IAM Role authentication.
+func AlphaAwsIamRoleCustomServiceAccountName(r Reconcilable) resource.ServiceAccountOption {
+	if name := AlphaCustomServiceAccountName(r); name != "" {
+		return func(sa *corev1.ServiceAccount) {
+			sa.SetName(name)
+		}
+	}
+
+	return func(*corev1.ServiceAccount) {}
+}
+
+// AlphaCustomServiceAccountName returns a custom name for a ServiceAccount, if
+// requested by the component.
+// Alpha feature, only relevant for components with IAM Role authentication.
+func AlphaCustomServiceAccountName(r Reconcilable) string {
+	if WantsOwnServiceAccount(r) {
+		return r.GetAnnotations()[annotationAlphaIamRoleServiceAccount]
+	}
+
+	return ""
+}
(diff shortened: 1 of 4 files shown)


smittnacht:
Hi may I ask why this PR has been closed? For me this looks like a helpful feature and a great contribution.

sebgoa:
cc/ @tzununbekov can you comment ?

tzununbekov:
We delayed the merge because we didn't really want to add features through the objects annotations rather than CRD parameters - once annotation introduced, we should keep it mind for compatibility and consistency in future versions. However, as Antoine noted, it should be easy to add as a part of the object's specification. I'll take a look at it.

smittnacht:
Yes I can fully understand these points. It is always hard to change an interface description, even if it was declared as alpha.

Another question regarding this change: Currently the creation of the ServiceAccount is always controlled by TriggerMesh. Would it be possible to reference an existing ServiceAccount (created externally) and completely disable the creation of the ServiceAccount in TriggerMesh.

antoineco:
Closing the PR was my mistake, I cleared my fork and this happened due to removing the associated branch.
Feel free to reuse it how you please though.

PR #1318, opened 2023-02-09
Restore KOFLAGS to limit number of jobs

tzununbekov:
PR triggermesh/triggermesh#1317 overwrote the flag that limited the number of ko build threads. That lead to CI timeouts.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -134,26 +134,22 @@ jobs:
           name: Publishing container images to Docker Hub
           environment:
             KO_DOCKER_REPO: docker.io/triggermesh
-            KOFLAGS: --jobs=4
+            KOFLAGS: --jobs=4 --platform=linux/amd64,linux/arm64
             DIST_DIR: /tmp/dist/
           command: |
-            IMAGE_TAG=${CIRCLE_TAG:-${CIRCLE_SHA1}} \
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -137,9 +137,7 @@ jobs:
             KOFLAGS: --jobs=4
             DIST_DIR: /tmp/dist/
           command: |
-            IMAGE_TAG=${CIRCLE_TAG:-${CIRCLE_SHA1}} \
-            KOFLAGS="--platform linux/amd64,linux/arm64" \
```
sameersbn, line 141: Actually your change overrode the `--jobs=4` KOFLAG. Please move the new args to `KOFLAGS` to the env variable on line 137 for example.

.circleci/config.yml
```diff
@@ -137,9 +137,7 @@ jobs:
             KOFLAGS: --jobs=4
```
sameersbn, line 137:
```suggestion
            KOFLAGS: --jobs=4 --platform linux/amd64,linux/arm64
```

.circleci/config.yml
tzununbekov: Nice catch, thanks

sameersbn (approved):

PR #1319, opened 2023-02-09
Add SolaceSource & SolaceTarget

### LINKED ISSUE #1298: Add a Solace source
Way back then we had a prototype Solace source, we should get it back in this repo.

the original PR was at: https://github.com/triggermesh/event-sources/pull/85/files

Solace can give us connectivity to SAP HANA.

FranBarrera:
Closes #1298


cmd/solacesource-adapter/main.go
```diff
@@ -0,0 +1,27 @@
+/*
+Copyright 2022 TriggerMesh Inc.
```
JeffNeff, line 2:
```suggestion
Copyright 2023 TriggerMesh Inc.
```

cmd/solacetarget-adapter/main.go
```diff
@@ -0,0 +1,27 @@
+/*
+Copyright 2022 TriggerMesh Inc.
```
JeffNeff, line 2:
```suggestion
Copyright 2023 TriggerMesh Inc.
```

config/300-solacesource.yaml
```diff
@@ -0,0 +1,324 @@
+# Copyright 2022 TriggerMesh Inc.
```
JeffNeff, line 1:
```suggestion
# Copyright 2023 TriggerMesh Inc.
```

config/300-solacesource.yaml
```diff
@@ -0,0 +1,324 @@
...
+  name: solacesources.sources.triggermesh.io
+  labels:
+    eventing.knative.dev/source: 'true'
+    duck.knative.dev/source: 'true'
+    knative.dev/crd-install: 'true'
+    triggermesh.io/crd-install: 'true'
+  annotations:
+    registry.knative.dev/eventTypes: |
+      [
+        { "type": "io.triggermesh.solace.event" }
```
JeffNeff, line 27: If we could do the schema and match this registry it would be nice :) 
```
      [
        {
          "type": "com.amazon.cognito-identity.sync_trigger",
          "schema": "https://raw.githubusercontent.com/triggermesh/triggermesh/main/schemas/com.amazon.cognito-identity.sync_trigger.json",
          "description": ""
        }
      ]
```

config/300-solacesource.yaml
```diff
@@ -0,0 +1,324 @@
...
+        type: object
+        properties:
+          spec:
+            description: Desired state of the event source.
+            type: object
+            properties:
+              url:
+                description: Solace server used to bootstrap the connection.
+                type: string
+              queueName:
```
JeffNeff, line 58: queueName is not required?

pkg/sources/adapter/solacesource/adapter.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+
+		tlsCfg.InsecureSkipVerify = env.SkipVerify
+		connOption = amqp.ConnTLSConfig(tlsCfg)
+	}
+
+	amqpOpts := []amqp.ConnOption{
+		amqp.ConnIdleTimeout(idleTimeout),
+		connOption,
+	}
```
odacremolbap, line 88: Anonymous is being created but maybe overwritten later.
AFAIK SASL and TLS can (and should) be used together.

This might be a better pattern for creating the client.

```suggestion
	amqpOpts := []amqp.ConnOption{
		amqp.ConnIdleTimeout(idleTimeout),
	}

	if env.SASLEnable {
		amqpOpts = append(amqpOpts, amqp.ConnSASLPlain(env.Username, env.Password))
	} else {
		amqpOpts = append(amqpOpts, amqp.ConnSASLAnonymous())
	}

	if env.TLSEnable {
		tlsCfg := &tls.Config{}
		if env.CA != "" {
			addCAConfig(tlsCfg, env.CA)
		}

		if env.ClientCert != "" || env.ClientKey != "" {
			if err := addTLSCerts(tlsCfg, env.ClientCert, env.ClientKey); err != nil {
				logger.Panicw("Could not parse the TLS Certificates", zap.Error(err))
			}
		}

		tlsCfg.InsecureSkipVerify = env.SkipVerify
		amqpOpts = append(amqpOpts, amqp.ConnTLSConfig(tlsCfg))
	}
```

pkg/sources/adapter/solacesource/adapter.go
```diff
@@ -0,0 +1,183 @@
...
+
+var _ pkgadapter.Adapter = (*solacesourceAdapter)(nil)
+
+type solacesourceAdapter struct {
+	ceClient cloudevents.Client
+	logger   *zap.SugaredLogger
+	mt       *pkgadapter.MetricTag
+
+	amqpClient *amqp.Client
+	amqpOpts   []amqp.ConnOption
```
odacremolbap, line 46: I think options can be removed from the struct

pkg/sources/adapter/solacesource/adapter.go
```diff
@@ -0,0 +1,183 @@
...
+
+	// Create a receiver
+	receiver, err := session.NewReceiver(
+		amqp.LinkSourceAddress(a.queueName),
+	)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		ctx, cancel := context.WithTimeout(ctx, 1*time.Second)
```
odacremolbap, line 123: Maybe give it more than 1 second to exit. I would say something that after something like 15s we could say that graceful shutdown is not possible.

pkg/sources/adapter/solacesource/adapter.go
```diff
@@ -0,0 +1,183 @@
...
+		receiver.Close(ctx)
+		cancel()
+	}()
+
+	for {
+		// Receive message
+		msg, err := receiver.Receive(ctx)
+		if err != nil {
+			a.logger.Panicw("Error receiving messages", zap.Error(err))
```
odacremolbap, line 132: This start method is able to return an error. That sounds better than panicking here.

pkg/sources/adapter/solacesource/adapter.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+
+		err = msg.Accept()
+		if err != nil {
+			a.logger.Panicw("Error accepting messages", zap.Error(err))
+
+		}
```
odacremolbap, line 139: Accepting an event (marking it as delivered) should be done after it has been delivered to the broker. If done here we might inform Solace that the event has been received, then if we fail to deliver the event would be lost.

Move after `emitEvent

pkg/sources/adapter/solacesource/adapter.go
```diff
@@ -0,0 +1,183 @@
...
+
+		err = a.emitEvent(ctx, msg)
+		if err != nil {
+			a.logger.Panicw("Error sending event", zap.Error(err))
```
odacremolbap, line 143: Log an error (not panic), then continue the loop.
Failing to emit an event should mean that:

- we do not ack receiving it to Solace.
- we continue the loop, do not exit the adapter.

pkg/sources/adapter/solacesource/adapter.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+
+		if ctx.Err() != nil {
+			return nil
+		}
```
odacremolbap, line 148: If the context was closed after the call to receive.Receive, the msg var would be empty and msg.Accept and emitEvent would most probably error.

Also Receive when the context if finished returns the ctx.Err() as an error, which means that it will go through its error check and report it as an error.
Would it make sense to move this block just after the call to Receive?

pkg/targets/adapter/solacetarget/adapter.go
```diff
@@ -0,0 +1,180 @@
...
+		Namespace:     envAcc.GetNamespace(),
+		Name:          envAcc.GetName(),
+	}
+
+	metrics.MustRegisterEventProcessingStatsView()
+
+	env := envAcc.(*envAccessor)
+
+	idleTimeout := time.Duration(env.IdleTimeout) * time.Minute
+	connOption := amqp.ConnSASLAnonymous()
```
odacremolbap, line 54: see source comments regarding timeout and conn options.

pkg/targets/adapter/solacetarget/adapter.go
```diff
@@ -0,0 +1,180 @@
...
+	session, err := a.amqpClient.NewSession()
+	if err != nil {
+		log.Fatal("Creating amqp session")
+	}
+
+	// Create a sender
+	sender, err := session.NewSender(amqp.LinkTargetAddress(a.queueName))
+	if err != nil {
+		log.Fatal("Creating sender link:", err)
+	}
```
odacremolbap, line 155: Opening a session per dispatch dos not sound efficient. Should this be moved to Start?
Also, there are some log.Fatal ... Should that be using a.logger?
(Try not to panic on methods that are able to return an error)

config/300-solacesource.yaml
odacremolbap: I think there is no schema for messages from Solace, sounds like it produces data at queues disregarding the contents.

odacremolbap (changes requested):
Good direction.
I think there might be some corner cases to be taken into account, see comments.

odacremolbap (approved):
all good

PR #1320, opened 2023-02-10
Revert "Publish adapter images for ARM64 architecture"

tzununbekov:
Reverts #1317

Looks like we don't have enough [memory](https://app.circleci.com/pipelines/github/triggermesh/triggermesh/3647/workflows/2832ddbc-7516-40ba-a8d5-93332054d24d/jobs/10086/resources) to build effectively two sets of images, amd64 and arm64, CI build step is failing.

That's sad, would be great to have native arm64 images available, they don't even need to be build in the same time as amd64, we could build them asynchronously, nightly.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -134,7 +134,7 @@ jobs:
           name: Publishing container images to Docker Hub
           environment:
             KO_DOCKER_REPO: docker.io/triggermesh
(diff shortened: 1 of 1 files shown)


sameersbn:
PR says 0 files changed.. weird!

sameersbn (approved):

PR #1321, opened 2023-02-13
Add support for Splunk compatible APIs

odacremolbap:
Splunk compatible APIS might not use the `/services/collector/event/1.0` path that is appended by the adapter.

This change keeps the user's paths instead of trimming it if present.

### DIFF
--- a/pkg/targets/adapter/splunktarget/adapter.go
+++ b/pkg/targets/adapter/splunktarget/adapter.go
@@ -121,7 +121,9 @@ func newClient(hecURL url.URL, hecToken, index, hostname string, skipTLSVerify b
 		Transport: httpTransport,
 	}
 
-	hecURL.Path = eventURLPath
+	if hecURL.Path == "" || hecURL.Path == "/" {
+		hecURL.Path = eventURLPath
+	}
 
 	return &splunk.Client{
 		HTTPClient: httpClient,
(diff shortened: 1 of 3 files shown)


pkg/targets/adapter/splunktarget/adapter.go
```diff
@@ -121,7 +121,9 @@ func newClient(hecURL url.URL, hecToken, index, hostname string, skipTLSVerify b
 		Transport: httpTransport,
 	}
 
-	hecURL.Path = eventURLPath
+	if hecURL.Path == "" || hecURL.Path == "/" {
```
JeffNeff, line 124: why are we expecting the `/` ?

odacremolbap: A user might write the Splunk URL like:

- https://my.splunk.url:8888/
- https://my.splunk.url:8888

both cases will make the adapter append the default Splunk path, resulting in:

https://my.splunk.url:8888/services/collector/event/1.0

whereas any other path like https://my.splunk.url:8888/mypath will keep the user's path and won't append any suffix.

JeffNeff (approved):

jmcx:
Added docs here: https://github.com/triggermesh/docs/commit/07b3fe861f27c9c5b277f74b8b3e05dcfd04de02

PR #1349, opened 2023-03-01
E2e: gcloud repository default branch

odacremolbap:
It looks that the default branch depends on the git client.

We are reading that default branch, if none we fallback to the previous behavior.

### DIFF
--- a/test/e2e/framework/gcloud/repositories/repositories.go
+++ b/test/e2e/framework/gcloud/repositories/repositories.go
@@ -37,7 +37,7 @@ import (
 
 const (
 	gitDefaultRemoteName = "origin"
-	gitDefaultBranchName = "main"
+	gitDefaultBranch     = "master"
 
 	gitCommitterName  = "TriggerMesh e2e"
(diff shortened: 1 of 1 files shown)


test/e2e/framework/gcloud/repositories/repositories.go
```diff
@@ -37,7 +37,7 @@ import (
 
 const (
 	gitDefaultRemoteName = "origin"
-	gitDefaultBranchName = "main"
+	gitDefaultBranch     = "master"
```
sebgoa, line 40: so you are reverting to master ?

odacremolbap: I am reverting to the default it the git client does not have anything configured.

The test failed for me locally, because my default branch is main.
But it looks like some other test runners (GitHub actions, not sure Jenkins at other people's environments running these tests) might have a different default branch, or a non configured default branch.

For non configured, I am keeping the old value.
For all others I am reading the value using an extra command line call.

I won't know if this is valid until it executes in other environments, but right now it is failing where it used to succeed.

Issue #1371, opened 2023-03-23
Workload Identity - Add Support for Custom KSA 


coding-trees:
Hello,

From my understanding in regards to Google Cloud services, the way that Workload Identity is implemented (https://github.com/triggermesh/triggermesh/pull/1223)  is it takes the name from `gcpServiceAccount` field and creates a KSA with the same name. Unfortunately for us, our company has varied naming conventions and our gcpServiceAccount naming convention is different from our kubernetes service account naming convention. 

If I have a workload identity set up where gcpServiceAccount is `test-gcp-134-svc` and the kubernetes service account is `ksa-test-abc-svc` the Workload Identity will not work as it will create a KSA named `test-gcp-134-svc` instead of `ksa-test-abc-svc`

Could an optional field be added that will take in the name of the KSA to create to allow for such scenarios?

Maybe added a `kubernetesServiceAccount` field to the [GoogleCloudSourceAuth](https://docs.triggermesh.io/1.24/reference/sources/#sources.triggermesh.io/v1alpha1.GoogleCloudSourceAuth) that when set will create the KSA with the name set in that field, otherwise create the KSA from the `gcpServiceAccount` field.

```yaml
auth:
  description:
  type: object
  properties:
    serviceAccountKey:
      description: Service account key used to authenticate the event source and allow it to interact with Google Cloud APIs. Only the JSON format is supported.
      type: object
      properties:
        value:
          description: Literal value of the service account key.
          type: string
        valueFromSecret:
          description: A reference to a Kubernetes Secret object containing the service account key.
          type: object
          properties:
            name:
              type: string
            key:
              type: string
          required:
            - name
            - key
      oneOf:
        - required: [value]
        - required: [valueFromSecret]
    gcpServiceAccount:
      description: GCP Service account name to impersonate Identity and Access Management (IAM) service accounts to access Google Cloud services.
      pattern: ^[a-z][a-z0-9-]+@[a-z0-9-]+\.[a-z]+\.[a-z]{2,63}$
      type: string
    kubernetesServiceAccount:
      description: Kubernetes Service account name bound to the gcpServiceAccount to act as an IAM service account. Optional, by default the kubernetes service account will be created with the same names as gcpServiceAccount.
      type: string
```

[...]

PR #1372, opened 2023-03-24
GCP Cloud Services - Custom KSA Names

coding-trees:
Adding new `kubernetesServiceAccount` field that when set will be the name of the KSA created. As suggested, made changes to `ServiceAccountOptions`. I noticed that Workload Identity was only implemented in the sources for GCP Cloud services, but let me know if I missed anything. 

These changes address feature request outlined in https://github.com/triggermesh/triggermesh/issues/1371


pkg/apis/common/v1alpha1/gcp_types.go
```diff
@@ -26,8 +26,11 @@ const annotationGcpSA = "iam.gke.io/gcp-service-account"
 
 // GcpServiceAccountAnnotation returns a functional option that sets the GCP
 // Service Account annotation on Kubernetes ServiceAccount.
-func GcpServiceAccountAnnotation(gcpSA string) resource.ServiceAccountOption {
+func GcpServiceAccountAnnotation(gcpSA string, k8sSA string) resource.ServiceAccountOption {
 	return func(sa *corev1.ServiceAccount) {
+		if k8sSA != "" {
+			sa.ObjectMeta.Name = k8sSA
+		}
```
tzununbekov, line 33: This change concerns me because `GcpServiceAccountAnnotation` in its name and comment says that it sets Kubernetes object annotation, while in fact it implicitly overrides its name. Maybe we could have a separate functional option to override the SA name and move this logic to the `ServiceAccountOptions()` methods? 

Besides that, there is a potential bug - [`ServiceAccountName`](https://github.com/triggermesh/triggermesh/blob/v1.24.1/pkg/reconciler/adapter.go#L69-L85) function that's used to generate the original KSA name, also used in adapter reconciliation ([knsvc](https://github.com/triggermesh/triggermesh/blob/v1.24.1/pkg/reconciler/adapter.go#L237), [deployment](https://github.com/triggermesh/triggermesh/blob/v1.24.1/pkg/reconciler/adapter.go#L149)) which means that the name override won't be picked up by the adapters.

tzununbekov (changes requested):
Thanks for the contribution, @coding-trees!
There are a couple of things that need to be addressed before merging this PR. Feel free to comment if there are any questions or if I missed something in the review.

pkg/apis/common/v1alpha1/gcp_types.go
coding-trees: Yeah I see how this isn't the best place for this. I will revert it.

If I'm understanding correctly, the change would be:

In [serviceAccount.go](https://github.com/triggermesh/triggermesh/blob/v1.24.1/pkg/reconciler/resource/serviceAccount.go) add a new functional option like so:

```go
func SetServiceAccountName(name string) ServiceAccountOption {
	return func(sa *corev1.ServiceAccount) {
		sa.SetName(name)
	}
}
```

then in all of the GCP cloud sources' life cycle an additional check can be added in [ServiceAccountOptions()](https://github.com/triggermesh/triggermesh/blob/v1.24.1/pkg/apis/sources/v1alpha1/googlecloudpubsub_lifecycle.go#L73) like so:

```go
if k8sSA := s.Spec.Auth.KubernetesServiceAccount; k8sSA != nil {
	saOpts = append(saOpts, resource.SetServiceName(k8sSA))
}
```

Thanks

config/300-googlecloudauditlogssource.yaml
```diff
@@ -140,6 +140,9 @@ spec:
                       to access Google Cloud services.
                     pattern: ^[a-z][a-z0-9-]+@[a-z0-9-]+\.[a-z]+\.[a-z]{2,63}$
                     type: string
+                  kubernetesServiceAccount:
+                    description: Kubernetes Service account name bound to the gcpServiceAccount to act as an IAM service account. Optional, by default the kubernetes service account will be created with the same names as gcpServiceAccount.
```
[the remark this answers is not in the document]
coding-trees: Yeah it's clearer that way - thanks for the suggestions, I'll apply it to all of the GCP Cloud sources.

pkg/apis/common/v1alpha1/gcp_types.go
tzununbekov: Yes, that looks good to me. There is still [`ServiceAccountName`](https://github.com/triggermesh/triggermesh/blob/v1.24.1/pkg/reconciler/adapter.go#L69-L85) part that also needs to be aware of the name overwrite to pass a correct value to the adapter deployment. Not sure what would be the best solution there.

coding-trees: Would something like this work? 

In the [object_interface](https://github.com/triggermesh/triggermesh/blob/main/pkg/apis/common/v1alpha1/object_interfaces.go), add a new function 

```go
func ServiceAccountName(r Reconcilable) bool {
	saProvider, ok := r.(ServiceAccountProvider)
	return ok && saProvider.ServiceAccountName()
}
```

which means updating the `ServiceAccountProvider`

```go
type ServiceAccountProvider interface {
	WantsOwnServiceAccount() bool
	ServiceAccountOptions() []resource.ServiceAccountOption
	ServiceAccountName() string
}
```

then in the adapter.go it could be something like

```go
func ServiceAccountName(rcl v1alpha1.Reconcilable) string {
	if v1alpha1.WantsOwnServiceAccount(rcl) {
		rclName := rcl.GetName()
		if svcAcctName := v1alpha1.ServiceAccountName(rcl); svcAcctName != nil {
			rclName = svcAcctName
		}
		

		// Edge case: we need to make sure some characters are inserted
		// between the component name and the component instance's name
		// to avoid clashing with the shared "{kind}-adapter"
		// ServiceAccount in case the component instance is named
		// "adapter". We picked 'i' for "instance" to keep it short yet
		// distinguishable.
		return kmeta.ChildName(ComponentName(rcl)+"-i-", rclName)
	}

	return MTAdapterObjectName(rcl)
}
```

the new `ServiceAccountProvider.ServiceAccountName()` would only be implemented by the GCP Cloud services for now.

tzununbekov: I think that changing the `ServiceAccountProvider` interface is excessive because we'll have to update all our component types that implement it right now (not only GCP, 24 of them). I'm suggesting that we keep an old interface and just update the way we get the SA name for adapters:

First, KSA name override function. I put it in the `gcp_types.go` because we don't need it anywhere else right now. Can be moved to common options if it changes.
```diff
diff --git a/pkg/apis/common/v1alpha1/gcp_types.go b/pkg/apis/common/v1alpha1/gcp_types.go
index 45a9995d..d16a71c6 100644
--- a/pkg/apis/common/v1alpha1/gcp_types.go
+++ b/pkg/apis/common/v1alpha1/gcp_types.go
@@ -31,3 +31,11 @@ func GcpServiceAccountAnnotation(gcpSA string) resource.ServiceAccountOption {
                metav1.SetMetaDataAnnotation(&sa.ObjectMeta, annotationGcpSA, gcpSA)
        }
 }
+
+// K8sServiceAccountName returns a functional option that overwrites the
+// Kubernetes Service Account name.
+func K8sServiceAccountName(name string) resource.ServiceAccountOption {
+       return func(sa *corev1.ServiceAccount) {
+               sa.SetName(name)
+       }
+}
```

This part is the same, propagated to other  GCP lifecycle files:
```diff
diff --git a/pkg/apis/sources/v1alpha1/googlecloudauditlogs_lifecycle.go b/pkg/apis/sources/v1alpha1/googlecloudauditlogs_lifecycle.go
index 190d4b54..ea7a9744 100644
--- a/pkg/apis/sources/v1alpha1/googlecloudauditlogs_lifecycle.go
+++ b/pkg/apis/sources/v1alpha1/googlecloudauditlogs_lifecycle.go
@@ -90,6 +90,9 @@ func (s *GoogleCloudAuditLogsSource) ServiceAccountOptions() []resource.ServiceA
        if gcpSA := s.Spec.Auth.GCPServiceAccount; gcpSA != nil {
                saOpts = append(saOpts, v1alpha1.GcpServiceAccountAnnotation(*gcpSA))
        }
+	if k8sSA := s.Spec.Auth.KubernetesServiceAccount; k8sSA != nil {
+ 		saOpts = append(saOpts, v1alpha1.K8sServiceAccountName(*k8sSA))
+	}
        return saOpts
 }
```

Here is the main difference - instead of generating the SA name from `ServiceAccountName` to reconcile the adapters, we get the SA object with all the mutations (including name override):
```diff
diff --git a/pkg/reconciler/adapter.go b/pkg/reconciler/adapter.go
index f0f454e1..f9b7e85f 100644
--- a/pkg/reconciler/adapter.go
+++ b/pkg/reconciler/adapter.go
@@ -146,7 +146,7 @@ func commonAdapterDeploymentOptions(rcl v1alpha1.Reconcilable) []resource.Object
                resource.PodLabel(appPartOfLabel, partOf),
                resource.PodLabel(appManagedByLabel, managedBy),

-               resource.ServiceAccount(ServiceAccountName(rcl)),
+               resource.ServiceAccount(serviceAccount(rcl)),

                resource.EnvVar(envComponent, app),
                resource.EnvVar(envSinkTimeout, strconv.FormatInt(int64(defaultSinkTimeout.Seconds()), 10)),
@@ -234,7 +234,7 @@ func commonAdapterKnServiceOptions(rcl v1alpha1.Reconcilable) []resource.ObjectO
                resource.PodLabel(appPartOfLabel, partOf),
                resource.PodLabel(appManagedByLabel, managedBy),

-               resource.ServiceAccount(ServiceAccountName(rcl)),
+               resource.ServiceAccount(serviceAccount(rcl)),

                resource.EnvVar(envComponent, app),
                resource.EnvVar(envSinkTimeout, strconv.FormatInt(int64(defaultSinkTimeout.Seconds()), 10)),
@@ -258,13 +258,17 @@ func commonAdapterKnServiceOptions(rcl v1alpha1.Reconcilable) []resource.ObjectO
        return objectOptions
 }

-// newServiceAccount returns a ServiceAccount object with its OwnerReferences
+// serviceAccount returns a ServiceAccount object with its OwnerReferences
 // metadata attribute populated from the given owners.
-func newServiceAccount(rcl v1alpha1.Reconcilable, owners []kmeta.OwnerRefable) *corev1.ServiceAccount {
-       return resource.NewServiceAccount(rcl.GetNamespace(), ServiceAccountName(rcl),
+func serviceAccount(rcl v1alpha1.Reconcilable, owners ...kmeta.OwnerRefable) *corev1.ServiceAccount {
+       sa := resource.NewServiceAccount(rcl.GetNamespace(), ServiceAccountName(rcl),
                resource.Owners(owners...),
                resource.Labels(CommonObjectLabels(rcl)),
        )
+       for _, m := range serviceAccountMutations(rcl) {
+               m(sa)
+       }
+       return sa
 }
```

`resource.ServiceAccount` receives the SA and reads its name (original or overridden):
```diff
diff --git a/pkg/reconciler/resource/podspecable.go b/pkg/reconciler/resource/podspecable.go
index 31812bc3..970ca17d 100644
--- a/pkg/reconciler/resource/podspecable.go
+++ b/pkg/reconciler/resource/podspecable.go
@@ -55,7 +55,7 @@ func Container(c *corev1.Container) ObjectOption {
 }

 // ServiceAccount sets the ServiceAccount name of a PodSpecable.
-func ServiceAccount(sa string) ObjectOption {
+func ServiceAccount(sa *corev1.ServiceAccount) ObjectOption {
        return func(object interface{}) {
                var saName *string

@@ -66,7 +66,7 @@ func ServiceAccount(sa string) ObjectOption {
                        saName = &o.Spec.Template.Spec.ServiceAccountName
                }

-               *saName = sa
+               *saName = sa.Name
        }
 }
```

All KSA options are applied inside the `serviceAccount` function:
```diff
diff --git a/pkg/reconciler/reconcile.go b/pkg/reconciler/reconcile.go
index 10325984..b942ea24 100644
--- a/pkg/reconciler/reconcile.go
+++ b/pkg/reconciler/reconcile.go
@@ -297,11 +297,7 @@ func (r *GenericRBACReconciler[T, L]) reconcileRBAC(ctx context.Context,

        rcl := v1alpha1.ReconcilableFromContext(ctx)

-       desiredSA := newServiceAccount(rcl, owners)
-       for _, m := range serviceAccountMutations(rcl) {
-               m(desiredSA)
-       }
-
+       desiredSA := serviceAccount(rcl, owners...)
        currentSA, err := r.getOrCreateAdapterServiceAccount(ctx, desiredSA)
        if err != nil {
                return nil, err
```


And the CRDs are the same.
What do you think about this approach? I haven't tested it, please let me know if missed something.

coding-trees: Hi @tzununbekov  - this seems like a clean solution to this. I applied the changes you suggested and didn't see any errors - however, I couldn't get the e2e tests to run locally.

Thanks for your help with this.

tzununbekov: Nice! Our e2e requires a configured environment with the credentials to use external resources, we'll see how it goes on the main branch. Thank you.

tzununbekov:
@coding-trees there are a couple of CI jobs that failed, but should be easy to fix:
```
hack/crd-update/update-all-crds.sh,
```
-to ensure that all CRDs are formatted correctly
```
make codegen
```
-to update autogenerated client files.

And the fix in the test related to `ServiceAccount` function update - the easiest way to fix that is probably to pass a mock SA object to that function in the test.

coding-trees:
> @coding-trees there are a couple of CI jobs that failed, but should be easy to fix:
> 
> ```
> hack/crd-update/update-all-crds.sh
> ```
> 
> -to ensure that all CRDs are formatted correctly,
> 
> ```
> make codegen
> ```
> 
> -to update autogenerated client files.
> 
> And the fix in the test related to `ServiceAccount` function update - the easiest way to fix that is probably to pass a mock SA object to that function in the test.

Thanks for the tips, I was able to fix the tests and CRDs however when I ran `make codegen` it resulted in deleting ~903 files :0

image



Doesn't seem right so I didn't commit it - maybe I'm doing something wrong?

tzununbekov:
> > @coding-trees there are a couple of CI jobs that failed, but should be easy to fix:
> > ```
> > hack/crd-update/update-all-crds.sh
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > -to ensure that all CRDs are formatted correctly,
> > ```
> > make codegen
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > -to update autogenerated client files.
> > And the fix in the test related to `ServiceAccount` function update - the easiest way to fix that is probably to pass a mock SA object to that function in the test.
> 
> Thanks for the tips, I was able to fix the tests and CRDs however when I ran `make codegen` it resulted in deleting ~903 files :0
> 
> image
> 
> Doesn't seem right so I didn't commit it - maybe I'm doing something wrong?

Yeah, something is off with the codegen paths, I took care of it.

tzununbekov (approved):
@coding-trees thank you for the contribution 👍

coding-trees:
Thank you for all the help @tzununbekov

PR #1380, opened 2023-03-28
Add support for setting adapter nodeSelector

FranBarrera:
Addresses one of the features requested in https://github.com/triggermesh/triggermesh/issues/768.

Example:
```
spec:
  # ...
  adapterOverrides:
    nodeSelector:
      disktype: ssd
```

Result:
```
kubectl get deployment awss3source-sample -o jsonpath='{.spec.template.spec.nodeSelector}'
{"disktype":"ssd"}

```

### DIFF
--- a/pkg/reconciler/resource/deployment_test.go
+++ b/pkg/reconciler/resource/deployment_test.go
@@ -103,6 +103,9 @@ func TestNewDeploymentWithDefaultContainer(t *testing.T) {
 					Tolerations: []corev1.Toleration{{
 						Key: "taint", Operator: "Exists",
 					}},
+					NodeSelector: map[string]string{
+						"disktype": "ssd",
+					},
 					Containers: []corev1.Container{{
 						Name:  defaultContainerName,
 						Image: tImg,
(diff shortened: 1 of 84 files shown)


pkg/reconciler/resource/deployment_test.go
```diff
@@ -103,6 +103,9 @@ func TestNewDeploymentWithDefaultContainer(t *testing.T) {
 					Tolerations: []corev1.Toleration{{
 						Key: "taint", Operator: "Exists",
 					}},
+					NodeSelector: map[string]string{
+						"disktype": "ssd",
+					},
```
odacremolbap, line 108: Not sure ho this work, I cannot find the expected selector at the deployment that is being tested. Is it being set somewhere else? Same for the Service version.

odacremolbap (commented):
As suggested by the action output, can you

```console
hack/crd-update/update-all-crds.sh
```

pkg/reconciler/resource/deployment_test.go
FranBarrera: Should be fixed now

odacremolbap (approved):

PR #1381, opened 2023-03-29
Add support for setting adapter Affinity

FranBarrera:
Closes https://github.com/triggermesh/triggermesh/issues/768.

Example:
```
spec:
  # ...
  adapterOverrides:
    affinity:
      nodeAffinity:
        preferredDuringSchedulingIgnoredDuringExecution:
        - weight: 1
          preference:
            matchExpressions:
            - key: zone
              operator: In
              values:
              - zone-a
```

Result:
```
kubectl get deployment awss3source-sample -o jsonpath='{.spec.template.spec.affinity}'
{"nodeAffinity":{"preferredDuringSchedulingIgnoredDuringExecution":[{"preference":{"matchExpressions":[{"key":"zone","operator":"In","values":["zone-a"]}]},"weight":1}]}}

```


odacremolbap (dismissed):
Nicely done.

The CRDs are now humongous, but this is the way of implementing this. :heavy_check_mark:

tzununbekov:
Oh my... are you sure that we want our CRDs to look like this? This update makes them 5 times bigger (~12Kb vs 62Kb) 😨 
[Initially](https://github.com/triggermesh/triggermesh/issues/847), it was about the overrides documentation because it was a custom set of knobs, like a `public` switch, but do we really need to duplicate a well-known affinity object for this reason? 
My main concern is that CRDs are now used in the CLI for object generation and commands auto-completion, so it is often loaded in memory. Making it this big without a clear reason may negatively impact the UX.

FranBarrera:
@tzununbekov @odacremolbap I've updated the PR to make the affinity arbitrary and I've added checks at the Webhook to verify things like some wrong names in the affinity.

tzununbekov:
Thanks. I see a lot of things done for the webhook and I'm struggling to find the reason. Could you point me to the code that implements webhook validation logic?

FranBarrera:
@tzununbekov There are some Unmarshal that are done automatically.
 
 if I add an affinity like this:
```
  adapterOverrides:
    affinity:
      nodeAffinityy:
```
it will return this error at the creation time:
```
admission webhook "defaulting.webhook.triggermesh.io" denied the request: mutation failed: cannot decode incoming new object: json: unknown field "nodeAffinityy"
```

tzununbekov:
I'm pretty sure that the **Defaulting** webhook error is a side-effect of internal object processing and not its direct function; the **Validation** webhook should do that kind of validation. Nevertheless, we can live with that.

I tested the Affinity function and it doesn't work for me. I did some digging and it looks like it is not the way it works in the Knative. Are you sure that it works for you? Please try to create the Kn-based adapter and check if it is online.

FranBarrera:
@tzununbekov What do you mean that it doesn't works for you? I've tested creating a source, the adapter is running and checking if the deployment has the affinity set:
```
kubectl get deployment awss3source-sample -o jsonpath='{.spec.template.spec}'|grep affinity
{"affinity":{"nodeAffinity":{"preferredDuringSchedulingIgnoredDuringExecution":[{"preference":{"matchExpressions":[{"key":"zone","operator":"In","values":["zone-a"]}]},"weight":1}]}},"containers":[{"env":[{"name":"K_COMPONENT","value":"awss3source"},{"name":"K_SINK_TIMEOUT","value":"30"},{"name":"METRICS_PROMETHEUS_PORT","value":"9092"},{"name":"NAMESPACE","value":"default"},{"name":"NAME","value":"sample"},{"name":"K_SINK","value":"http://sockeye.default.svc.cluster.local"},{"name":"AWS_ACCESS_KEY_ID","valueFrom":{"secretKeyRef":{"key":"AWS_ACCESS_KEY_ID","name":"aws"}}},{"name":"AWS_SECRET_ACCESS_KEY","valueFrom":{"secretKeyRef":{"key":"AWS_SECRET_ACCESS_KEY","name":"aws"}}},{"name":"ARN","value":"arn:aws:sqs:us-east-1:043455440429:s3-events_fran-dev"},{"name":"SQS_MESSAGE_PROCESSOR","value":"s3"},{"name":"K_LOGGING_CONFIG"},{"name":"K_METRICS_CONFIG","value":"{\"Domain\":\"\",\"Component\":\"\",\"PrometheusPort\":0,\"PrometheusHost\":\"\",\"ConfigMap\":null}"},{"name":"K_TRACING_CONFIG","value":"{\"backend\":\"\",\"debug\":\"false\",\"sample-rate\":\"0\"}"}],"image":"ko.local/awssqssource-adapter-4d82fe36bba98f7a2ff73fb3da133a45:f9c6f813f1038432fbc67602b261dddad581669c588b880a0af4ce1441fa64ff","imagePullPolicy":"IfNotPresent","name":"adapter","ports":[{"containerPort":9092,"name":"metrics","protocol":"TCP"},{"containerPort":8080,"name":"health","protocol":"TCP"}],"resources":{},"startupProbe":{"failureThreshold":60,"httpGet":{"path":"/health","port":"health","scheme":"HTTP"},"periodSeconds":1,"successThreshold":1,"timeoutSeconds":1},"terminationMessagePath":"/dev/termination-log","terminationMessagePolicy":"FallbackToLogsOnError"}],"dnsPolicy":"ClusterFirst","restartPolicy":"Always","schedulerName":"default-scheduler","securityContext":{},"serviceAccount":"awss3source-adapter","serviceAccountName":"awss3source-adapter","terminationGracePeriodSeconds":30}
```

About the Webhook, to check our objects with the webhook we need to add them to the `NewDefaultingAdmissionController` and `NewValidationAdmissionController`, if we don't add our objects there it doesn't check anything.

tzununbekov:
It looks like you tested only a Deployment-based component, not a Kn-based one. Knative serving, by default, [does not allow](https://github.com/knative/serving/blob/v0.36.3/pkg/apis/config/features.go#L54) setting the affinity parameter, it should be enabled through the Features and Extensions [configuration](https://knative.dev/docs/serving/configuration/feature-flags/#kubernetes-node-affinity). Otherwise, adapter deployment won't pass Serving validation. Btw, the same goes for the `nodeSelector` attribute:
```
Failed to create adapter Service "foo-transformation": admission webhook "validation.webhook.serving.knative.dev" denied the request: validation failed: must not set the field(s): spec.template.spec.affinity
``` 
IMO, this detail must have been mentioned because without knowing that, setting these adapter overrides may cost some hours of debugging the "outdated adapter specs" kind of issue. 

Regarding the webhook implementation, I think we're on different pages. What I'm saying is that there is not much sense in having a webhook controller up and running while none of our components have Validation and Default methods implemented (those numerous `SetDefaults` and `Validate` methods that you created for all components and left empty - they are responsible for admission controller logic).
I see two options here: 
1. Do not do any validation on our side and let the K8s controller handle incorrect overrides,
2. Define a common mechanism to run validation on shared objects (such as `adapterOverrides`) not to duplicate it on all resources.

I leaning towards the first option. Looks like Knative also does not care about validating these features ([here](https://github.com/knative/serving/blob/v0.36.3/pkg/apis/serving/k8s_validation.go)).

pkg/apis/extensions/v1alpha1/register.go
```diff
@@ -44,12 +46,33 @@ var (
 	AddToScheme = SchemeBuilder.AddToScheme
 )
 
+// ObjectWebhook is an interface to use in the webhook.
+type ObjectWebhook interface {
+	runtime.Object
+	apis.Validatable
+	apis.Defaultable
+	kmeta.OwnerRefable
+}
```
tzununbekov, line 55: _Here and in the other `*_register.go` files below:_
I don't think that we need to duplicate interfaces just to pass some additional parameters in one module. If I really wanted to re-use object lists, I'd define a common structure somewhere in `pkg/apis/common/v1alpha1/types.go`:

```go
// GroupObject holds the API group object types.
//
// +k8s:deepcopy-gen=false
type GroupObject struct {
	Single, List runtime.Object
}
```

And then:

```diff
diff --git a/pkg/apis/extensions/v1alpha1/register.go b/pkg/apis/extensions/v1alpha1/register.go
index e65a8af2..ea2c1e53 100644
--- a/pkg/apis/extensions/v1alpha1/register.go
+++ b/pkg/apis/extensions/v1alpha1/register.go
@@ -17,9 +17,8 @@ limitations under the License.
 package v1alpha1

 import (
+       "github.com/triggermesh/triggermesh/pkg/apis/common/v1alpha1"
        "github.com/triggermesh/triggermesh/pkg/apis/extensions"
-       "knative.dev/pkg/apis"
-       "knative.dev/pkg/kmeta"

        metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
        "k8s.io/apimachinery/pkg/runtime"
@@ -46,33 +45,16 @@ var (
        AddToScheme = SchemeBuilder.AddToScheme
 )

-// ObjectWebhook is an interface to use in the webhook.
-type ObjectWebhook interface {
-       runtime.Object
-       apis.Validatable
-       apis.Defaultable
-       kmeta.OwnerRefable
-}
-
-// Objects holds instances of ObjectWebhook and runtime.Objects List.
-// +k8s:deepcopy-gen=false
-type Objects struct {
-       Single ObjectWebhook
-       List   runtime.Object
-}
-
 // AllTypes is a list of all the types defined in this package.
-var AllTypes = []Objects{
-       {&Function{}, &FunctionList{}},
+var AllTypes = []v1alpha1.GroupObject{
+       {Single: &Function{}, List: &FunctionList{}},
 }

 // Adds the list of known types to Scheme.
 func addKnownTypes(scheme *runtime.Scheme) error {
-       allTypes := make([]runtime.Object, 0, len(AllTypes)*2)
        for _, t := range AllTypes {
-               allTypes = append(allTypes, t.Single, t.List)
+               scheme.AddKnownTypes(SchemeGroupVersion, t.Single, t.List)
        }
-       scheme.AddKnownTypes(SchemeGroupVersion, allTypes...)
        metav1.AddToGroupVersion(scheme, SchemeGroupVersion)
        return nil
 }
```

pkg/apis/flow/v1alpha1/register.go
tzununbekov, line 1: Same comment as for `pkg/apis/extensions/v1alpha1/register.go`

pkg/apis/routing/v1alpha1/register.go
tzununbekov, line 1: Same comment as for `pkg/apis/extensions/v1alpha1/register.go`

pkg/apis/sources/v1alpha1/register.go
tzununbekov, line 1: Same comment as for `pkg/apis/extensions/v1alpha1/register.go`

pkg/apis/targets/v1alpha1/register.go
tzununbekov, line 1: Same comment as for `pkg/apis/extensions/v1alpha1/register.go`

tzununbekov (changes requested):
A couple of small comments before the merge.

pkg/apis/extensions/v1alpha1/register.go
FranBarrera: @tzununbekov The line that you comment here: `{Single: &Function{}, List: &FunctionList{}},` doesn't seems to be replicable for all components, maybe I'm wrong but I don't think we have these common structs for all groups  like sources, targets, flow and routing. We have a struct for each component, so I guess it'll complicate more the code.

tzununbekov: > maybe I'm wrong but I don't think we have these common structs for all groups like sources, targets, flow and routing

Which common structs do you mean?

FranBarrera: Maybe I'm not understanding the solution you propose but I mean for extensions there is only Function and FunctionList but for example for sources we have x and xList per source, that seems like a lot to add, that also happens for others like targets, flow, routing.

tzununbekov: I did exactly what you proposed with the only difference - I didn't create a new interface for each group but created a common `GroupObject` structure in `pkg/apis/common/v1alpha1/types.go` (as it is mentioned in the first comment) so that all resources can be registered as a Function example in this diff:

```go
var AllTypes = []v1alpha1.GroupObject{
  {Single: &Function{}, List: &FunctionList{}},
}
```
it also works for other resources, eg:

```go
var AllTypes = []v1alpha1.GroupObject{
       {Single: &JQTransformation{}, List: &JQTransformationList{}},
       {Single: &Synchronizer{}, List: &SynchronizerList{}},
       {Single: &Transformation{}, List: &TransformationList{}},
       {Single: &XMLToJSONTransformation{}, List: &XMLToJSONTransformationList{}},
       {Single: &XSLTTransformation{}, List: &XSLTTransformationList{}},
}
```

If it is still confusing, I can push the full version of this update, I have it all stashed locally.

FranBarrera: Oh I see, thanks

config/300-awscloudwatchlogssource.yaml
```diff
@@ -222,9 +223,15 @@ spec:
                   nodeSelector:
                     description: NodeSelector only allow the object pods to be created at nodes where all selector labels
                       are present, as documented at https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector.
+                      NodeSelector only works on kubernetes deployment-based components, not on knative-based components.
                     type: object
                     additionalProperties:
                       type: string
+                  affinity:
+                    description: Scheduling constraints of the pod. More info at https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity.
+                      Affinity only works on kubernetes deployment-based components, not on knative-based components.
```
tzununbekov, line 232: This is not quite right.

```
<Affinity|NodeSelector|Tolerations> require additional configuration for Knative-based deployments: https://knative.dev/docs/serving/configuration/feature-flags/ 
```

tzununbekov (approved):

mukundjalan:
@FranBarrera can you please add an example of how can I add `nodeSelector` to the adapter?

PR #1383, opened 2023-03-30
Azure Queue Storage Source - Add Visibility timeout parameter

### LINKED ISSUE #1382: AzureQueueStorageSource - Configurable Timeout Support
The azure storage queue source in Triggermesh currently has a [20 second timeout](https://github.com/triggermesh/triggermesh/blob/main/pkg/sources/adapter/azurequeuestoragesource/adapter.go#L148). Azure queue however, supports message level timeouts upto 7 days.
An option to configure the default timeout would definitely help.

I had a use case where I needed a longer timeout. Ended up writing my own pollers since the timeout was not configurable in 
Triggermesh.

FranBarrera:
Closes #1382

This PR adds a new optional parameter to set the visibility timeout to Azure Queue Storage Source following the ISO8601, more info: https://en.wikipedia.org/wiki/ISO_8601

### DIFF
--- a/config/300-azurequeuestoragesource.yaml
+++ b/config/300-azurequeuestoragesource.yaml
@@ -76,6 +76,14 @@ spec:
                     required:
                     - name
                     - key
+              visibilityTimeout:
+                description: Specifies the new visibility timeout value, in seconds, relative to server time.
+                  The new value must be larger than or equal to 0, and can't be larger than 7 days.
+                  The visibility timeout of a message can't be set to a value later than the expiry time.
+                  You can update a message until it has been deleted or has expired. Default PT20S (20s)
+                  More information on Duration format - https://www.iso.org/iso-8601-date-and-time-format.html - https://en.wikipedia.org/wiki/ISO_8601
+                type: string
+                pattern: '^P(?:([0-6]D)?(?:T(?:([0-9]|1[0-9]|2[0-3])H)?(?:([0-5]?[0-9])M)?(?:([0-5]?[0-9]|60)S)?)?|7D)$'
               sink:
                 description: The destination of events sourced from Azure Storage.
                 type: object
--- a/config/300-azurequeuestoragesource.yaml
+++ b/config/300-azurequeuestoragesource.yaml
@@ -80,8 +80,8 @@ spec:
                 description: Specifies the new visibility timeout value, in seconds, relative to server time. The new value
                   must be larger than or equal to 0, and can't be larger than 7 days. The visibility timeout of a message
                   can't be set to a value later than the expiry time. You can update a message until it has been deleted or
-                  has expired. Default PT20S (20s) More information on Duration format - https://www.iso.org/iso-8601-date-and-time-format.html
-                  - https://en.wikipedia.org/wiki/ISO_8601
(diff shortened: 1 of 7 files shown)


odacremolbap (approved):

tzununbekov:
_nit:_ k8s' `duration` format is [documented](https://github.com/kubernetes/apiextensions-apiserver/blob/29400d7010f44f7d6ad2c6762604185143382598/pkg/apis/apiextensions/v1/types_jsonschema.go#L51) as Go's `time.Duration`. Timeouts in our components ([`visibilityTimeout`](https://github.com/triggermesh/triggermesh/blob/v1.24.1/config/300-awssqssource.yaml#L67-L73), [`timeout`](https://github.com/triggermesh/triggermesh/blob/v1.24.1/config/304-synchronizer.yaml#L73-L76)) are also based on Go's duration format. Some other common K8s components that I was able to check also use `time.Duration` format. It feels a bit weird to have the same attribute use different input formats inside the same project (even the same "visibilityTimeout" parameter). Why did you decide to switch to ISO format? Is it because of OpenAPI 3.0 spec? Maybe we could do something to make it consistent across the components?

FranBarrera:
> _nit:_ k8s' `duration` format is [documented](https://github.com/kubernetes/apiextensions-apiserver/blob/29400d7010f44f7d6ad2c6762604185143382598/pkg/apis/apiextensions/v1/types_jsonschema.go#L51) as Go's `time.Duration`. Timeouts in our components ([`visibilityTimeout`](https://github.com/triggermesh/triggermesh/blob/v1.24.1/config/300-awssqssource.yaml#L67-L73), [`timeout`](https://github.com/triggermesh/triggermesh/blob/v1.24.1/config/304-synchronizer.yaml#L73-L76)) are also based on Go's duration format. Some other common K8s components that I was able to check also use `time.Duration` format. It feels a bit weird to have the same attribute use different input formats inside the same project (even the same "visibilityTimeout" parameter). Why did you decide to switch to ISO format? Is it because of OpenAPI 3.0 spec? Maybe we could do something to make it consistent across the components?

I agree on having something consistent across the components but go time is language specific, while ISO is a standard.

Knative uses ISO for durations too: https://knative.dev/docs/eventing/event-delivery/#configuring-subscription-event-delivery
Maybe we should move other components to ISO? We already have [brokers](https://github.com/triggermesh/brokers) using ISO.

tzununbekov:
Go time format is natively supported by k8s, so I don't see why we can't keep using it. Personally, I'm not a fan of the ISO format because it's not intuitive, but having the same attributes declared in different formats is even worse than that. Hence, if you think that switching to the ISO is essential and worth introducing breaking change, then go for it. Hopefully, we will not follow Knative's specification style in the future.

FranBarrera:
Ok, no problem, I'll revert it to go time and then we can discuss what standard we should use

tzununbekov:
That works for me, thank you. I noticed that the [core k8s](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/) components use `<param>Seconds` attribute name to avoid any confusion in formats. Maybe we could use a similar approach. The only concern is backward compatibility for SQS source spec.

jmcx:
Sources_-_TriggerMesh

@FranBarrera any chance we can get some better documentation for the attribute to appear in the CRD?

PR #1385, opened 2023-04-04
Event schemas update

tzununbekov:
Triggermesh sources and targets event schema added


sebgoa (approved):

schemas/com.amazon.cloudwatch.logs.log.json
FranBarrera, line 1: This change requires to update the url in the CRD and also we've changed here the schema type, so we should update it in the CRD too.

tzununbekov: Thanks, it was unintentional leftover of names re-organization:
https://github.com/triggermesh/triggermesh/pull/1388

config/301-googlecloudworkflowstarget.yaml
```diff
@@ -23,7 +23,10 @@ metadata:
   annotations:
     registry.triggermesh.io/acceptedEventTypes: |
       [
-        { "type": "io.trigermesh.google.workflows.run" }
+        {
+          "type": "io.trigermesh.google.workflows.run",
+          "schema": "https://raw.githubusercontent.com/triggermesh/triggermesh/main/schemas/io.trigermesh.google.workflows.run.json"
```
FranBarrera, line 28: This is wrong, the schema should be in the `registry.knative.dev/eventTypes` annotation, it also happens on others crds.

tzununbekov: `registry.knative.dev/eventTypes` are events produced by the components, `registry.triggermesh.io/acceptedEventTypes`, as the name suggests, event types expected by the target. It is not an error, this annotation was in the CRD from the very beginning.

FranBarrera: I agree but the schemas that we added are always in the `registry.knative.dev/eventTypes` following the knative way:
https://github.com/knative/eventing/blob/main/docs/spec/sources.md#source-registry

If we are going to add schemas in two different annotations we should update all the other schemas.

tzununbekov: All schemas are up to date. As I said above, events produced by the components are listed under `registry.knative.dev/eventTypes`, events consumed by the components - `registry.triggermesh.io/acceptedEventTypes`. It is not something introduced in this PR, it's always been like this.

PR #1391, opened 2023-04-13
HTTPTarget improvements

tzununbekov:
HTTPTarget improvements:
1. Expected request body type switched from `string` to `json.RawMessage` to handle JSON data. The use-case to reproduce the issue is simple: Webhook source -> Transformation -> HTTP Target. Transformation in this context can either shift the whole Webhook payload to the `body` key or compose it from the multiple values of the source event. In both cases, the resulting `body` will be a JSON object, which is expected transformation behavior. Before this update, the target adapter complained about it: `json: cannot unmarshal object into Go struct field RequestData.body of type string`,
2. CR's `spec.response.*` parameters are not required and without them the adapter fails to send back the responses (CE context validation). Default values for CE type and source attributes are added,
3. HTTPTarget resource implements `EventSource` and `EventReceiver` interfaces to comply with the rest of the components. Resource's status exposes expected and produced CE attributes.

_Upd:_ 
4. Since the target's expected event has a rather strict [format requirement](https://github.com/triggermesh/triggermesh/blob/8b1373837239390c017809faf3347ac18a3afe9c/pkg/targets/adapter/httptarget/adapter.go#L148-L151), JSON schema was added to explicitly announce the request contract. The target adapter requires an incoming event to be of `io.triggermesh.http.request` type, otherwise event processing will fail.

### DIFF
--- a/config/301-httptarget.yaml
+++ b/config/301-httptarget.yaml
@@ -56,6 +56,7 @@ spec:
             description: Desired state of event target.
             properties:
               response:
+                description: HTTP target response event attributes.
                 type: object
                 properties:
                   eventType:
@@ -218,6 +219,18 @@ spec:
               observedGeneration:
                 type: integer
                 format: int64
+              ceAttributes:
+                type: array
+                items:
+                  type: object
+                  properties:
+                    type:
+                      type: string
+                    source:
+                      type: string
+                  required:
+                  - type
+                  - source
               conditions:
                 type: array
                 items:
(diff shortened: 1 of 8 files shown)


odacremolbap:
`HTTPTarget` is able to interact with non JSON systems.
Will it be able to do so after this proposed change?

tzununbekov:
> `HTTPTarget` is able to interact with non JSON systems.
> Will it be able to do so after this proposed change?

This change should be backward compatible in terms of produced HTTP requests, i.e. `body: "hello world"` is still a valid request (`content-type` should set either plain/text or application/json since the string is a valid JSON object). If you have a particular test case to ensure compatibility, I can check it.

Issue #385, opened 2021-12-30
Migrate to new `Azure/azure-sdk-for-go` resource management packages


antoineco:
Microsoft recommends using the new subpackages of `Azure/azure-sdk-for-go/sdk/resourcemanager` to manage resources on Azure: https://github.com/Azure/azure-sdk-for-go/tree/v60.3.0#management-new-releases

FranBarrera:
Closing this one, it was fixed with the last [PR](https://github.com/triggermesh/triggermesh/pull/1398)

PR #1398, opened 2023-04-25
Update AzureEventHubs Source & Target

FranBarrera:
Addresses part of the  https://github.com/triggermesh/triggermesh/issues/385
Replace the deprecated library `github.com/Azure/azure-event-hubs-go/v3` with `github.com/Azure/azure-sdk-for-go/sdk/messaging/azeventhubs`


tzununbekov (changes requested):

pkg/sources/adapter/azureeventhubssource/adapter.go
```diff
@@ -297,3 +336,66 @@ func sanitizeEvent(validErrs event.ValidationError, origEvent *cloudevents.Event
...
+	return connStr
+}
+
+// parseEventHubResourceID parses the given resource ID string to a
+// structured resource ID, and validates that this resource ID refers to a
+// EventHub entity.
+func parseEventHubResourceID(resIDStr string) (*v1alpha1.AzureResourceID, error) {
+	resID := &v1alpha1.AzureResourceID{}
+
+	err := json.Unmarshal([]byte(strconv.Quote(resIDStr)), resID)
```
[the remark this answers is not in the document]
FranBarrera: Actually I don't know, I copy this part from other azure components. I think we could create a separate PR to try to move all of this common code in a common file and check this.

pkg/sources/adapter/azureeventhubssource/adapter.go
```diff
@@ -188,49 +206,70 @@ func (a *adapter) Start(ctx context.Context) error {
...
-	closeCtx, cancel := context.WithTimeout(context.Background(), drainTimeout)
-	defer cancel()
-	if err := a.ehClient.Close(closeCtx); err != nil {
-		return fmt.Errorf("terminating active Event Hub message receivers: %w", err)
+		for _, event := range events {
+			err = a.handleMessage(ctx, event)
+			if err != nil {
+				return err
+			}
+		}
```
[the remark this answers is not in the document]
FranBarrera: It should be fixed now, it handles the pod termination signal properly and prevents event loss

pkg/sources/adapter/azureeventhubssource/adapter.go
```diff
@@ -188,49 +204,74 @@ func (a *adapter) Start(ctx context.Context) error {
...
+	defer func() {
+		if err := partitionClient.Close(ctx); err != nil {
+			a.logger.Errorf("closing partition client for partition %s: %v", partitionID, err)
+		}
+	}()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
```
tzununbekov, line 243: Exit on the context termination is a normal shutdown procedure, it should not return an error:

```suggestion
			// optional shutdown message
			// h.logger.Debug("Shutting down Azure Event Hubs adapter")
			return nil
```

pkg/sources/adapter/azureeventhubssource/adapter.go
```diff
@@ -188,49 +204,74 @@ func (a *adapter) Start(ctx context.Context) error {
...
+		default:
+			receiveCtx, cancel := context.WithTimeout(ctx, connTimeout)
+			events, err := partitionClient.ReceiveEvents(receiveCtx, 100, nil)
```
tzununbekov, line 246: I'm still concerned about the logic behind this line - do we receive events from the source either every 20 seconds or if there are 100+ messages in the "queue"? 20 seconds seems like a lot, maybe we could lower the default value and also expose it as a CRD parameter? I'd also put the batch size into CRD if we go this path.

tzununbekov (commented):
Looks better now

pkg/sources/adapter/azureeventhubssource/adapter.go
tzununbekov: I really think that the best option would be to have these parameters customizable because there is no universal delivery configuration that would fit everyone. If you want, we can file it as an issue and implement it later or on the first user request.

tzununbekov (approved):
Looks good, thank you!

PR #1400, opened 2023-04-27
AWS sources and targets cross-account/remote IAM support

### LINKED ISSUE #1253: Support for assume role
We have an org that does not allow iam user access.
We are allowed to access objects via the "assume role"
https://docs.aws.amazon.com/sdk-for-go/api/aws/credentials/stscreds/

While the following maybe a solution, it is not good enough, as this is only for EKS instances and not for instances outside of EKS.
https://github.com/triggermesh/triggermesh/pull/1243

tzununbekov:
AWS components' auth support for cross-account IAM impersonation.

Available auth schemas:

_static creds_
```yaml
  ...
  auth:
    credentials:
      accessKeyID:
        value: <redacted>
      secretAccessKey:
        value: <redacted>
```
-client session is created with the IAM user account whose Access Key specified in the spec. Controller also uses the same access key.

_IAM Role_
```yaml
  ...
  auth:
    iamRole: <redacted>
```
-IAM role impersonation available on EKS only. New SA will be created for the adapter, controller's SA requires manual update to include EKS role annotation. STS trust relationship configuration required on IAM.

_**New: cross-account IAM Role**_
```yaml
  ...
  auth:
    credentials:
      accessKeyID:
        value: <redacted>
      secretAccessKey:
        value: <redacted>
      assumeIamRole: <redacted>
```
-after client session is created with the static credentials, `assumeIamRole` will be used to gain IAM permissions. STS trust relationship configuration required on IAM.

Resolves #1253

### DIFF
--- a/config/300-awscloudwatchlogssource.yaml
+++ b/config/300-awscloudwatchlogssource.yaml
@@ -119,7 +119,7 @@ spec:
                       assumeIamRole:
                         description: |-
                           The ARN of an IAM role for cross-account or remote EKS cluster authorization.
-                          For more information please refer to the AWS General Reference at https://docs.aws.amazon.com/IAM/latest/UserGuide/tutorial_cross-account-with-roles.html 
+                          For more information please refer to the AWS General Reference at https://docs.aws.amazon.com/IAM/latest/UserGuide/tutorial_cross-account-with-roles.html
                         type: string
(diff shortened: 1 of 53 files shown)


tzununbekov:
@FranBarrera thanks!
@jmcx this update may need to be reflected in the docs. Not breaking nor critical though.

Issue #1396, opened 2023-04-21
Unable to use iam role for authentication with awss3source


jewel-mascarenhas:
I had earlier used the aws access key and secret for the AWSS3Source to authenticate which worked very well for me, however when I try to use the IAM role for authentication it gives me the following error:

**AWSS3Source in version "v1alpha1" cannot be handled as a AWSS3Source: strict decoding error: unknown field "spec.auth.iamrole"** 

```
apiVersion: sources.triggermesh.io/v1alpha1
kind: AWSS3Source
metadata:
  name: s3data
spec:
  arn: arn:aws:s3:::xxxxx

  eventTypes:
  - s3:ObjectCreated:*
  - s3:ObjectRemoved:*

  auth:
    iamrole: arn:aws:iam::xxxxxxx:role/new-eks-service-role

  sink:
    ref:
      apiVersion: eventing.triggermesh.io/v1alpha1
      kind: RedisBroker
      name: myredis

```
Can anyone please suggest how can I overcome this ?

sebgoa:
Hi @jewel-mascarenhas we will reply shortly. in the meantime you can join us on Slack at https://join.slack.com/t/triggermesh-community/shared_invite/zt-1kngevosm-MY7kqn9h6bT08hWh8PeltA where we can help in a more real-time fashion

FranBarrera:
Hey @jewel-mascarenhas the CRD's elements are casing sensitive, so for the iam role, should be `iamRole`, you can check it [here](https://github.com/triggermesh/triggermesh/blob/617c8a2d15c812e422140a29b5ceb4c2169a7afc/config/300-awss3source.yaml#L189) Let us know if you find any other issues.

jewel-mascarenhas:
Hello @FranBarrera thank you for the help, yes it did solve the above issue, however the AWSS3Source still fails to come up with the error NOClient, resulting in no pod/deployment coming up.

**The AWSS3Source yaml is as follows:**

```
apiVersion: sources.triggermesh.io/v1alpha1
kind: AWSS3Source
metadata:
  name: s3data-new-1

spec:
  arn: arn:aws:s3:::eks-triggermesh

  eventTypes:
  - s3:ObjectCreated:*
  - s3:ObjectRemoved:*

  auth:
    iamRole: arn:aws:iam::xxxxxx:role/new-eks-service-role

  sink:
    ref:
      apiVersion: eventing.triggermesh.io/v1alpha1
      kind: RedisBroker
[...]

PR #1243, opened 2022-12-14
AWS Targets Auth update

tzununbekov:
- IAM Role auth support in AWS targets:
   - Comprehend
   - DynamoDB
   - EventBridge
   - Kinesis
   - Lambda
   - S3
   - SNS
   - SQS

  Resolves #1238

- **BREAKING CHANGE**: All AWS target CRDs are updated to match AWS Sources' [auth structure](https://github.com/triggermesh/triggermesh/blob/v1.22.0/config/300-awscloudwatchlogssource.yaml#L67-L127).
  Should be mentioned in the release note when merged (cc @sameersbn).

  _Before:_
  ```
  apiVersion: targets.triggermesh.io/v1alpha1
  kind: AWS*
  spec:
    ...
    awsApiKey:
      secretKeyRef:
        name: aws
        key: AWS_ACCESS_KEY_ID
    awsApiSecret:
      secretKeyRef:
        name: aws
        key: AWS_SECRET_ACCESS_KEY
  ```

  _After:_
  ```
  apiVersion: targets.triggermesh.io/v1alpha1
  kind: AWS*
  spec:
    ...
    auth:
      credentials:
        accessKeyID:
          valueFromSecret:
            name: aws
            key: AWS_ACCESS_KEY_ID
        secretAccessKey:
          valueFromSecret:
            name: aws
            key: AWS_SECRET_ACCESS_KEY
  ```
  
  Hence, IAM role auth is now available as:
  ``` 
  apiVersion: targets.triggermesh.io/v1alpha1
  kind: AWS*
  spec:
    ...
    auth:
      iamRole: arn:aws:iam::<redacted>:role/<role-name>
  ```

  Resolves #1114

___

Note for Docs (cc @jmcx)

After this PR is merged and released, all AWS Target docs and samples should be updated according to the "_Before/After_" example above.

Here is a draft "EKS IAM Role auth" instruction:

> TriggerMesh AWS Sources and Targets on EKS support authentication based on [IAM Roles](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html). Authentication configuration consists of two simple steps: enable OIDC provider for EKS cluster and create IAM Role. There are several different ways to make these configuration steps, but in this instruction, we'll go with the CLI tools: `aws` and `eksctl` ([installation](https://docs.aws.amazon.com/eks/latest/userguide/eksctl.html)).
> 
> Enabling OIDC provider:
> 
> ```
[...]

PR #1401, opened 2023-04-27
AWSS3Source - Fix client to use iamRole

FranBarrera:
Closes https://github.com/triggermesh/triggermesh/issues/1396

Note:
This PR fix the issue with the aws client to use the `iamRole` but it's still needed to follow the `EKS IAM Role Auth` instructions explained in this PR: https://github.com/triggermesh/triggermesh/pull/1243

### DIFF
--- a/pkg/sources/client/s3/client.go
+++ b/pkg/sources/client/s3/client.go
@@ -24,7 +24,6 @@ import (
 
 	awscore "github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
-	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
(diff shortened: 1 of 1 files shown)


tzununbekov (approved):
@FranBarrera did you figure out why previous version didn't work? I'm not quite sure what was changed in this PR from functional point of view

FranBarrera:
@tzununbekov Yes, the problem was with this code: https://github.com/triggermesh/triggermesh/pull/1401/files#diff-c1ccaf7d16fe601df5071a72ffe5d1c25c8953ce39807b3a2fba4d480183f485L90-L91

That code was failing to get the cred value, it was requiring to put our generated iam user in the users  iam role trusted policy:

This was the error in the triggermesh-controller:
```
Error creating AWS API clients: retrieving AWS IAM Role: AccessDenied: User: arn:aws:sts::0000000000000:assumed-role/0000000c/i-000000000 is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::0000000000000:role/fran-test-s3
```

Now it works like the others AWS Sources and doesn't require extra permissions.

PR #1410, opened 2023-05-03
Add support for Azure ServiceBus Target

### LINKED ISSUE #1399: Add support for AzureServiceBus as a target
As request by a user, we should add support for Azure Service Bus as a target.

It seems like there is no reason to have a dedicated component for Topics and Queues, we could combine them into one. And we should also release a new source that combines the two, and announce deprecation of the two separate ones.

FranBarrera:
Closes #1399
This PR adds support for Azure Service Bus Target.
Include support for azure service bus topic and queue


pkg/targets/adapter/azureservicebustarget/adapter.go
```diff
@@ -0,0 +1,271 @@
...
+		msg = jsonEvent
+	}
+	err := a.sender.SendMessage(ctx, &azservicebus.Message{
+		Body: msg,
+	}, nil)
+	if err != nil {
+		a.logger.Errorw("Error sending message to Service Bus", zap.Error(err))
+		a.sr.ReportProcessingError(true, ceTypeTag, ceSrcTag)
+		return a.replier.Error(&event, targetce.ErrorCodeAdapterProcess, err, nil)
+	}
```
odacremolbap, line 158: nit: could be slightly more concise
```suggestion
	if err := a.sender.SendMessage(ctx, &azservicebus.Message{Body: msg}, nil);  err != nil {
		a.logger.Errorw("Error sending message to Service Bus", zap.Error(err))
		a.sr.ReportProcessingError(true, ceTypeTag, ceSrcTag)
		return a.replier.Error(&event, targetce.ErrorCodeAdapterProcess, err, nil)
	}
```

pkg/targets/reconciler/azureservicebustarget/adapter.go
```diff
@@ -0,0 +1,111 @@
...
+
+	var webSocketsEnable bool
+	if wss := o.Spec.WebSocketsEnable; wss != nil {
+		webSocketsEnable = *wss
+	}
+
+	if sasAuth := o.Spec.Auth.SASToken; sasAuth != nil {
+		envs = common.MaybeAppendValueFromEnvVar(envs, common.EnvHubKeyName, sasAuth.KeyName)
+		envs = common.MaybeAppendValueFromEnvVar(envs, common.EnvHubKeyValue, sasAuth.KeyValue)
+		envs = common.MaybeAppendValueFromEnvVar(envs, common.EnvHubConnStr, sasAuth.ConnectionString)
```
odacremolbap, line 88: these values do not match this snippet from the adapter, the reconciler setting the `EVENTHUB_` prefix while the adapter is documenting to receive the `SERVICEBUS_` prefix.

```go
	// The environment variables below aren't read from the envConfig struct
	// by the Service Bus SDK, but rather directly using os.Getenv().
	// They are nevertheless listed here for documentation purposes.
	_ string `envconfig:"AZURE_TENANT_ID"`
	_ string `envconfig:"AZURE_CLIENT_ID"`
	_ string `envconfig:"AZURE_CLIENT_SECRET"`
	_ string `envconfig:"SERVICEBUS_KEY_NAME"`
	_ string `envconfig:"SERVICEBUS_KEY_VALUE"`
	_ string `envconfig:"SERVICEBUS_CONNECTION_STRING"`
```

odacremolbap (changes requested):
Looking very good.
Added some nits.

odacremolbap (approved):

PR #1433, opened 2023-05-19
DatadogTarget: Customize site variable

adrien-f:
Datadog is multi-site and each site gets its own domain so we need to be able to customize it.

See https://docs.datadoghq.com/getting_started/site/

### DIFF
--- a/pkg/apis/targets/v1alpha1/datadog_types.go
+++ b/pkg/apis/targets/v1alpha1/datadog_types.go
@@ -48,6 +48,10 @@ type DatadogTargetSpec struct {
 	// DatadogApiKey represents how Datadog credentials should be provided in the secret
 	DatadogAPIKey SecretValueFromSource `json:"apiKey"`
 
+	// DatadogSite controls the site of the Datadog intake API, defaults to `datadoghq.com`
+	// +optional
+	DatadogSite *string `json:"site,omitempty"`
+
 	// MetricPrefix is prepended to the name of the associated metrics.
 	// +optional
 	MetricPrefix *string `json:"metricPrefix"`
--- a/pkg/targets/adapter/datadogtarget/adapter.go
+++ b/pkg/targets/adapter/datadogtarget/adapter.go
@@ -37,9 +37,9 @@ import (
 )
 
 const (
-	apiBaseURL     = "https://api.datadoghq.com"
-	logsAPIBaseURL = "https://http-intake.logs.datadoghq.com"
-	apiKeyHeader   = "DD-API-KEY"
+	apiBaseDomain     = "https://api"
+	logsAPIBaseDomain = "https://http-intake.logs"
+	apiKeyHeader      = "DD-API-KEY"
 )
 
 const (
@@ -70,7 +70,9 @@ func NewTarget(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClien
 	}
 
 	return &datadogAdapter{
-		apiKey: env.APIKey,
+		apiKey:     env.APIKey,
+		apiURL:     fmt.Sprintf("%s.%s", apiBaseDomain, env.Site),
+		apiLogsURL: fmt.Sprintf("%s.%s", logsAPIBaseDomain, env.Site),
 
 		replier:    replier,
 		httpClient: http.DefaultClient,
@@ -84,7 +86,9 @@ func NewTarget(ctx context.Context, envAcc pkgadapter.EnvConfigAccessor, ceClien
 var _ pkgadapter.Adapter = (*datadogAdapter)(nil)
 
 type datadogAdapter struct {
-	apiKey string
+	apiKey     string
+	apiURL     string
+	apiLogsURL string
 
 	replier    *targetce.Replier
 	httpClient *http.Client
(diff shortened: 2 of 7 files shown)


adrien-f:
It's my first time updating a CRD so let me know if I missed anything !

pkg/apis/targets/v1alpha1/datadog_types.go
```diff
@@ -48,6 +48,10 @@ type DatadogTargetSpec struct {
 	// DatadogApiKey represents how Datadog credentials should be provided in the secret
 	DatadogAPIKey SecretValueFromSource `json:"apiKey"`
 
+	// DatadogSite controls the site of the Datadog intake API, defaults to `datadoghq.com`
+	// +optional
+	DatadogSite *string `json:"datadogSite,omitempty"`
```
odacremolbap, line 53: CRD element is named `site` vs this one named `datadogSite`
```suggestion
	DatadogSite *string `json:"site,omitempty"`
```

pkg/targets/adapter/datadogtarget/config.go
```diff
@@ -30,6 +30,9 @@ type envAccessor struct {
 	// APIKey defines a Datadog API key to be used for authentication
 	APIKey string `envconfig:"DD_CLIENT_API_KEY" required:"true"`
 
+	// Site defines the site of the Datadog intake API, defaults to `datadoghq.com`
+	Site string `envconfig:"DD_CLIENT_SITE" default:"datadoghq.eu"`
```
odacremolbap, line 34: We should default it to the previous value `datadoghq.com` to make it backsward compatible (also the comment points to the previous value so most probably this is a glitch)

pkg/targets/adapter/datadogtarget/adapter.go
```diff
@@ -85,6 +86,7 @@ var _ pkgadapter.Adapter = (*datadogAdapter)(nil)
 
 type datadogAdapter struct {
 	apiKey string
+	site   string
```
odacremolbap, line 89: nit: we could do a very slight optimization here.

Creating 2 fields at the adapter called
- `apiURL`
- `apiLogsURL`

At the `NewTarget` function we add the base to the site suffix and assign the result to those variables above.

That way we avoid an `fmt.Sprintf` for each processed event.

odacremolbap (changes requested):
Nicely done, thanks so much for bringing this.

I asked for some optimizations, let's see what you think.

odacremolbap:
> It's my first time updating a CRD so let me know if I missed anything !

Your CRD update is good.
There might be missing some generated files. The API structures must allow serializing by _deep copying_, and thus require `DeepCopy` functions, that we generate through a `make` target.

```
make codegen
```

Can you execute that and include the updated file in this PR?

pkg/apis/targets/v1alpha1/datadog_types.go
adrien-f: Indeed, that was Goland autocomplete 🤦

pkg/targets/adapter/datadogtarget/adapter.go
adrien-f: It's done ✅ I was also tempted to move the `newLogsAPIRequest` and `newAPIRequest` into the struct to access it directly, let me know if that's something we'd prefer.

odacremolbap:
This is what I get after the `make codegen` command.

```console
❯ make codegen
+ Generating deepcopy funcs for sources/v1alpha1 targets/v1alpha1 flow/v1alpha1 extensions/v1alpha1 routing/v1alpha1
+ Generating clientsets for sources/v1alpha1 targets/v1alpha1 flow/v1alpha1 extensions/v1alpha1 routing/v1alpha1
+ Generating listers for sources/v1alpha1 targets/v1alpha1 flow/v1alpha1 extensions/v1alpha1 routing/v1alpha1
+ Generating informers for sources/v1alpha1 targets/v1alpha1 flow/v1alpha1 extensions/v1alpha1 routing/v1alpha1
+ Generating injection for sources/v1alpha1 targets/v1alpha1 flow/v1alpha1 extensions/v1alpha1 routing/v1alpha1
❯ git status
On branch datadog-site
Your branch is up to date with 'manomano/datadog-site'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   pkg/apis/targets/v1alpha1/deepcopy_generated.go

no changes added to commit (use "git add" and/or "git commit -a")
```

and this is the file's diff

```diff
diff --git a/pkg/apis/targets/v1alpha1/deepcopy_generated.go b/pkg/apis/targets/v1alpha1/deepcopy_generated.go
index 31c50f22..b3b197a6 100644
--- a/pkg/apis/targets/v1alpha1/deepcopy_generated.go
+++ b/pkg/apis/targets/v1alpha1/deepcopy_generated.go
@@ -1301,6 +1301,11 @@ func (in *DatadogTargetList) DeepCopyObject() runtime.Object {
 func (in *DatadogTargetSpec) DeepCopyInto(out *DatadogTargetSpec) {
        *out = *in
        in.DatadogAPIKey.DeepCopyInto(&out.DatadogAPIKey)
+       if in.DatadogSite != nil {
+               in, out := &in.DatadogSite, &out.DatadogSite
+               *out = new(string)
+               **out = **in
+       }
        if in.MetricPrefix != nil {
                in, out := &in.MetricPrefix, &out.MetricPrefix
                *out = new(string)
```

Yes, the command is expected to re-generate the kubernetes code under `pkg/client` but the first step should be to generate the `deep-copy`.

Can you post here the output when you exec the `make codegen`?

adrien-f:
Alright for some reason, they were generated in a folder `github.com/triggermesh/triggermesh`, probably a GOPATH/GOMOD thing.

odacremolbap (approved):
All good.

Thanks so much @adrien-f

jmcx:
FYI docs added here: https://github.com/triggermesh/docs/pull/397

PR #1453, opened 2023-06-20
AWS IAM auth SA name support

tzununbekov:


FranBarrera (approved):
+1 (I haven't tested it manually)

jmcx:
For docs, I'm assuming for now that this has no impact on the spec.auth.credentials.assumeIamRole parameter which remains in the same place. Let me know if that isn't right!

jmcx:
@tzununbekov your *Bonus 1* comment makes it sound like this feature request would be addressed by this PR: https://github.com/triggermesh/triggermesh/issues/1463

How ever, the comment from the user there indicates that this might only work for sources and not targets? Any thoughts? Perhaps there is still a TODO for GCP targets.

jmcx:
docs PR https://github.com/triggermesh/docs/pull/394

PR #1454, opened 2023-06-20
Add support to namespaced installation

### LINKED ISSUE #1439: Evaluate effort for TriggerMesh to be scoped to a single namespace

FranBarrera:
closes #1439

This PR add support to a single namespaced installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,26 @@ Deploy the platform:
 kubectl apply -f https://github.com/triggermesh/triggermesh/releases/latest/download/triggermesh.yaml
 ```
 
+## Namespaced installation
+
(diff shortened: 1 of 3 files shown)


odacremolbap (approved):
+1 if you tested manually

odacremolbap:
@FranBarrera regarding testing

did you test multi-tenant components?
they might spin up controllers, and their context should also be set with the namespaced key.

sameersbn:
We have had requests from users to configure triggermesh-controller to access to one or more namespaces. So triggermesh-controller needs to know which these namespaces are. Another thing I noticed is that the controller creates roles on the fly `awssqs-adapter` role for instance. this requires tm-controller to have rbac admin privileges. This is something certain type of users are not comfortable with.

tzununbekov (approved):

PR #1456, opened 2023-06-26
Return no content status code on empty event data

### LINKED ISSUE #992: [WebhookSource]Empty response should be 204
WebhookSource should reply with 204 when returning an empty response to caller.

nduyphuong:
Resolve #992

### DIFF
--- a/pkg/sources/adapter/webhooksource/webhook.go
+++ b/pkg/sources/adapter/webhooksource/webhook.go
@@ -197,8 +197,11 @@ func (h *webhookHandler) handleAll(ctx context.Context) http.HandlerFunc {
 		if result := h.ceClient.Send(ctx, event); !cloudevents.IsACK(result) {
 			h.handleError(fmt.Errorf("could not send Cloud Event: %w", result), http.StatusInternalServerError, w)
 		}
-
-		w.WriteHeader(http.StatusOK)
+		if event.Data() != nil {
+			w.WriteHeader(http.StatusOK)
+			return
+		}
+		w.WriteHeader(http.StatusNoContent)
 	}
 }
 
--- a/pkg/sources/adapter/webhooksource/webhook.go
+++ b/pkg/sources/adapter/webhooksource/webhook.go
@@ -194,14 +194,16 @@ func (h *webhookHandler) handleAll(ctx context.Context) http.HandlerFunc {
 			return
 		}
 
-		if result := h.ceClient.Send(ctx, event); !cloudevents.IsACK(result) {
+		rEvent, result := h.ceClient.Request(ctx, event)
+		if !cloudevents.IsACK(result) {
 			h.handleError(fmt.Errorf("could not send Cloud Event: %w", result), http.StatusInternalServerError, w)
-		}
-		if event.Data() != nil {
-			w.WriteHeader(http.StatusOK)
 			return
 		}
-		w.WriteHeader(http.StatusNoContent)
+		if rEvent.Data() == nil {
+			w.WriteHeader(http.StatusNoContent)
+			return
+		}
+		w.WriteHeader(http.StatusOK)
 	}
 }
 
--- a/pkg/sources/adapter/webhooksource/webhook.go
+++ b/pkg/sources/adapter/webhooksource/webhook.go
@@ -199,7 +199,7 @@ func (h *webhookHandler) handleAll(ctx context.Context) http.HandlerFunc {
 			h.handleError(fmt.Errorf("could not send Cloud Event: %w", result), http.StatusInternalServerError, w)
 			return
 		}
-		if rEvent.Data() == nil {
+		if rEvent == nil || rEvent.Data() == nil {
 			w.WriteHeader(http.StatusNoContent)
 			return
 		}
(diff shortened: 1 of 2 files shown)


odacremolbap:
@nduyphuong thanks a lot for your contribution!

The original issue is not well explained, sorry for that.

A source can be connected to any HTTP endpoint that consumes CloudEvents. The Webhooks source could be used either to produce events to a broker, which produces no response, or to a CloudEvents consumer that might produce a response.

We are yet to define how the WebhookSource should handle responses, for now only error responses are defined. A successful response would return a CloudEvent, which contains attributes and data.
When a response to the WebhookSource contains no data, we know that we should answer 204.

So looking at the code:

```go
		if result := h.ceClient.Send(ctx, event); !cloudevents.IsACK(result) {
			h.handleError(fmt.Errorf("could not send Cloud Event: %w", result), http.StatusInternalServerError, w)
		}
		if event.Data() != nil {
			w.WriteHeader(http.StatusOK)
			return
		}
		w.WriteHeader(http.StatusNoContent)
```

I wouls say that insted of calling `Send` at the CloudEvents client, we should use [Request](https://github.com/cloudevents/sdk-go/blob/50b18a01e534c07c4714c87255683c169442896d/v2/client/client.go#L31), then read if the response event data is empty, and if so use 204.

If you need guidance on that we are here to help.

nduyphuong:
@odacremolbap  Thanks for your warm support, I've made some change and updated test also. Please help to correct me since i'm just getting started to familiar with the code base

nduyphuong:
@odacremolbap Thanks for your warm support, I've made some changes and updated test also

odacremolbap:
@nduyphuong this is looking fine now.

We could add the new case to the tests, if you will.

The steps would be:

1. Create a helper function that creates a new event for tests. Add the type and source constants to the top of the file. Let the function produce body-full and body-less events.

```golang
func newEvent(body string) *event.Event {
       e := event.New(event.CloudEventsVersionV1)
       e.SetType(tResponseEventType)
       e.SetSource(tResponseEventSource)

       if body != "" {
               e.SetData("text/json", body)
       }

       return &e
}
```

Adding these 2 new parameters to the test case struct, we will use them at your `replierFn`. To make them backwards compatible we will consider that no result defined at cases means ACK, and no response event defined creates a default body-full response. Alternatively we could add the `responseResult` and `responseEvent` to all other cases.

```golang
               responseResult protocol.Result
               responseEvent  *event.Event
```

The `replierFunc` would need to take care of those 2 new test parameters, providing a default behavior that does not make the rest of tests fail.

```golang
			replierFn := func(inMessage event.Event) (*event.Event, protocol.Result) {
				// If the test case does not define a response event use a default one
				e := c.responseEvent
				if e == nil {
					e = newEvent(`{"test":"default"}`)
				}

				// If the test case does not define a result return ACK
				r := c.responseResult
				if r == nil {
					r = protocol.ResultACK
				}
				return e, r
			}

```

Then we could define a new test case that tests these conditions

```golang
		"empty response": {
			body: read("arbitrary message"),

			responseEvent: newEvent(""),

			expectedCode:       http.StatusNoContent,
			expectedEventData:  "arbitrary message",
			expectedExtensions: expectedExtensionsBase,
		},
```

What I don't love from my proposal here is that usually `WebHookSource` is connected to a broker, which returns no content, while here we are defaulting all cases to body-full instead. But that is probably a nit.

odacremolbap:
@nduyphuong 

It looks that there is a linting warning at my suggested code.

```
   Running [/home/runner/golangci-lint-1.53.3-linux-amd64/golangci-lint run --out-format=github-actions --out-${NO_FUTURE}format=colored-line-number --timeout 15m] in [] ...
  Error: pkg/sources/adapter/webhooksource/webhook_test.go:313:12: Error return value of `e.SetData` is not checked (errcheck)
  		e.SetData("text/json", body)
```

We don't expect an error from that call, and test cases are limited. I would say panicking is allowed in this context, sort of ...

```go
		if err := e.SetData("text/json", body); err != nil {
			panic(err)
		}

```

pkg/sources/adapter/webhooksource/webhook.go
```diff
@@ -194,10 +194,15 @@ func (h *webhookHandler) handleAll(ctx context.Context) http.HandlerFunc {
 			return
 		}
 
-		if result := h.ceClient.Send(ctx, event); !cloudevents.IsACK(result) {
+		rEvent, result := h.ceClient.Request(ctx, event)
+		if !cloudevents.IsACK(result) {
 			h.handleError(fmt.Errorf("could not send Cloud Event: %w", result), http.StatusInternalServerError, w)
+			return
+		}
+		if rEvent.Data() == nil {
```
odacremolbap, line 202: Avoid panics when no response is returned

```suggestion
		if rEvent == nil || rEvent.Data() == nil {
```

odacremolbap (changes requested):

odacremolbap (approved):

odacremolbap:
@nduyphuong thanks so much for this.

I think we might have next steps, like converting responses into payloads when a body is returned.
We might want to use the structured CloudEvents encoding.

Would you like to discuss and work on that?

nduyphuong:
@odacremolbap  Thanks for the quick review. About the next steps, can you please kindly help to describe the intention a little bit more

PR #1464, opened 2023-07-03
give tm-controller clusterscope watch privileges to knative core objects

sameersbn:


config/kustomization/namespaced/controller-rbac.yaml
```diff
@@ -0,0 +1,68 @@
...
+  - deployments
+  verbs:
+  - list
+  - watch
+  # knative core objects
+- apiGroups:
+  - serving.knative.dev
+  - eventing.knative.dev
+  - messaging.knative.dev
+  - messaging.triggermesh.dev
```
tzununbekov, line 35: Do we have this group or domain? Couldn't find it anywhere

tzununbekov (commented):
Looking good. Only one question above

config/kustomization/namespaced/controller-rbac.yaml
sameersbn: them come from the `addressable-resolver` kn clusterrole. This is the binding https://github.com/triggermesh/triggermesh/blob/main/config/202-clusterrolebindings.yaml#L101-L115

tzununbekov (approved):

